PR #171, opened 2013-10-26
Follow-up on Shared ref

joker-eph:
Continuation of the previous PR. API change: default ctor is now state unitialized and static member make_ref is mandatory for build an initialized ref.
It is more clear and removes ambiguity within constructors.


serge-sans-paille:
wow :-)

PR #377, opened 2014-12-17
Allow masking a numpy mask

### LINKED ISSUE #376: Fails to compile (with lots of errors)
```python
import numpy as np

#pythran export limit (complex128[])
def limit (x, epsilon=1e-6):
    out = np.empty (shape=x.shape, dtype=x.dtype)
    #out = np.empty_like (x, dtype=x.dtype)
    mask1 = np.abs(x) < epsilon
    out[mask1] = 0
    mask2 = np.logical_not(mask1)
    out[mask2] = x[mask2] / np.abs(x[mask2])
    return out

#pythran export clip(complex128[], float64)
def clip (z, _max):
    mask = np.abs(z) > _max
## print ('#clipped:', np.sum(mask))
    z[mask] = (limit (z[mask]) * _max)
    return z
```

serge-sans-paille:
Fixes #376

### DIFF
--- a/pythran/pythonic/types/numpy_fexpr.hpp
+++ b/pythran/pythonic/types/numpy_fexpr.hpp
@@ -144,6 +144,11 @@ namespace pythonic {
                     if(i<0) i += shape[0];
                     return fast(i);
                 }
+                template<class E>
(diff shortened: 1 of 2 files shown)


pbrunet:
Sorry for this long absence. Let's go to review codes while everybody is sleeping here :-)

PR #378, opened 2014-12-21
32bit compatibility for ndarray

serge-sans-paille:
numpy's ndarray use ssize_t to store their shape, while Pythran uses
long, for compatibility with other Python integer. This does not cause
any issue on 64 bit platform, but it was problematic under 32 bit
platform where sizeof(long) != sizeof(ssize_t).

this patch proposes a non-intrusive fix.

### DIFF
--- a/pythran/pythonic/types/ndarray.hpp
+++ b/pythran/pythonic/types/ndarray.hpp
@@ -276,7 +276,8 @@ namespace pythonic {
                 }
 
                 /* from a foreign pointer */
-                ndarray(T* data, long const* pshape):
+                template <class S>
+                ndarray(T* data, S const* pshape):
                     mem(data),
                     buffer(mem->data),
                     shape()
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
Updated!

pythran/pythonic/types/ndarray.hpp
```diff
@@ -1100,6 +1101,40 @@ namespace pythonic {
...
+      PyObject *from_descr(PyTypeObject *subtype, PyArray_Descr *descr, T *dims,
+                           void *data, int flags, PyObject *obj) {
+        npy_intp shape[N];
+        std::copy(dims, dims + N, shape);
+        return pyarray_new<npy_intp, N>{}(subtype, descr, shape, flags, obj);
+      }
+      PyObject *from_data(T *dims, int typenum, void *data) {
+        npy_intp shape[N];
+        std::copy(dims, dims + N, shape);
+        return pyarray_new<npy_intp, N>{}(shape, typenum, data);
```
pbrunet: I don't understand :-$ Why the from_data is not required here?

serge-sans-paille: gotcha

pbrunet:
Should we add a travis compilation with -m32?
Of course it will be too long to do it for each commit but may be we can do it only for the master branch?

serge-sans-paille:
On Mon, Jan 05, 2015 at 06:48:28AM -0800, pbrunet wrote:
> Should we add a travis compilation with -m32?
> Of course it will be too long to do it for each commit but may be we can do it
> only for the master branch?

I tried this, but it fails on python headers...

PR #380, opened 2014-12-31
Mrproper/generic algorithms

serge-sans-paille:


serge-sans-paille:
Updated!

pbrunet:
JFalcou said something about possibility to accumulate in registers instead of doing it in an external accumulator. I don't know if you plan to do it so I merge it. You will have to create another PR to add this.

PR #383, opened 2015-01-13
Add range values evaluation to use fast operators

pbrunet:


pythran/analyses/range_values.py
```diff
@@ -0,0 +1,448 @@
...
+
+        As it is flow insensitif, it compares it with olds values and update it
+        if needed.
+        """
+        if variable not in self.result:
+            self.result[variable] = Range(low, high)
+        else:
+            self.result[variable] = \
+                Range(min(low, self.result[variable].low),
+                      max(high, self.result[variable].high))
```
serge-sans-paille: I see a range class with an update method here ;-)

pythran/analyses/range_values.py
```diff
@@ -0,0 +1,448 @@
...
+        Range(low=-inf, high=inf)
+        >>> RangeValues.combine(Range(1, 5), Range(3, 8), ast.BitAnd())
+        Range(low=-inf, high=inf)
+        >>> RangeValues.combine(Range(-1, 5), Range(3, 8), ast.FloorDiv())
+        Range(low=-1, high=1)
+        >>> RangeValues.combine(Range(-1, 5), Range(-5, -4), ast.FloorDiv())
+        Range(low=-2, high=0)
+        >>> RangeValues.combine(Range(-1, 5), Range(-5, 3), ast.FloorDiv())
+        Range(low=-inf, high=inf)
+        """
```
serge-sans-paille: here too ;-)
You could but this in a separated file too, if you feel it's valuable

pythran/analyses/range_values.py
```diff
@@ -0,0 +1,448 @@
...
+            if range2.low <= 0 and range2.high >= 0:
+                return RangeValues.UNKNOWN_RANGE
+            # This is valid because range2 doesn't change its sign
+            res = [v1 / v2 for v1, v2 in itertools.product(range1, range2)]
+            min_, max_ = min(res), max(res)
+        elif isinstance(op, ast.Mod):
+            min_, max_ = 0, min(range2.high, max(abs(range1.high),
+                                                 abs(range1.low)))
+        elif isinstance(op, ast.Pow):
+            res = [v1 ** v2 for v1, v2 in itertools.product(range1, range2)]
```
serge-sans-paille: I see a numpy.div.outer here :-)

pythran/analyses/range_values.py
```diff
@@ -0,0 +1,448 @@
...
+        >>> import ast
+        >>> from pythran import passmanager, backend
+        >>> node = ast.parse("def foo(a, b): pass")
+        >>> pm = passmanager.PassManager("test")
+        >>> res = pm.gather(RangeValues, node)
+        >>> res['a']
+        Range(low=-inf, high=inf)
+        """
+        for global_name in self.globals:
+            self.result[global_name] = self.UNKNOWN_RANGE
```
serge-sans-paille: even for constants (think pi)?

pythran/analyses/range_values.py
```diff
@@ -0,0 +1,448 @@
...
+        Range(low=2, high=2)
+        """
+        assert isinstance(node.target, ast.Name), "For apply on variables."
+        self.add(node.target.id, *self.visit(node.iter))
+        old_range = copy.deepcopy(self.result)
+        map(self.visit, node.body)
+        for name, (low, high) in old_range.iteritems():
+            if self.result[name].low < low:
+                self.result[name] = \
+                    self.result[name]._replace(low=-float("inf"))
```
serge-sans-paille: if you had a class, you could replace inplace ;-)

pythran/analyses/range_values.py
```diff
@@ -0,0 +1,448 @@
...
+        >>> pm = passmanager.PassManager("test")
+        >>> res = pm.gather(RangeValues, node)
+        >>> res['a']
+        Range(low=-inf, high=2)
+        >>> res['b']
+        Range(low=2, high=inf)
+        >>> res['c']
+        Range(low=2, high=2)
+        """
+        old_range = copy.deepcopy(self.result)
```
serge-sans-paille: mmmh, tu oublies d'ajouter la condition du while avant d'évaluer le body là

pythran/analyses/range_values.py
```diff
@@ -0,0 +1,448 @@
...
+        map(self.visit, node.body)
+        for name, (low, high) in old_range.iteritems():
+            if self.result[name].low < low:
+                self.result[name] = \
+                    self.result[name]._replace(low=-float("inf"))
+            if self.result[name].high > high:
+                self.result[name] = \
+                    self.result[name]._replace(high=float("inf"))
+        map(self.visit, node.orelse)
```
serge-sans-paille: I am not very happy with the formalism. I think your equation looks like

Ranges(after_while) = Range(before_while) U Range(while_body) *

where the * operator is a widening operator. The equation is not as accurate as it could be (you're missing the condition) and it's not explicit. You could hade a widen operator to the Range class ;-)

pythran/analyses/range_values.py
pbrunet: It looks to be a good idea but : 
```Python
In [10]: numpy.divide.outer([1, 2], [3, 4])
Out[10]: 
array([[0, 0],
       [0, 0]])
```

I don't understand this result :-(

pythran/analyses/range_values.py
pbrunet: I will do it in the next commit to have it with "range combiners"... Should we have a special folder for this kind of code?

pythran/analyses/range_values.py
pbrunet: Yes because we need a value if we use it later.
I could have use a : `self.result.get(variable_name, UNKNOWN_RANGE)` in the visit_Name but this way, I have a better error report if I missed something.

pythran/analyses/range_values.py
pbrunet: :) namedtuple is already a good improvement compare to tuple we have in others part of the code :) but it will appear soon ;)

pythran/analyses/range_values.py
pbrunet: La condition du while ne peut pas changer la valeurs des variables, c'est pour ça que je l'ignore.

pythran/analyses/range_values.py
pbrunet: I think the equation is correct but I don't understand how condition will modify it?
What is not enough explicit? (ok to add it in the class ;) )

pythran/analyses/range_values.py
serge-sans-paille: > It looks to be a good idea but :
> In [10]: numpy.divide.outer([1, 2], [3, 4])
> Out[10]:
> array([[0, 0],
>        [0, 0]])

Remember floating point div versus integer div ?

pythran/analyses/range_values.py
serge-sans-paille: > La condition du while ne peut pas changer la valeurs des variables, c'est pour

It gives you more info

while a > 1:
    b = a +2 # here you know b in ]3, +inf[

pythran/analyses/range_values.py
serge-sans-paille: On Tue, Jan 13, 2015 at 01:25:56PM -0800, pbrunet wrote:
> In pythran/analyses/range_values.py:
> > +        Range(low=-inf, high=2)
> > +        >>> res['b']
> > +        Range(low=2, high=inf)
> > +        >>> res['c']
> > +        Range(low=2, high=2)
> > +        &quot;&quot;&quot;
> > +        old_range = copy.deepcopy(self.result)
> > +        map(self.visit, node.body)
> > +        for name, (low, high) in old_range.iteritems():
> > +            if self.result[name].low < low:
> > +                self.result[name] = \
> > +                    self.result[name]._replace(low=-float(&quot;inf&quot;))
> > +            if self.result[name].high > high:
> > +                self.result[name] = \
> > +                    self.result[name]._replace(high=float(&quot;inf&quot;))
> > +        map(self.visit, node.orelse)
> I think the equation is correct but I don't understand how condition will
> modify it?

it's correct, but not as accurate as it could be.

pythran/range.py
```diff
@@ -0,0 +1,205 @@
...
+            self.low = -float("inf")
+        if self.high > range_.high:
+            self.high = float("inf")
+
+
+UNKNOWN_RANGE = Range(-float("inf"), float("inf"))
+
+
+def range_values(args):
+    """ Function use to compute returned range value of [x]range function. """
```
serge-sans-paille: use*d*

pythran/range.py
```diff
@@ -0,0 +1,205 @@
...
+    if len(args) == 1:
+        return (0, args[0].high)
+    elif len(args) == 2:
+        return (args[0].low, args[1].high)
+    elif len(args) == 3:
+        is_neg = args[2].low < 0
+        is_pos = args[2].high > 0
+        if is_neg and is_pos:
+            return (-float("inf"), float("inf"))
```
serge-sans-paille: maybe define a   ``Range.inf`` value to avoid this idiom in various place?

pythran/range.py
```diff
@@ -0,0 +1,205 @@
...
+    """ Return boolean value return range. """
+    return Range(0, 1)
+
+
+def cmp_values(args):
+    """ Return Possible range for cmp function. """
+    return Range(-1, 1)
+
+
+def positif_values(args):
```
serge-sans-paille: positi*ve*

pythran/range.py
```diff
@@ -0,0 +1,205 @@
...
+    """ Return positif range with an upper bound. """
+    return Range(0, max([arg.high for arg in args]))
```
serge-sans-paille: the hooks seems unnecessary

pythran/tables.py
```diff
@@ -187,7 +188,7 @@ def update_effects(self, node):
         "clear": MethodIntr(),
         "copy": ConstMethodIntr(),
         "get": ConstMethodIntr(),
-        "has_key": ConstMethodIntr(),
+        "has_key": ConstMethodIntr(return_range=prange.bool_values),
```
serge-sans-paille: I wonder if we could state that it returns a bool, so we know it's a ``prange.bool_values``?

pythran/analyses/range_values.py
```diff
@@ -0,0 +1,327 @@
...
+        >>> res = pm.gather(RangeValues, node)
+        >>> res['a']
+        Range(low=2, high=2)
+        >>> res['b']
+        Range(low=2, high=2)
+        """
+        assigned_range = self.visit(node.value)
+        for target in node.targets:
+            if isinstance(target, ast.Name):
+                self.add(target.id, copy.deepcopy(assigned_range))
```
serge-sans-paille: why is the deep copy needed?

pythran/analyses/range_values.py
```diff
@@ -0,0 +1,327 @@
...
+        >>> pm = passmanager.PassManager("test")
+        >>> res = pm.gather(RangeValues, node)
+        >>> res['a']
+        Range(low=1, high=2)
+        """
+        if isinstance(node.target, ast.Name):
+            self.result[node.target.id].update(
+                combine(self.result[node.target.id],
+                        self.visit(node.value),
+                        node.op))
```
serge-sans-paille: ok, because of this kind of in-place modification.
BTW, it seems to me you're not using alias information here, which could lead to inaccurate (false?) informations.

pythran/analyses/range_values.py
```diff
@@ -0,0 +1,327 @@
...
+        >>> res['c']
+        Range(low=2, high=2)
+        """
+        assert isinstance(node.target, ast.Name), "For apply on variables."
+        if isinstance(node.iter, ast.Call):
+            for alias in self.aliases[node.iter.func].aliases:
+                if isinstance(alias, Intrinsic):
+                    self.add(node.target.id,
+                             alias.return_range_content(map(self.visit,
+                                                            node.iter.args)))
```
serge-sans-paille: nice one! You're not far from implementing the whole array stuff :-)

pythran/tables.py
pbrunet: Hum, you are right. It could avoid the the decltype(...) in typing.
I will add a BR to not forget it.

pythran/analyses/range_values.py
```diff
@@ -0,0 +1,327 @@
...
+        Range(low=-inf, high=2)
+        >>> res['b']
+        Range(low=2, high=inf)
+        >>> res['c']
+        Range(low=2, high=2)
+        """
+        old_range = copy.deepcopy(self.result)
+        map(self.visit, node.body)
+        for name, range_ in old_range.iteritems():
+            self.result[name].widen(range_)
```
serge-sans-paille: I reaaaaaaaaaaaalllllllllllllllyyyyyyyy appreciate the code simplification!

pythran/analyses/range_values.py
pbrunet: In case of : 
```a = b = 2```
without the deepcopy, the range matching with a and b is the same. So with:
```
a = b  = 2
a += 2
```
will change the range value of b and we don't want it.
(I add a comment about it)

pythran/analyses/range_values.py
pbrunet: No because integer type never alias.

serge-sans-paille:
Any performance improvement feedback?

PR #384, opened 2015-01-19
Pure runtime

pbrunet:
Sorry, it is based on range_value so look only at the second commit here.
It shows:
- use of tags (should be improve but it is a base for discussion)
- use of this attribute for numpy.fromfunction (BTW, fromfunction looks like an expression template to me to avoid temporary array creation, but it works only if it is pure...)


pbrunet:
Is this second commit OK? (then I will rebase but I don't want to use travis CPUs for nothing).

serge-sans-paille:
I think the inline check is ok and will be pruned by the compiler. You could still use the tag but syndicate the common code in a tag-agnostic function too.

serge-sans-paille:
meta -trait : I failed to this in pure template, but the idea is

```
        template<typename T>\
            struct name\
            {\
                typedef char yes;\
                typedef struct { char _[2]; } no;\
                template <class C> static constexpr yes _test(typename C::member*) { return yes{}; }\
                template <class C> static constexpr no _test(...) { return no{}; }\
                static constexpr bool value = std::is_same<decltype(_test<typename std::remove_reference<T>::type>(nullptr)), yes>::value;\
            }
        MEMBER_TYPE_TRAIT(is_iterable, iterator);
```

pbrunet:
This one have been updated.

PR #387, opened 2015-02-07
Mrpoper/harmonize declarations

serge-sans-paille:


serge-sans-paille:
This one is rebased upon the nexpr stuff...

PR #389, opened 2015-02-08
Mrproper/document supported functions

serge-sans-paille:


pbrunet:
Can you rebase it? It looks like it is based on an old version of the "PEP8 Upgrade" commit.

PR #392, opened 2015-02-12
Do not raise a random error if omp detection fails

serge-sans-paille:
If the subprocess call used to detect libgomp fails, just give up this path and let the
remaining part raise an ImportError.

### DIFF
--- a/omp/__init__.py
+++ b/omp/__init__.py
@@ -28,9 +28,14 @@ class OpenMP(object):
         paths = os.environ.get('LD_LIBRARY_PATH', '').split(':')
         for gomp in ('libgomp.so', 'libgomp.dylib'):
             cmd = [cxx, '-print-file-name=' + gomp]
-            path = os.path.dirname(check_output(cmd).strip())
-            if path:
-                paths.append(path)
+            # the subprocess can fail in various ways
+            # in that case just give up that path
+            try:
+                path = os.path.dirname(check_output(cmd).strip())
+                if path:
+                    paths.append(path)
+            except OSError:
+                pass
 
         # Try to load find libgomp shared library using loader search dirs
         libgomp_path = find_library("gomp")
--- a/pythran/pythonic/types/ndarray.hpp
+++ b/pythran/pythonic/types/ndarray.hpp
@@ -1022,12 +1022,12 @@ namespace pythonic {
             }
         };
 
-    template<class T>
-    std::tuple<types::slice> make_slices(long const* strides, long const* offsets, long const* dims, utils::int_<1>) {
+    template<class T, class S>
+    std::tuple<types::slice> make_slices(long const* strides, long const* offsets, S const* dims, utils::int_<1>) {
       return std::tuple<types::slice>(types::slice(*offsets, *offsets + *dims * *strides, *strides / sizeof(T)));
     }
-    template<class T, size_t N>
-      auto make_slices(long const* strides, long const* offsets, long const * dims, utils::int_<N>)
+    template<class T, class S, size_t N>
+      auto make_slices(long const* strides, long const* offsets, S const * dims, utils::int_<N>)
         -> decltype(std::tuple_cat(make_slices<T>(strides, offsets, dims,  utils::int_<1>()),
                                    make_slices<T>(strides + 1, offsets + 1, dims +1, utils::int_<N-1>())))
       {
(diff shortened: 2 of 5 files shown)


omp/__init__.py
```diff
@@ -28,9 +28,14 @@ def __init__(self):
...
-            path = os.path.dirname(check_output(cmd).strip())
-            if path:
-                paths.append(path)
+            # the subprocess can fail in various ways
+            # in that case just give up that path
+            try:
+                path = os.path.dirname(check_output(cmd).strip())
+                if path:
+                    paths.append(path)
+            except:
```
pbrunet: From pep8:

```
When catching exceptions, mention specific exceptions whenever possible instead of using a bare except: clause.
```
```
A bare except: clause will catch SystemExit and KeyboardInterrupt exceptions, making it harder to interrupt a program with Control-C, and can disguise other problems. If you want to catch all exceptions that signal program errors, use except Exception: (bare except is equivalent to except BaseException: ).
```

What will happen in Pythran if user explicitly ask for -fopenmp but doesn't support it?

serge-sans-paille: On Tue, Feb 17, 2015 at 10:35:40AM -0800, pbrunet wrote:
> > @@ -28,9 +28,14 @@ def __init__(self):
> >          paths = os.environ.get('LD_LIBRARY_PATH', '').split(':')
> >          for gomp in ('libgomp.so', 'libgomp.dylib'):
> >              cmd = [cxx, '-print-file-name=' + gomp]
> > -            path = os.path.dirname(check_output(cmd).strip())
> > -            if path:
> > -                paths.append(path)
> > +            # the subprocess can fail in various ways
> > +            # in that case just give up that path
> > +            try:
> > +                path = os.path.dirname(check_output(cmd).strip())
> > +                if path:
> > +                    paths.append(path)
> > +            except:
> 
> From pep8:
> 
> ```
> When catching exceptions, mention specific exceptions whenever possible instead of using a bare except: clause.
> ```
> ```
> A bare except: clause will catch SystemExit and KeyboardInterrupt exceptions, making it harder to interrupt a program with Control-C, and can disguise other problems. If you want to catch all exceptions that signal program errors, use except Exception: (bare except is equivalent to except BaseException: ).
> ```

Thanks for the quote! I'll fix this.

> What will happen in Pythran if user explicitly ask for -fopenmp but doesn't support it?

-fopenmp is only triggered at c++ compile time so it's likely the C++
compiler will grunt.

serge-sans-paille:
@pbrunet updated, let see what travis thinks of this one!

pbrunet:
LGTM as soon as travis is OK

serge-sans-paille:
@pbrunet one of the check is ok, I think it's enough!

pbrunet:
Yes, I am just trying to run validation on my computer with sivart.

serge-sans-paille:
@pbrunet : any success?

pbrunet:
Sorry I was hoping to discusse about it on IRC.

I run sivart with this config : http://pastebin.com/bJzAJ5j2

and only the jessy64 worked (The other crashed but I don't know why) so I didn't test it on 32 bits  I think.

If you prefer I can merge it as the content seem good and I guess you already run exactly the same tests.

serge-sans-paille:
On Sun, Mar 08, 2015 at 08:39:44AM -0700, pbrunet wrote:
> Sorry I was hoping to discusse about it on IRC.
> 
> I run sivart with this config : http://pastebin.com/bJzAJ5j2

At first glance, that looks good...

> and only the jessy64 worked (The other crashed but I don't know why) so I didn't test it on 32 bits  I think.

I'll check on 32 bit only this morning, just to be sure.

serge-sans-paille:
@pbrunet I've add a last portability commit to this one, you can have a look :-)

pbrunet:
It looks good to me but I don't understand the motivation for this modification.

serge-sans-paille:
On Thu, Mar 26, 2015 at 10:20:51AM -0700, pbrunet wrote:
> It looks good to me but I don't understand the motivation for this modification.

The problem is that NPY_LONG does not default to the size of a long.
Quoting the doc

 NPY_LONG�

     Equivalent to either NPY_INT or NPY_LONGLONG, depending on the
     platform

So the approach is to have a full coverage of types using their actual
size, instead of relying on the inconsistent behavior of numpy.

The issue was only visible on 32 bit platforms.

pbrunet:
From the doc, NPY_LONG may default to the size of a long. In C++, long may be equivalent to int or long long depending on the platform too. Am I wrong?

serge-sans-paille:
On Thu, Mar 26, 2015 at 12:13:46PM -0700, pbrunet wrote:
> From the doc, NPY_LONG may default to the size of a long. In C++, long may be equivalent to int or long long depending on the platform too. Am I wrong?

well, fro the doc NPY_LONG may default to  NPY_INT or NPY_LONGLONG that
is a 32 or 64 bits integer...

serge-sans-paille:
@pbrunet : both travis and sivart are ok with this one :-)

PR #394, opened 2015-02-17
Harmonize nt2 fix for is_nan and is_finite

pbrunet:


pbrunet:
Only pep8 fail on this PR (because of pep8 update). It will be fix later as I don't modify any Python file here.

jfalcou:
what's the missing function from NT2 ?

serge-sans-paille:
@jfalcou NT2 does not provide an implementation for is_nan and is_finite when the argument is a boolean, which is indeed a very useful corner case ;-)

jfalcou:
well if needed, that's a 3mn fix on our side. SHoudl have pinged us :p

serge-sans-paille:
On Tue, Feb 17, 2015 at 11:18:44PM -0800, Joel Falcou wrote:
> well if needed, that's a 3mn fix on our side. SHoudl have pinged us :p 

That would prevent you from carefully following what we're doing ;-)

joke aside, we now link to the release branch of boost.simd, so the fix
was neeed for pythran *and* it will be removed when you update
boost.simd and launch a release!

jfalcou:
i just checked nd bth is_nan and is_finite in scalar mode support bool. So something's fishy

pbrunet:
@jfalcou  I tried to remove our overload and get an error : https://travis-ci.org/pbrunet/pythran/jobs/51383504

I may check it as soon as I am back in France.

jfalcou:
Yes False is the cullprit. We added bool scalar overload for predicates in https://github.com/MetaScale/nt2/pull/911 

This will be reviewed & merged soon.

jfalcou:
https://github.com/NumScale/nt2/commit/c198bf69690dd0d1fbec597e052bc0b192c114a8

fix this, care to test just to tell us.

PR #395, opened 2015-02-20
Bugfix/fast and correct itertools combinations

serge-sans-paille:


serge-sans-paille:
Apart from the pep8 update, this PR is ok for review.

serge-sans-paille:
Updated & rebased, let see what travis thinks of this

pbrunet:
What about std::move when we return?
There was a discussion with @joker-eph on IRC about it and he said we could break RVO optimisation doing this but I am not confortable enough with this to make a decision.
@jfalcou can you explain why it is so bad compare to rewriting the full (long ...) type?

serge-sans-paille:
travis loves me <3

serge-sans-paille:
we have this constructor. ready to merge?

pbrunet:
I don't understand :-$

serge-sans-paille:
:-) @jfalcou said it's ok to use the move constructor that way ;-)

serge-sans-paille:
@pbrunet I updated the move constructor

pbrunet:
It was manually merged

PR #397, opened 2015-02-22
More use of list a numpy_expr

pbrunet:


pythran/pythonic/numpy/abs.hpp
```diff
@@ -2,11 +2,17 @@
...
 #include "pythonic/types/numexpr_to_ndarray.hpp"
 #include "pythonic/utils/numpy_traits.hpp"
+
 #include <nt2/include/functions/abs.hpp>
 
+// nt2 does not handle bool specialization, which is indeed a corner case...
+namespace nt2 {
+    constexpr bool abs(bool v) { return v;}
+}
+
```
serge-sans-paille: This should go in another commit

pythran/pythonic/numpy/around.hpp
```diff
@@ -28,11 +28,6 @@ namespace pythonic {
                 return a / (long)nt2::pow(typename types::numpy_expr_to_ndarray<E>::type::dtype(10),std::max(0L,-decimals)) * (long)nt2::pow(typename types::numpy_expr_to_ndarray<E>::type::dtype(10), std::max(0L,-decimals));
             }
 
-        template<class T>
-            typename types::numpy_expr_to_ndarray<types::list<T>>::type around(types::list<T> const& l, long decimals=0) {
-                return around(typename types::numpy_expr_to_ndarray<types::list<T>>::type(l), decimals);
-            }
-
```
serge-sans-paille: great

pythran/pythonic/numpy/array2string.hpp
```diff
@@ -8,10 +8,10 @@
 namespace pythonic {
 
     namespace numpy {
-        template<class T, size_t N>
-            types::str array2string(types::ndarray<T,N> const& a) {
+        template<class E>
+            types::str array2string(E && a) {
                 std::ostringstream oss;
-                oss << a;
+                oss << std::forward<E>(a);
```
serge-sans-paille: Super cool

pythran/pythonic/numpy/array_equal.hpp
```diff
@@ -9,24 +9,12 @@ namespace pythonic {
 
     namespace numpy {
         template<class U, class V>
-            typename std::enable_if<types::has_shape<U>::value and types::has_shape<V>::value,bool>::type array_equal(U const& u, V const&v) {
+            bool array_equal(U const& u, V const&v) {
                 if(u.shape == v.shape) {
-                    return all(u == v);
+                    return all(types::numpy_expr<operator_::proxy::eq, U, V>(u, v));
```
serge-sans-paille: why?

pythran/pythonic/numpy/array_equal.hpp
```diff
@@ -9,24 +9,12 @@ namespace pythonic {
...
-                return array_equal(typename types::numpy_expr_to_ndarray<types::list<U>>::type(u), v);
-            }
-        template<class U, class V>
-            typename std::enable_if<types::has_shape<U>::value,bool>::type array_equal(U const& u, types::list<V> const&v) {
-                return array_equal(u, typename types::numpy_expr_to_ndarray<types::list<V>>::type(v));
-            }
-        template<class U, class V>
-            bool array_equal(types::list<U> const& u, types::list<V> const&v) {
-                return array_equal(typename types::numpy_expr_to_ndarray<types::list<U>>::type(u), typename types::numpy_expr_to_ndarray<types::list<V>>::type(v));
-            }
```
serge-sans-paille: Yeah

pythran/pythonic/numpy/array_equiv.hpp
```diff
@@ -43,21 +43,6 @@ namespace pythonic {
...
-        template<class U, class V>
-            typename std::enable_if<types::has_shape<U>::value,bool>::type array_equiv(U const& u, types::list<V> const&v) {
-                return array_equiv(u, typename types::numpy_expr_to_ndarray<types::list<V>>::type(v));
-            }
-
-        template<class U, class V>
-            bool array_equiv(types::list<U> const& u, types::list<V> const&v) {
-                return array_equiv(typename types::numpy_expr_to_ndarray<types::list<U>>::type(u), typename types::numpy_expr_to_ndarray<types::list<V>>::type(v));
-            }
-
```
serge-sans-paille: Awesome

pythran/pythonic/numpy/array_str.hpp
```diff
@@ -3,21 +3,13 @@
...
 namespace pythonic {
-
     namespace numpy {
-        template<class T, size_t N>
-            types::str array_str(types::ndarray<T,N> const& a) {
-                std::ostringstream oss;
-                oss << a;
-                return oss.str();
-            }
-
```
serge-sans-paille: fantastic

pythran/pythonic/numpy/binary_repr.hpp
```diff
@@ -27,7 +27,7 @@ namespace pythonic {
         types::str binary_repr(long number, long width) {
             types::str out = binary_repr(std::abs(number));
             if(number>=0) {
-                return base_repr(std::abs(number), 2, width - out.size());
+                return base_repr(number, 2, width - out.size());
```
serge-sans-paille: :-)

pythran/pythonic/numpy/clip.hpp
```diff
@@ -7,32 +7,25 @@
...
+            template<class T, class Mi>
+                typename __combined<T, Mi>::type
+                clip(T const& v, Mi a_min) {
+                    if(v<a_min) return a_min;
+                    else return v;
+                }
+        }
+#define NUMPY_NARY_FUNC_NAME clip
+#define NUMPY_NARY_FUNC_SYM wrapper::clip
+#include "pythonic/types/numpy_nary_expr.hpp"
```
serge-sans-paille: Amazing

pythran/pythonic/numpy/eye.hpp
```diff
@@ -2,14 +2,15 @@
...
 
 namespace pythonic {
 
     namespace numpy {
+
         template<class dtype = double>
-            types::ndarray<typename types::numpy_type<dtype>::type, 2> eye(long N, long M=-1, long k=0, dtype d=dtype())
+            types::ndarray<typename types::numpy_type<dtype>::type, 2> eye(long N, long M, long k=0, dtype d=dtype())
             {
-                if(M<0) M = N;
```
serge-sans-paille: no corner case hidden in the doc of eye here?

pythran/pythonic/numpy/fabs.hpp
```diff
@@ -2,21 +2,16 @@
...
 namespace pythonic {
 
     namespace numpy {
-#define NUMPY_NARY_FUNC_NAME fabs
-#define NUMPY_NARY_FUNC_SYM nt2::abs
-#include "pythonic/types/numpy_nary_expr.hpp"
 
-    }
+            ALIAS(fabs, abs)
+            PROXY(pythonic::numpy, fabs);
```
serge-sans-paille: côôl

pythran/pythonic/numpy/flatnonzero.hpp
```diff
@@ -31,12 +31,6 @@ namespace pythonic {
                 long shape[1] = { iter - buffer };
                 return types::ndarray<long, 1>(buffer, shape);
             }
-        template<class E>
-            auto flatnonzero(types::list<E> const & l)
-            -> decltype(flatnonzero(asarray(l)))
-            {
-                return flatnonzero(asarray(l));
-            }
```
serge-sans-paille: nice one

pythran/pythonic/numpy/isclose.hpp
```diff
@@ -3,50 +3,30 @@
...
-            typename std::enable_if<types::numpy_expr_to_ndarray<U>::N == types::numpy_expr_to_ndarray<V>::N,
-                                    types::ndarray<bool, types::numpy_expr_to_ndarray<V>::N>
-                                   >::type
-            isclose(U const& u, V const& v, double rtol=1e-5, double atol=1e-8, bool equal_nan=false) {
-                types::ndarray<bool, types::numpy_expr_to_ndarray<U>::N> out(u.shape, false);
-                _isclose(u.begin(), u.end(), v.begin(), out.begin(), rtol, atol, equal_nan, utils::int_<types::numpy_expr_to_ndarray<U>::N>());
-                return out;
-            }
-
-        PROXY(pythonic::numpy, isclose);
```
serge-sans-paille: yes!

pythran/pythonic/numpy/sort.hpp
```diff
@@ -61,16 +61,6 @@ namespace pythonic {
...
-        /*
-        template<class E>
-            typename types::numpy_expr_to_ndarray<typename std::remove_cv<typename std::remove_reference<E>::type>::type>::type
-            sort(E && expr, long axis=-1)
-            {
-                typename types::numpy_expr_to_ndarray<typename std::remove_cv<typename std::remove_reference<E>::type>::type>::type out{std::forward<E>(expr)};
-                _sort(out, axis);
-                return out;
-            }
-            */
```
serge-sans-paille: easy one ;-)

pythran/pythonic/types/list.hpp
```diff
@@ -148,63 +148,52 @@ namespace pythonic {
...
-                    init_shape(res, e[0], utils::int_<L-1>{});
-                  }
-                  operator array<long, value>() const {
-                    array<long, value> res;
-                    init_shape(res, the_list, utils::int_<value>{});
-                    return res;
-                  }
-                } shape;
-
-
```
serge-sans-paille: what about operations like list.append that update the shape?

pythran/pythonic/numpy/array_equal.hpp
pbrunet: Because operator== will return a boolean for list which is not iterable while this numpy_expr notation is the same for all arraylike.

pythran/pythonic/numpy/eye.hpp
pbrunet: No. I know that some function use the -1 value for default argument while it should be None when we read the documentation.

pythran/pythonic/types/list.hpp
pbrunet: I already noticed this problem but choose to ignore it (for now).

I did this modification for 3 reasons:
- To have a common interface for future arrayLike Interface
- This class doesn't work on Ubuntu (:-s because Ubuntu still use gcc 4.8.x which generate wrong code and create an infinite loop)
- It happend that calling the shape argument of a list failed (because of missing size() function for some arguments)
ex: `np.sum([np.array([1, 2]), np.array([3, 4])])`

pythran/pythonic/numpy/abs.hpp
pbrunet: It is not a fix.
I missmatch between branches. I was rebased on your "warning free" branch which introduced this patch in isclose.hpp
We should just wait for your branch to be merge then I will rebase on it and all will be fine :-)

pythran/pythonic/numpy/eye.hpp
pbrunet: I checked and we should throw an exception in this case.

pythran/pythonic/types/list.hpp
serge-sans-paille: > I already noticed this problem but choose to ignore it (for now).
> 
> I did this modification for 3 reasons:
> - To have a common interface for future arrayLike Interface
> - This class doesn't work on Ubuntu (:-s because Ubuntu still use gcc 4.8.x which generate wrong code and create an infinite loop)
> - It happend that calling the shape argument of a list failed (because of missing size() function for some arguments)
> ex: `np.sum([np.array([1, 2]), np.array([3, 4])])`

ok for size(), but couldn't you rewrite the class so that it works on
gcc 4.8 instead of introducing a new issue?

pythran/pythonic/numpy/array_equal.hpp
serge-sans-paille: > Because operator== will return a boolean for list which is not iterable while this numpy_expr notation is the same for all arraylike.
Right!

pythran/pythonic/types/list.hpp
pbrunet: The problem is that we can't add the list reference in this new struct as list is not fully declared.

The common interface for array like is still important IMO. I would prefer update this shape when we call append for example.

serge-sans-paille: > The problem is that we can't add the list reference in this new struct as list is not fully declared.

What if the fake_shape is an inner class?

> The common interface for array like is still important IMO. I would prefer update this shape when we call append for example.

that sounds like maintaining an invariant to me...

serge-sans-paille:
travis says no :-(

serge-sans-paille:
@pbrunet: do you plan to fix this one to have it merged into the next release?

pbrunet:
No. It will not be mergeable for next release.

pythran/pythonic/numpy/abs.hpp
serge-sans-paille: time has come!

serge-sans-paille:
this one seems obsoleted by #407, @pbrunet if that's ok with you, close it!

pythran/pythonic/numpy/abs.hpp
pbrunet: It would be great be here is the same issue than the one with dot. We can't change the nt2 branch for now because of boost deps :-s

PR #399, opened 2015-03-07
Do not rely on nt2::pow when index is a known integer

serge-sans-paille:
Doing so makes it possible to achieve > x10 speedup while keeping numpy
correctness.

### DIFF
--- a/pythran/optimizations/square.py
+++ b/pythran/optimizations/square.py
@@ -4,6 +4,7 @@ from pythran.passmanager import Transformation
 from pythran.analyses.ast_matcher import ASTMatcher, AST_any
 
 import ast
+import copy
 
 
 class Square(Transformation):
@@ -36,6 +37,7 @@ class Square(Transformation):
         Transformation.__init__(self)
 
     def replace(self, value):
+        self.update = self.need_import = True
         return ast.Call(ast.Attribute(ast.Name('numpy', ast.Load()),
                                       'square', ast.Load()),
                         [value], [], None, None)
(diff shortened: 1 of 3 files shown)


jfalcou:
meanwhile, care to push an issue @NT2 on having support for pow<N>(x) using the same algorithm but unrolled by TMP ?

pythran/optimizations/square.py
```diff
@@ -48,18 +50,35 @@ def visit_Module(self, node):
...
+                return node_pow
+
     def visit_BinOp(self, node):
         self.generic_visit(node)
         if ASTMatcher(Square.POW_PATTERN).search(node):
-            self.update = self.need_import = True
             return self.replace(node.left)
+        elif isinstance(node.op, ast.Pow) and isinstance(node.right, ast.Num):
+            n = node.right.n
+            if int(n) == n and n > 0:
```
pbrunet: I guess some extra transformation can be performed with n < 0.

pythran/optimizations/square.py
```diff
@@ -4,6 +4,7 @@
 from pythran.analyses.ast_matcher import ASTMatcher, AST_any
 
 import ast
+import copy
 
 
 class Square(Transformation):
```
pbrunet: It is not about square anymore

pythran/optimizations/square.py
```diff
@@ -48,18 +50,35 @@ def visit_Module(self, node):
...
+            return ast.Num(1)
+        elif n == 1:
+            return node
+        else:
+            node_square = self.replace(node)
+            node_pow = self.expand_pow(node_square, n >> 1)
+            if n & 1:
+                return ast.BinOp(node_pow, ast.Mult(), copy.deepcopy(node))
+            else:
+                return node_pow
```
pbrunet: Why did you implement it here instead of in the Pythonic runtime?
I agree that for now, this rewriting would certainly have better performances as it would totally avoid all parity comparison for exponent at runtime.
One the other side, It will not work for values not known at compile time and it would certainly not correctly use SIMD register. (The compiler will not known that for x^n, x is used n times and we don't need to unload it to reload the same value for the next matrix)

pythran/optimizations/square.py
serge-sans-paille: > I guess some extra transformation can be performed with n < 0.
Correct!

pythran/optimizations/square.py
serge-sans-paille: On Sun, Mar 08, 2015 at 09:27:36AM -0700, pbrunet wrote:
> > @@ -48,18 +50,35 @@ def visit_Module(self, node):
> >              node.body.insert(0, importIt)
> >          return node
> >  
> > +    def expand_pow(self, node, n):
> > +        if n == 0:
> > +            return ast.Num(1)
> > +        elif n == 1:
> > +            return node
> > +        else:
> > +            node_square = self.replace(node)
> > +            node_pow = self.expand_pow(node_square, n >> 1)
> > +            if n & 1:
> > +                return ast.BinOp(node_pow, ast.Mult(), copy.deepcopy(node))
> > +            else:
> > +                return node_pow
> 
> Why did you implement it here instead of in the Pythonic runtime?
> I agree that for now, this rewriting would certainly have better performances as it would totally avoid all parity comparison for exponent at runtime.
> One the other side, It will not work for values not known at compile time and it would certainly not correctly use SIMD register. (The compiler will not known that for x^n, x is used n times and we don't need to unload it to reload the same value for the next matrix)

It works relatively for simd, because the compiler can optimize common
subepxressions etc.

NT2 handles the case where the exponenet is an integer variable, so
that's ok too.

I've opened a BR on nt2 so that they impelmented a paramterict pow<N>,
we'll rely on it once it's on their release branch

see https://github.com/NumScale/nt2/issues/914

serge-sans-paille:
travis is happy with this one \o/

PR #400, opened 2015-03-10
makepkg cannot be invoked as sudo

xantares:


serge-sans-paille:
ok, thanks for the update!

PR #402, opened 2015-03-17
Nan comparison leads to an infinite loop in the optimization loop

### LINKED ISSUE #401: nan assignment creates an infinite loop
As reported by Leopold Haimberger, the following snippet creates an infinite loop:

```python
#pythran export plus(float[],float[])
import numpy
def plus(a,b):

    b[0]=numpy.nan
    return a+b
```

serge-sans-paille:
ast matcher could not match a NaN litteral as comparison between nan is
always false.

Fixes #401


serge-sans-paille:
Travis is happy with this one, and so am I!

PR #405, opened 2015-03-18
Bugfix/more accurate error message

serge-sans-paille:


serge-sans-paille:
@pbrunet : Apart from the ``str::join `` warning, fixed in another PR, this PR looks ok.

PR #406, opened 2015-03-18
Add pwd in search paths

pbrunet:
Fix # 403

### DIFF
--- a/pythran/transformations/handle_import.py
+++ b/pythran/transformations/handle_import.py
@@ -311,7 +311,7 @@ class ImportRegistry(object):
     def generate_ImportList(self):
         """List of imported functions to be added to the main module.  """
         import_list = []
-        for _, mod in self.modules.items():
+        for mod in self.modules.values():
             if mod.is_main_module:
                 # don't need to import anything from the main module
                 continue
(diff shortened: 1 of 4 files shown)


pythran/transformations/handle_import.py
```diff
@@ -311,18 +311,17 @@ def import_module(self, name):
     def generate_ImportList(self):
         """List of imported functions to be added to the main module.  """
         import_list = []
-        for _, mod in self.modules.items():
+        for mod in self.modules.itervalues():
```
serge-sans-paille: ``values()`` for py3 compat?

pythran/transformations/handle_import.py
```diff
@@ -311,18 +311,17 @@ def import_module(self, name):
...
             if mod.is_main_module:
                 # don't need to import anything from the main module
                 continue
-            for alias, module_name in mod.dependent_modules.items():
+            for alias, module_name in mod.dependent_modules.iteritems():
```
serge-sans-paille: :-)

serge-sans-paille:
LGTM as soon as Travis agrees

scripts/pythran
```diff
@@ -149,6 +150,9 @@ try:
 
     module_name, ext = os.path.splitext(os.path.basename(args.input_file))
 
+    # Add compiled module path to search for imported modules
+    sys.path.append(os.path.dirname(args.input_file))
+
```
serge-sans-paille: sorry to comment now that travis is ok, but ``scripts/pythran`` seems a bad place to add this: what if we use pythran as a library ? the path would not be added... (as in ``%%pythran`` magic)

pbrunet:
Fixed.

For %%pythran magic, it would not have been a problem because pwd is already added in sys.path as it is an interactive session.

serge-sans-paille:
On Sun, Mar 22, 2015 at 10:45:22AM -0700, pbrunet wrote:
> Fixed.
> 
> For %%pythran magic, it would not have been a problem because pwd is already added in sys.path as it is an interactive session.

Correct! Thanks for the fix!

PR #407, opened 2015-03-19
Decl impl

pbrunet:
I create this PR even if the full conversion is not done because it touch almost all pythonic files in a not mergeable way. I would like to be approved on this before continuing. (Integration will have to be done before the full rework is done except if no other development happen before the end of this PR)


serge-sans-paille:
do you plan to have this merged *before* the release?

pbrunet:
I think it will be too long for a full merge before the release but looking at the current PR, only the 32 bits one is missing for the release, isn't it?

pbrunet:
This one is ready to be merged. I will stash the WIPbuiltin if it is OK (I prefer not to run travis until I know it is all fine).

pythran/pythonic/__builtin__/abs_impl.hpp
```diff
@@ -0,0 +1,20 @@
...
+    namespace __builtin__ {
+
+        ALIAS(abs, nt2::abs);
+        PROXY_IMPL(pythonic::__builtin__, abs);
+
+    }
+
+}
+
+#endif
```
serge-sans-paille: Why did you put this one in a separate file? It's not referenced elsewhere...

pythran/pythonic/bisect/bisect_impl.hpp
```diff
@@ -0,0 +1,25 @@
...
+            }
+
+        PROXY_IMPL(pythonic::bisect, bisect);
+
+    }
+
+}
+
+#endif
+
```
serge-sans-paille: ok, I get it, there are twice as many files now, with the _impl version coming last. Why did you choose this layout?

pythran/pythonic/__builtin__/abs_impl.hpp
pbrunet: As all others, it is a separation between implementation and declaration. Declaration is done in pythran/pythonic/__builtin__/abs.hpp

pythran/pythonic/bisect/bisect_impl.hpp
pbrunet: I choose this to break dependencies between files and classes.
Usually, we create .hpp and .cpp files in a c++ project but it is not done here because we want to keep inlining as much as possible. I don't see any other solution to both keep inlining and correct file dependencies like in other c++ project.

serge-sans-paille: On Mon, Mar 23, 2015 at 01:37:56AM -0700, pbrunet wrote:
> > +
> > +    namespace bisect {
> > +        template <class X, class A>
> > +            size_t bisect(X const& x, A const& a, long lo, long hi) {
> > +                if(hi<0) hi = x.end() - x.begin();
> > +                return std::upper_bound(x.begin()+lo, x.begin()+hi, a) - x.begin();
> > +            }
> > +
> > +        PROXY_IMPL(pythonic::bisect, bisect);
> > +
> > +    }
> > +
> > +}
> > +
> > +#endif
> > +
> 
> I choose this to break dependencies between files and classes.
> Usually, we create .hpp and .cpp files in a c++ project but it is not done here because we want to keep inlining as much as possible. I don't see any other solution to both keep inlining and correct file dependencies like in other c++ project.

There 's something I'm missing here: in that case you should have the
impl include the header and not the following, so that we can include
the header without getting the impl?

pbrunet: Yes but it will required to generate correct inclusion like:

```
#include <lala.h>
#include <lolo.h>
#include <lala_impl.h>
#include <lolo_impl.h>
```
in the Pythran generated file but it can't be done until all the xxxx_impl.hpp file are created. So for now, I include them here and once all of the xxx_impl.hpp files exists, I will generated correct include directives.
I can do all of this in a same PR but it means no modification can be done until the end of this PR...

pythran/pythonic/types/numpy_texpr.hpp
```diff
@@ -200,5 +200,15 @@ struct __combined<pythonic::types::numpy_texpr<E>, K> {
...
 
+template<class E, class T, size_t N>
+struct __combined<pythonic::types::numpy_texpr<E>, pythonic::types::ndarray<T, N>> {
+    typedef pythonic::types::ndarray<T, N> type;
+};
+
+template<class E, class T, size_t N>
+struct __combined<pythonic::types::ndarray<T, N>, pythonic::types::numpy_texpr<E>> {
+    typedef pythonic::types::ndarray<T, N> type;
+};
```
serge-sans-paille: there's the same kind of issue as the one I already spotted for gexpr here, even if my solution has some flaws.

pbrunet: I don't get the issue. If it is a strange combine, I evaluate in and save it in a ndarray. Why it is incorrect? (Pythran compiled code works with these cases).

serge-sans-paille: On Mon, Mar 23, 2015 at 06:56:34AM -0700, pbrunet wrote:
> > @@ -200,5 +200,15 @@ struct __combined<pythonic::types::numpy_texpr<E>, K> {
> >      typedef pythonic::types::numpy_texpr<E> type;
> >  };
> >  
> > +template<class E, class T, size_t N>
> > +struct __combined<pythonic::types::numpy_texpr<E>, pythonic::types::ndarray<T, N>> {
> > +    typedef pythonic::types::ndarray<T, N> type;
> > +};
> > +
> > +template<class E, class T, size_t N>
> > +struct __combined<pythonic::types::ndarray<T, N>, pythonic::types::numpy_texpr<E>> {
> > +    typedef pythonic::types::ndarray<T, N> type;
> > +};
> 
> I don't get the issue. If it is a strange combine, I evaluate in and save it in a ndarray. Why it is incorrect? (Pythran compiled code works with these cases).

a = np.array([[1,2],[3,4]])
b = a.T
b[:,0] = [0,0]
print a

que se passe-t-il ?

pythran/pythonic/types/numpy_expr.hpp
```diff
@@ -54,7 +54,7 @@ namespace pythonic {
                 typedef decltype(Op()(std::declval<typename std::remove_reference<Args>::type::dtype>()...)) dtype;
 
                 std::tuple<typename std::remove_reference<Args>::type...> args;
-                array<long, value> shape;
+                types::array<long, value> shape;
```
serge-sans-paille: so you keep a long here (I think it's a good choice), so why do you want size_t everywhere?

pythran/pythonic/types/numpy_texpr.hpp
pbrunet: Hum ok, aliasing issue.
We are really not good enough on these kind of issue as even for  the Pythran compiler (compile time part) do not detect this AFAIK.

pythran/pythonic/types/numpy_expr.hpp
pbrunet: I think it is better to have correct typing when it is possible as backend compiler will not it can't be negative and certainly do some tricky improvement (You spoke about signed/unsigned comparison and some kind of change like this)
I didn't do the modification here because it is a parameter from user input in some cases so it would lead to too many change in the code so I don't want to do it at the same time.

serge-sans-paille:
As far as i understand, this PR is the first step for a larger one. The goal of this PR is most certainly to avoid header dependency hell that is starting to pollute the already-no-so-nice pythonic code base. I think this deserve a different directory namespace instead of the _impl suffix. I've seen the .tcc extension for that purpose, or use a ``src`` directory inside pythonic that matches the existing hierarchy, as in:

```
pythonic/
   numpy/
      cos.hpp   // fwd decl
   src/
      numpy/
         cos.hpp // impl
```

serge-sans-paille:
just to be sure : you don't want this into next release, do you?

pbrunet:
No, It is a too big clean up. We will not be able to add it before PyData.

pbrunet:
I don't really like the "src" directory because I think we will have real source for Pythonic (and a pythonic.so) in the future. (Ok, inlining is good but complex function are not inlined anyway)
Indeed, the _impl suffix is what we use for kaapi/starpu and klang :)

pbrunet:
We can use a "libs" folder. Just like clang. They have a libs folder for src and header with "internal scope" and include folder for exported interface.

serge-sans-paille:
On Mon, Mar 23, 2015 at 09:57:37AM -0700, pbrunet wrote:
> I don't really like the "src" directory because I think we will have real source for Pythonic (and a pythonic.so) in the future. (Ok, inlining is good but complex function are not inlined anyway)
> Indeed, the _impl suffix is what we use for kaapi/starpu and klang :)

I was about to tell you that we should adopt the standard include/stuff
for declaration, as done in nt2 :-)

So what about

pythonic/include/numpy/cos.hpp # declaration

and

pythonic/numpy/cos.hpp # definition

pbrunet:
Ok for this. But how can we do this modification "smoothly"?
I suggest to : 
1) Continue with xxx_impl.hpp
2) Make this change when we will need to change includes directive generation.

Do you have any other plan for this?

pbrunet:
This one works on my travis : https://travis-ci.org/pbrunet/pythran/builds/56282288
For now, it contains header spliting for : 
- list
-__builtin__
- bisect
I do not perform modification on the numpy part as it is the most active for now.

pythran/pythonic/__builtin__/abs.hpp
```diff
@@ -1,7 +1,8 @@
-#ifndef PYTHONIC_BUILTIN_ABS_HPP
-#define PYTHONIC_BUILTIN_ABS_HPP
+#ifndef PYTHONIC_BUILTIN_ABS_IMPL_HPP
+#define PYTHONIC_BUILTIN_ABS_IMPL_HPP
```
serge-sans-paille: that's the first file where you rename the include guard. Why?

pbrunet: :-s It is just a rebase issue from the previous version.

serge-sans-paille:
the concat commit can now be safely removed

serge-sans-paille:
I wonder if it's not too messy to have the array_like class and the split header in the same PR?

serge-sans-paille:
If I understand this PR correctly, the goal is to have the declaration in a file, and the definition in an other file. then if I depend of function foo, I'm going to :+1: 
```c++
#include "include/pythonic/module/foo.hpp" // get the declaration

// use the function

#include "pythonic/module/hoo.hpp" // get the definition
```

Am I right?

pbrunet:
Yes it is how it will work. I agree that they may have been split. I did it this way because the array_like part have a lot of modification on list.hpp and the module split too (in a not mergeable way). As these modification are really tedious, I don't want to do the job twice.
Also, they are related as the array_like feature can't be totally use because of function dependencies (solved with the module split part) in the numpy_gexpr file for example.

serge-sans-paille:
@pbrunet: just so you knwo, I'm currently porting list / ndarray so that the remaining commits will be ported automatically.

serge-sans-paille:
Obsoleted by #445

serge-sans-paille:
@pbrunet I only ported the changes up to the str split, the remaining should be ok

I did not keep the long → size_t transformation

PR #431, opened 2015-04-14
fast access generation may happen with filtering indexing

pbrunet:


pythran/pythonic/types/ndarray.hpp
```diff
@@ -518,6 +519,15 @@ namespace pythonic {
...
 
+                /*
+                 * Default fast accessor when no fast overload are provided
+                 */
+                template <class Type>
+                    auto fast(Type && t) -> decltype(this->operator[](std::forward<Type>(t)))
+                    {
+                        return (*this)[std::forward<Type>(t)];
+                    }
+
```
serge-sans-paille: this seem dangerous to me. Your overload will take priority over conversion.

```c++
#include <cstdio>
#include <utility>

template<class T>
long foo(T&& t) { puts(__PRETTY_FUNCTION__); return std::forward<T>(t); }
long foo(long t) { puts(__PRETTY_FUNCTION__); return t;}
int main() {
  foo(1);
  foo(1L);
  return 0;
}
```
outputs
```
long foo(T &&) [T = int]
long foo(long
```

serge-sans-paille:
I'll wait for this one to be merged to release 0.7

PR #436, opened 2015-04-18
Use a member function instead of a member variable to implement shape

serge-sans-paille:
This harmonizes ndarray, array and list to that respect.

### DIFF
--- a/pythran/pythonic/types/tuple.hpp
+++ b/pythran/pythonic/types/tuple.hpp
@@ -76,34 +76,22 @@ namespace {
 namespace pythonic {
 
     namespace types {
+
+        namespace details {
+            template<class E, size_t Value> void init_shape(array<long, Value>& res, E const& e, utils::int_<1>) {
+              res[Value - 1] = e.size();
+            }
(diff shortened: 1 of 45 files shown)


serge-sans-paille:
travis is OK :-)

pbrunet:
Yes but there are comments.

serge-sans-paille:
updated!

pythran/pythonic/types/tuple.hpp
```diff
@@ -296,11 +311,11 @@ namespace pythonic {
 
                 template<class... Types>
                     operator std::tuple<Types...>() const {
-                        return std::tuple<Types...>(to_tuple_impl<N>()(*this));
+                        return array_to_tuple(*this, typename utils::gens<N+1>::type{});
                     }
 
-                auto to_tuple() const -> decltype(to_tuple_impl<N>()(*this)) {
-                    return to_tuple_impl<N>()(*this);
```
pbrunet: I like the new implementation but to_tuple_impl should be remove in this case. (It is not use anymore)

serge-sans-paille: done

serge-sans-paille:
@pbrunet ready to go!

PR #437, opened 2015-04-19
Feature/more list as array

serge-sans-paille:


pbrunet:
I thinks you just need to rebase it.

serge-sans-paille:
@pbrunet that's originally your commit, so... I merge it, travis is ok.

PR #440, opened 2015-04-20
Fix compile-tile computation of reverse sequence

### LINKED ISSUE #439: Error in broadcast_copy
This simple code doesn't compile

#pythran export f2m(float64[][], float64[][])
def f2m(f, m):
	m[:, 0] =  f[:, 0]

serge-sans-paille:
This fixes #439

### DIFF
--- a/pythran/pythonic/types/numpy_gexpr.hpp
+++ b/pythran/pythonic/types/numpy_gexpr.hpp
@@ -476,7 +476,7 @@ namespace pythonic {
                 template<class E>
                   typename std::enable_if<not may_overlap_gexpr<E>::value, numpy_gexpr&>::type
                   _copy(E const& expr) {
-                    //static_assert(value>=  utils::dim_of<E>::value, "dimensions match");
+                    static_assert(value>=  utils::dim_of<E>::value, "dimensions match");
                     return utils::broadcast_copy<numpy_gexpr&, E, value, value - utils::dim_of<E>::value,
                                                  is_vectorizable and is_vectorizable_array<E>::value>(*this, expr);
                   }
--- a/pythran/pythonic/types/numpy_gexpr.hpp
+++ b/pythran/pythonic/types/numpy_gexpr.hpp
@@ -469,12 +469,14 @@ namespace pythonic {
                 template<class E>
                   typename std::enable_if<may_overlap_gexpr<E>::value, numpy_gexpr&>::type
                   _copy(E const& expr) {
+                    static_assert(value >=  utils::dim_of<E>::value, "dimensions match");
                     return utils::broadcast_copy<numpy_gexpr&, ndarray<typename E::dtype, E::value>, value, value - utils::dim_of<E>::value,
                                                  is_vectorizable>(*this, ndarray<typename E::dtype, E::value>{expr});
                   }
                 template<class E>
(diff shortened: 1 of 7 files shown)


pythran/pythonic/types/numpy_gexpr.hpp
```diff
@@ -468,12 +468,14 @@ namespace pythonic {
...
                   typename std::enable_if<may_overlap_gexpr<E>::value, numpy_gexpr&>::type
                   _copy(E const& expr) {
+                    static_assert(value >=  utils::dim_of<E>::value, "dimensions match");
                     return utils::broadcast_copy<numpy_gexpr&, ndarray<typename E::dtype, E::value>, value, value - utils::dim_of<E>::value,
                                                  is_vectorizable>(*this, ndarray<typename E::dtype, E::value>{expr});
                   }
                 template<class E>
                   typename std::enable_if<not may_overlap_gexpr<E>::value, numpy_gexpr&>::type
                   _copy(E const& expr) {
+                    //static_assert(value>=  utils::dim_of<E>::value, "dimensions match");
```
pbrunet: Why not the static_assert here?

pythran/pythonic/utils/seq.hpp
```diff
@@ -20,7 +20,7 @@ namespace pythonic {
             struct rgens : rgens<N-1, sizeof...(S), S...> { };
 
         template<int ...S>
-            struct rgens<1, S...> {
+            struct rgens<0, S...> {
```
pbrunet: It should not be the sizeof...(S) + 1 instead? To be consistent with gens?
It looks like gens<5> => gens<4, 4> => gens<3, 3, 4> => gens<2, 2, 3, 4> => gens<1, 1, 2, 3, 4> => seq<1, 2, 3, 4>
whil rgens<5> => rgens<4, 0> => rgens<3, 1, 0> => rgens<2, 2, 1, 0> => rgens<1, 3, 2, 1, 0> => rgens<0, 4, 3, 2, 1, 0> => seq<4, 3, 2, 1, 0>
with sizeof...(S) -1, you will have seq<4, 3, 2, 1>

BTW, I have to understand this each time I come here. Can you just add a comment about what is happening? 
SOmethink like:
```
rgens<S> => seq<S-1, ..., 1>
gens<S> => seq<1, ..., S - 1>
```

If you prefer to keep it that way, it is not a problem to me but please add a comment about the values in the sequence.

Also, I wonder if gens<S> should not return seq<0, S-1>? It would avoid the +1, -1 when we use it.
Or may be we should just switch to use C++14 : http://en.cppreference.com/w/cpp/utility/integer_sequence

pbrunet:
You need a rebase here :-s

serge-sans-paille:
rebase done!

pythran/pythonic/types/numpy_gexpr.hpp
serge-sans-paille: > > +                    //static_assert(value>=  utils::dim_of<E>::value, "dimensions match");
> 
> Why not the static_assert here?
ahah, debug code inside

serge-sans-paille:
@pbrunet trvis is ok with this one!

pbrunet:
Great, I just add comments on seq.hpp and merge it.

pbrunet:
manually merged

PR #441, opened 2015-04-20
Detection of unimplemented numpy methods in order to avoid their use

Pikalchemist:
copyto and isclose numpy methods are not always implemented, depending of the version of numpy

### DIFF
--- a/pythran/tables.py
+++ b/pythran/tables.py
@@ -873,6 +873,13 @@ except ImportError:
     logger.warn("Pythran support disabled for module: omp")
     del MODULES["omp"]
 
+# check and delete unimplemented numpy methods
+for method in MODULES['numpy'].keys():
+    if (method not in sys.modules['numpy'].__dict__ and not
+            (method[-1:] == '_' and method[:-1] in cxx_keywords and
+                method[:-1] in sys.modules['numpy'].__dict__)):
+        del MODULES['numpy'][method]
+
 
 # populate argument description through introspection
 def save_arguments(module_name, elements):


pythran/tables.py
```diff
@@ -873,6 +873,12 @@ def update_effects(self, node):
     logger.warn("Pythran support disabled for module: omp")
     del MODULES["omp"]
 
+# copyto and isclose are not always supported by numpy
```
serge-sans-paille: maybe add the version of numpy that cause the issue in the comment here, otherwise the whol thing looks like: why do we support this at all?

pythran/tables.py
```diff
@@ -873,6 +873,12 @@ def update_effects(self, node):
...
+unsupported_numpy_methods = ('copyto', 'isclose')
```
serge-sans-paille: and in the name here, as in unsupported_numpyXY_methods

pbrunet: Why it is wrong to support "more than what is provided by numpy"? I mean, users should be happy to be able to use more functionnalities.

pythran/tables.py
```diff
@@ -873,6 +873,14 @@ def update_effects(self, node):
     logger.warn("Pythran support disabled for module: omp")
     del MODULES["omp"]
 
+# check and deletion of unimplemented numpy methods
+unsupported_numpy_methods = []
+for method in MODULES['numpy']:
+    if not method in sys.modules['numpy'].__dict__ and not (method[-1:] == '_' and method[:-1] in cxx_keywords and method[:-1] in sys.modules['numpy'].__dict__):
+        unsupported_numpy_methods.append(method)
+for method in unsupported_numpy_methods:
+    del MODULES['numpy'][method]
```
serge-sans-paille: why a double loop here? just iterator over ``MODULES['numpy'].keys()`` !

BTW, your code is certainly not pep8 compliant, please run the validation before submitting a PR ;-)

pythran/tables.py
serge-sans-paille: It's wrong because it fails later on during argument introspection, and
probably constant folding too.

So basically, if you have a slightly old numpy version, pythran crahses
all the time. This fixes that.

PR #444, opened 2015-04-24
Fix pythran -L option parsing

serge-sans-paille:
Credits go to Marc Gailhanou for the bug report!


serge-sans-paille:
ready to merge!

PR #446, opened 2015-04-26
Globals

pbrunet:
This one is ready for review.

* static variable for globals look useless as constant folding already expend these values.
* local_decl and local_declaration looks really similare. I didn't merge them because
they don't return the same type. We should be able to return only strings but it can't be done without modification in typing (and I am not sure it is a good idea...). If we return only ast.Name, we have to create dummy ast.Name (for function name for example) and it will also need modification in typing.
We could return a dict with both string_name and ast.Name but typing will have to handle dummy ast.Name (once again, we have to touch typing)

### DIFF
--- a/doc/INTERNAL.rst
+++ b/doc/INTERNAL.rst
@@ -203,23 +203,24 @@ cases of Python programs::
     $> pythran internal_top_level_assign.py
     $> python -c 'import internal_top_level_assign'
     13
-    14
 
 Then let's test Python's flow control keywords: for, while, if-else:
 
 Loop statement at top-level::
 
-    $> printf 's=0\nfor i in range(10): print s' > internal_top_level_loop.py
+    $> printf 's=0\nfor i in range(2): print s' > internal_top_level_loop.py
     $> pythran internal_top_level_loop.py
     $> python -c 'import internal_top_level_loop'
-    45
+    0
+    0
 
 While statement at top-level::
 
     $> printf 'i=0\nwhile True:\n print i\n break\nprint i' > internal_top_level_while.py
     $> pythran internal_top_level_while.py
     $> python -c 'import internal_top_level_while'
-    10
+    0
+    0
 
 If-else statement at top-level::
 
--- a/doc/INTERNAL.rst
+++ b/doc/INTERNAL.rst
@@ -199,7 +199,7 @@ show that Pythran can compile it::
 Assignment and AugAssignment statements at top-level are one of the most common
 cases of Python programs::
 
-    $> printf 'a = 1 + (2 + 2) * 3\nprint a\na += 1\nprint a' > internal_top_level_assign.py
+    $> printf 'a = 1 + (2 + 2) * 3\nprint a' > internal_top_level_assign.py
     $> pythran internal_top_level_assign.py
     $> python -c 'import internal_top_level_assign'
     13
@@ -209,14 +209,14 @@ Then let's test Python's flow control keywords: for, while, if-else:
 
 Loop statement at top-level::
 
-    $> printf 's=0\nfor i in range(10):s+=i\nprint s' > internal_top_level_loop.py
+    $> printf 's=0\nfor i in range(10): print s' > internal_top_level_loop.py
     $> pythran internal_top_level_loop.py
     $> python -c 'import internal_top_level_loop'
     45
 
 While statement at top-level::
 
-    $> printf 'i=0\nwhile i<10:i+=1\nprint i' > internal_top_level_while.py
+    $> printf 'i=0\nwhile True:\n print i\n break\nprint i' > internal_top_level_while.py
     $> pythran internal_top_level_while.py
     $> python -c 'import internal_top_level_while'
     10
--- a/pythran/analyses/global_declarations.py
+++ b/pythran/analyses/global_declarations.py
@@ -1,22 +1,46 @@
-"""
-GlobalDeclarations gathers top-level declarations
-"""
+""" GlobalDeclarations gathers top-level declarations. """
 
 from pythran.passmanager import ModuleAnalysis
 
 
 class GlobalDeclarations(ModuleAnalysis):
-    """Generates a function name -> function node binding"""
+
+    """ Gather all kind of identifier defined at global scope.
+
+    >>> import ast
+    >>> from pythran import passmanager
+    >>> from pythran.analyses import GlobalDeclarations
+    >>> node = ast.parse('''
+    ... import math
+    ... from math import cos
+    ... c = 12
+    ... def foo(a):
+    ...     b = a + 1''')
+    >>> pm = passmanager.PassManager("test")
+    >>> pm.gather(GlobalDeclarations, node).keys()
+    ['cos', 'foo', 'c', 'math']
+
+    """
+
     def __init__(self):
+        """ Result is an identifier with matching definition. """
         self.result = dict()
         super(GlobalDeclarations, self).__init__()
 
     def visit_Import(self, node):
+        """ Import module define a new variable name. """
         self.result.update((a.name, a) for a in node.names)
 
     def visit_ImportFrom(self, node):
+        """ Imported functions define a new variable name. """
         self.result.update((a.asname or a.name, a) for a in node.names)
 
     def visit_FunctionDef(self, node):
+        """ Function definition defined a new identifier. """
         self.result[node.name] = node
         # no generic visit here, so no diving into function body
+
+    def visit_Assign(self, node):
+        """ Assigned global variable define a new identifier. """
+        for target in node.targets:
+            self.result[target.id] = node
--- a/pythran/backend.py
+++ b/pythran/backend.py
@@ -11,7 +11,7 @@ from pythran.analyses import YieldPoints, IsAssigned, ASTMatcher, AST_any
 from pythran.analyses import RangeValues, PureExpressions
 from pythran.cxxgen import Template, Include, Namespace, CompilationUnit
 from pythran.cxxgen import Statement, Block, AnnotatedStatement, Typedef
-from pythran.cxxgen import Value, FunctionDeclaration, EmptyStatement
+from pythran.cxxgen import Value, FunctionDeclaration, EmptyStatement, Static
 from pythran.cxxgen import FunctionBody, Line, ReturnStatement, Struct, Assign
 from pythran.cxxgen import For, While, TryExcept, ExceptHandler, If, AutoFor
 from pythran.cxxtypes import Assignable, DeclType, NamedType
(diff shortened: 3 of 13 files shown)


doc/INTERNAL.rst
```diff
@@ -200,27 +200,28 @@ show that Pythran can compile it::
...
 
 While statement at top-level::
 
-    $> printf 'i=0\nwhile i<10:i+=1\nprint i' > internal_top_level_while.py
+    $> printf 'i=0\nwhile True:\n print i\n break\nprint i' > internal_top_level_while.py
     $> pythran internal_top_level_while.py
     $> python -c 'import internal_top_level_while'
-    10
+    0
+    0
```
serge-sans-paille: why did you change these examples? The new version do not look that nice, especially the break in the while.

pythran/analyses/global_declarations.py
```diff
@@ -1,22 +1,46 @@
...
+    ... from math import cos
+    ... c = 12
+    ... def foo(a):
+    ...     b = a + 1''')
+    >>> pm = passmanager.PassManager("test")
+    >>> pm.gather(GlobalDeclarations, node).keys()
+    ['cos', 'foo', 'c', 'math']
+
+    """
+
```
serge-sans-paille: I really appreciate the documentation effort!

pythran/analyses/global_declarations.py
```diff
@@ -1,22 +1,46 @@
...
+        """ Imported functions define a new variable name. """
         self.result.update((a.asname or a.name, a) for a in node.names)
 
     def visit_FunctionDef(self, node):
+        """ Function definition defined a new identifier. """
         self.result[node.name] = node
         # no generic visit here, so no diving into function body
+
+    def visit_Assign(self, node):
+        """ Assigned global variable define a new identifier. """
```
serge-sans-paille: an assert isinstance ast.Name would not hurt here. You could use dict.update to avoid the explicit loop ;-)

doc/INTERNAL.rst
pbrunet: Because i += 1 is a modification on globals which is not allowed anymore.

pythran/backend.py
```diff
@@ -573,7 +573,12 @@ def visit_Return(self, node):
                     Cxx.final_statement))
                 ])
         else:
-            stmt = ReturnStatement(self.visit(node.value))
+            value = self.visit(node.value)
+            if metadata.get(node, metadata.StaticReturn):
+                stmt = Block([Assign("static auto tmp_global", value),
+                              ReturnStatement("tmp_global")])
+            else:
+                stmt = ReturnStatement(value)
```
serge-sans-paille: What about adding a static qualifier to the type, and not modifying the back-end?

pythran/cxxtypes.py
```diff
@@ -40,7 +40,8 @@ def all_types(self):
     def __eq__(self, other):
         havesameclass = self.__class__ == other.__class__
         if havesameclass:
-            def same(x, y): return getattr(self, x) == getattr(other, y)
+            def same(x, y):
+                return getattr(self, x) == getattr(other, y)
```
serge-sans-paille: Fixed in the decl-impl branch?

pythran/syntax.py
```diff
@@ -46,18 +46,6 @@ def save_attribute(module):
...
-               " or imports")
-        for n in node.body:
-            if isinstance(n, ast.Expr) and isinstance(n.value, ast.Str):
-                continue
-            else:
-                if not any(isinstance(n, getattr(ast, t))
-                           for t in ('FunctionDef', 'Import', 'ImportFrom',)):
-                    raise PythranSyntaxError(err, n)
-        self.generic_visit(node)
-
```
serge-sans-paille: check we only have top-level assign to name?

pythran/transformations/expand_globals.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def visit_Module(self, node):
+        """Turn globals assignement to functionDef and visit function defs. """
+        module_body = list()
+        # Gather top level assigned variables.
+        for stmt in node.body:
+            if not isinstance(stmt, ast.Assign):
+                continue
+            for target in stmt.targets:
+                assert isinstance(target, ast.Name)
```
serge-sans-paille: this looks like an analyse to me, you should filter out Function from the global decl instead?

pythran/transformations/expand_globals.py
```diff
@@ -0,0 +1,82 @@
...
+                cst_value = self.visit(stmt.value)
+                for target in stmt.targets:
+                    assert isinstance(target, ast.Name)
+                    module_body.append(
+                        ast.FunctionDef(target.id,
+                                        ast.arguments([], None, None, []),
+                                        [ast.Return(value=cst_value)], []))
+                    metadata.add(module_body[-1].body[0], metadata.StaticReturn())
+            else:
+                self.local_decl = self.passmanager.gather(LocalDecl, stmt,
```
serge-sans-paille: I don't understand this compared to using an existing analyse either :-/

pythran/cxxtypes.py
pbrunet: Yes it is. (I fixed it here too to use the failfast arg on my computer :) )

pythran/transformations/expand_globals.py
pbrunet: Not only, I should filter, import and functions (indeed, everything except assignement to ast.Name)

I agree it looks like an analyse. I was just not sure we should create an analyse for this "very simple" case.

pythran/backend.py
pbrunet: May be. But I really wonder if this commit should be done as generated code is never used...

pythran/transformations/expand_globals.py
serge-sans-paille: > Not only, I should filter, import and functions (indeed, everything except assignement to ast.Name)

So you have a very simple filter here, based on the xisting
globaldeclaration?

doc/INTERNAL.rst
serge-sans-paille: > Because i += 1 is a modification on globals which is not allowed anymore.
ok. Should we put this in a functiondef instead of top-level, for the
sake of the doc?

pythran/transformations/expand_globals.py
pbrunet: I don't understand what you don't understand.

doc/INTERNAL.rst
pbrunet: No because we want to test the __init__ function for top level stmt. If we put it in a function def, it doesn't make sens.

pythran/transformations/expand_globals.py
pbrunet: Yes, it could be done that way.

pythran/backend.py
serge-sans-paille: > May be. But I really wonder if this commit should be done as generated code is never used...
??

pythran/transformations/expand_globals.py
serge-sans-paille: > > +                self.local_decl = self.passmanager.gather(LocalDecl, stmt,
> 
> I don't understand what you don't understand.
I missed thegather call. forget about it.

doc/INTERNAL.rst
serge-sans-paille: > No because we want to test the __init__ function for top level stmt. If we put it in a function def, it doesn't make sens.
ok

pythran/backend.py
pbrunet: I did the modification to have  static value but while I was looking to the code, I notice that this code is always constant folded so this function is never used.

serge-sans-paille: > I did the modification to have  static value but while I was looking to the code, I notice that this code is always constant folded so this function is never used.

Even when 

```python
a = [1,2,3,4]

def foo(i):
    return a[i]
```

?

pbrunet: Yes, just like with

```Python
def foo():
    return [1, 2, 3]

def bar(i):
    print foo()[i]
```

serge-sans-paille:
@pbrunet I wonder what's the status on this one ^^!

pbrunet:
I will not have time for the implementation using real globals instead of function.
I think we can:
1) I fix this one with your current comment
2) You merge it even if it is not the most efficient
3) Someone code a better globals implementation. (Now? Or after the better typing?)

Do you agree?

serge-sans-paille:
On Sun, Jun 14, 2015 at 10:43:18AM -0700, pbrunet wrote:
> I will not have time for the implementation using real globals instead of function.
> I think we can:
> 1) I fix this one with your current comment
> 2) You merge it even if it is not the most efficient
> 3) Someone code a better globals implementation. (Now? Or after the better typing?)

If 1) includes the fix for the static storage, I agree with it.
I dislike 2) because the performance issue does not look minor to me.
I prefer 1) over 3)

:-)

pbrunet:
As we said, the static storage is useless because of constant folding for this function. (1, 2, 3 were more steps than choices :) )

serge-sans-paille:
Can't we prevent constant folding in that case?

```
a = [1,2,...,100000]

def foo(i):
    return a[i]
```

As far as I understand, it is turned into

```
def global_a():
    return [1,2,...,100000]

def foo(i):
    return [1,2,...,100000][i]
```

Which seems a common case to me, and very bad for performance.

(sorry if history repeats again, that's an undesirable property of
history :-/)

pbrunet:
According to the constant folding rules, this list is "too big to be constant folded" so it will work.

IMO, It doesn't look to be a common case. most of the time, const globals variables are : scalar (PI, EARTH_RADIUS, CELERITY, ....) or dict of value (to group variable by interest).

serge-sans-paille:
@pbrunet maybe I could rework this if you allow me to?

pbrunet:
I just don't know the status of this one. I don't know what should be done and what was wrong.

serge-sans-paille:
On Sun, Dec 27, 2015 at 04:57:03AM -0800, pbrunet wrote:
> I just don't know the status of this one. I don't know what should be done and what was wrong.

If I recall correctly, I disliked (and still do) the fact that you
convert


a = [1, 2, 3,....., 10000]

def foo(i):
    return a[i]

into

def a():
    return [1, 2, 3,....., 10000]

def foo(i):
    return a()[i]

because this would lead to a multiple creattion of a big array.

The solution would be to generate

def a(l=[1, 2, 3,....., 10000]):
    return l

but this is not correctly converted by pythran, so let this be my next
task: correctly handle default parameter when it's an object!

pbrunet:
In fact, I changed the backend to generate static variables so the
performance penalty is not a problem and there are not multiple creation of
big array.

2015-12-27 18:51 GMT+01:00 serge-sans-paille <notifications@github.com>:

serge-sans-paille:
On Mon, Dec 28, 2015 at 03:34:13AM -0800, pbrunet wrote:
> In fact, I changed the backend to generate static variables so the
> performance penalty is not a problem and there are not multiple creation of
> big array.
ok, I'll review that (there seems to need a rebase though)

PR #447, opened 2015-04-27
Minimalist install

serge-sans-paille:
- remove complex hack dependency
- remove optional tcmalloc dependency
- document the pythranrc customization steps


doc/MANUAL.rst
```diff
@@ -327,6 +323,81 @@ glue. To do so use the ``-e`` switch. It will turn the python code into c++
...
+
+:``cxx``:
+
+    Path to the compiler to use
+
+:``cppflags``:
+
+    Additionnal preprocessor flags (``-I``, ``-D``, ``-U``). Pythran is
+    sensible to ``-DUSE_BOOST_SIMD`` and ``-DUSE_GMP``. The former turns on
+    Boost.simd vectorization and the latter turns on Multi-Precision Integers.
```
pbrunet: As DUSE_GMP is the default, can we disable this feature?
Also, should we document here : PYTHRAN_OPENMP_MIN_ITERATION_COUNT ?

pythran/toolchain.py
```diff
@@ -76,8 +76,10 @@ def _pythran_cppflags():
 
     def get(*x):
         return '-I' + os.path.join(curr_dir, *x)
-    return [get('.'), get('pythran'),
-            get('pythran', 'pythonic', 'patch')]
+    flags = [get('.'), get('pythran')]
+    if cfg.get('pythran', 'complex_hook'): # the patch is *not* portable
```
pbrunet: I think you should use "cfg.getbool". Otherwise it is almost always True.

doc/MANUAL.rst
```diff
@@ -327,6 +323,81 @@ glue. To do so use the ``-e`` switch. It will turn the python code into c++
...
+
+``[pythran]``
+=============
+
+This one contains internal configuration settings. Play with it at your own risk!
+
+:``optimizations``:
+
+    A list of import path pointing to transformation classes. This contains the
+    optimization pipeline of Pythran! I f you design your own optimizations,
```
pbrunet: s/I f/If

doc/MANUAL.rst
serge-sans-paille: > As DUSE_GMP is the default, can we disable this feature?

Yes, I suppressed this info.
> Also, should we document here : PYTHRAN_OPENMP_MIN_ITERATION_COUNT ?

correct!

pythran/toolchain.py
serge-sans-paille: > I think you should use "cfg.getbool". Otherwise it is almost always True.

Super-nice!

PR #449, opened 2015-05-04
Add random.shuffle(seq)

coyotte508:
I tried adding the other version, with a custom random number generator:

```cpp
    	template <class T, class function>
    	void shuffle(T & seq, function && randf) {
    		std::shuffle(seq.begin(), seq.end(), [&](int min, int max) -> int {
    			return (max - min) * randf() + min;
    		});
    	}
```

This works when calling it like that:

```python
import random
dests = range(64)

random.shuffle(dests, random.random)

print dests
```

but this fails:

```python
import random
dests = range(64)

def f():
    return 0.0

random.shuffle(dests, f)

print dests
```

So I didn't add the second version of `random.shuffle`.

### DIFF
--- a/pythran/pythonic/random/shuffle.hpp
+++ b/pythran/pythonic/random/shuffle.hpp
@@ -1,0 +1,48 @@
+#ifndef PYTHONIC_RANDOM_SHUFFLE_HPP
+#define PYTHONIC_RANDOM_SHUFFLE_HPP
+
+#include "pythonic/include/random/shuffle.hpp"
+
+#include "pythonic/utils/proxy.hpp"
+#include "pythonic/random/random.hpp"
+
+#include <limits>
+
+namespace pythonic {
+
+    namespace random {
+
+        template <class T>
+        void shuffle(T & seq) {
+            std::shuffle(seq.begin(), seq.end(), __random_generator);
+        }
+
+        namespace details {
+            template <class function>
+            struct URG {
+                URG(function && f) : randf(f) {}
+
+                typedef unsigned result_type;
+                result_type min() { return 0; }
+                /* -1 because of the floor() operation performed by the float->unsigned conversion */
+                result_type max() { return std::numeric_limits<result_type>::max() - 1;}
+                result_type operator()() {
+                    return randf() * std::numeric_limits<result_type>::max();
+                }
+
+                function randf;
+            };
+        }
+
+        template <class T, class function>
+        void shuffle(T & seq, function && randf) {    
+            std::shuffle(seq.begin(), seq.end(), details::URG<function>(std::forward<function>(randf)));
+        }
+
+        PROXY_IMPL(pythonic::random, shuffle)
+
+    }
+
+}
+
+#endif
--- a/pythran/tests/test_random.py
+++ b/pythran/tests/test_random.py
@@ -31,6 +31,15 @@ class TestRandom(TestEnv):
     def test_sample_(self):
         self.run_test("def sample_(n,k): from random import sample ; s = sum(sum(sample(range(n),k)) for x in range(n)) ; return abs(s/float(n*n)) < .05  ", 10**4, 4, sample_=[int, int])
 
+    def test_shuffle1(self):
+        self.run_test("def shuffle1(n): from random import shuffle ; r = range(n); shuffle(r); return len(r) == n and r != range(n) and sorted(r) == range(n)" , 10**4,  shuffle1=[int])
+
+    def test_shuffle2(self):
+        self.run_test("def shuffle2(n): from random import shuffle, random ; r = range(n); shuffle(r, lambda: 0); return len(r) == n and r != range(n)  and sorted(r) == range(n)" , 10**4,  shuffle2=[int])
+
+    def test_shuffle3(self):
+        self.run_test("def shuffle2(n): from random import shuffle, random ; r = range(n); shuffle(r, random); return len(r) == n and r != range(n)  and sorted(r) == range(n)" , 10**4,  shuffle2=[int])
+
     def test_choice(self):
         self.run_test("def choice_(n): from random import choice ; s= sum(choice(range(n)) for x in xrange(n)) ; return abs(s/n - n/2) < .05", 10**5,  choice_=[int])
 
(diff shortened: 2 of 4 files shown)


pbrunet:
Your implementation doesn't work because the third argument of std::shuffle need the interface:
http://en.cppreference.com/w/cpp/concept/UniformRandomNumberGenerator which is not provided by lambda functions. You may need to define a real functor to make it works.

You current code looks fine. It just required some tests.

coyotte508:
Ok, I fixed & added the tests.

pythran/pythonic/random/shuffle.hpp
```diff
@@ -0,0 +1,44 @@
...
+    namespace random {
+
+        template <class T>
+        void shuffle(T & seq) {
+            std::shuffle(seq.begin(), seq.end(), __random_generator);
+        }
+
+        template <class T, class function>
+        void shuffle(T & seq, function && randf) {
+            struct URG {
```
pbrunet: If you define the URG here, you will create a template version for each T type (so list<long> and list<float> will create 2 differentes URG version while they only depend on function).
Also, this is an implementation details so you should put it in a local "details" namespace and replace its call with something like
`details::URG<function>(std::forward<function>(randf))` (You may looks for details namespace in others part of the code).

pythran/pythonic/random/shuffle.hpp
```diff
@@ -0,0 +1,44 @@
...
+                URG(function && f) : randf(randf) {}
+
+                typedef unsigned result_type;
+                result_type min() { return 0; }
+                /* -1 because of the floor() operation in the operator () */
```
pbrunet: there is no "floor" anymore.

pythran/tests/test_random.py
```diff
@@ -31,6 +31,12 @@ def test_randint(self):
     def test_sample_(self):
         self.run_test("def sample_(n,k): from random import sample ; s = sum(sum(sample(range(n),k)) for x in range(n)) ; return abs(s/float(n*n)) < .05  ", 10**4, 4, sample_=[int, int])
 
+    def test_shuffle1(self):
+        self.run_test("def shuffle1(n): from random import shuffle ; r = range(n); shuffle(r); return len(r) == n and r != range(n)" , 10**4,  shuffle1=[int])
+
+    def test_shuffle2(self):
+        self.run_test("def shuffle2(n): from random import shuffle, random ; r = range(n); shuffle(r, lambda: random()); return len(r) == n and r != range(n);" , 10**4,  shuffle2=[int])
```
pbrunet: what if you use random directly? `shuffle(r, random)`

pythran/pythonic/random/shuffle.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+    namespace random {
+
+        template <class T>
+        void shuffle(T & seq) {
+            std::shuffle(seq.begin(), seq.end(), __random_generator);
+        }
+
+        template <class T, class function>
+        void shuffle(T & seq, function && randf) {
```
pbrunet: Can't you use the `std::function<T()>` type here?

(I don't know if it is possible because of functors, lambda and other advanced structures)

pythran/pythonic/random/shuffle.hpp
coyotte508: I meant the implicit truncation in the float-to-integer conversion. For example `5*0.99` will give `4`, so `max()` will be `5 - 1`.

pythran/pythonic/random/shuffle.hpp
coyotte508: Even if it is possible (and I'm not sure), doesn't a conversion to `std::function` create a slight overhead?

pythran/tests/test_random.py
coyotte508: It works too. I tried first lambda: 0, but both python and cpp implementations will differ (one seems to start from the beginning of the list and the other from the end), so I replaced 0 with `random()`, and forgot to remove the `lambda`.

pythran/pythonic/random/shuffle.hpp
pbrunet: I think no because it is just a pointer. There are no cast like for int/double.

pythran/tests/test_random.py
pbrunet: You should check for both : lambda and builtin function. Also, values should be check (to be sure you didn't remove/change some value). sorted can really help you here!!

pythran/pythonic/random/shuffle.hpp
coyotte508: I'm not sure. `std::function`'s usage is to store functions (in my opinion) but if you insist, I'll change it to that.

pythran/pythonic/random/shuffle.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+        template <class T>
+        void shuffle(T & seq) {
+            std::shuffle(seq.begin(), seq.end(), __random_generator);
+        }
+
+        namespace details {
+            template <class function>
+            struct URG {
+                URG(function && f) : randf(randf) {}
```
pbrunet: It should be randf(f) and not randf(randf)

pythran/tests/test_random.py
```diff
@@ -31,6 +31,15 @@ def test_randint(self):
...
         self.run_test("def sample_(n,k): from random import sample ; s = sum(sum(sample(range(n),k)) for x in range(n)) ; return abs(s/float(n*n)) < .05  ", 10**4, 4, sample_=[int, int])
 
+    def test_shuffle1(self):
+        self.run_test("def shuffle1(n): from random import shuffle ; r = range(n); shuffle(r); return len(r) == n and r != range(n) and sorted(r) == range(n)" , 10**4,  shuffle1=[int])
+
+    def test_shuffle2(self):
+        self.run_test("def shuffle2(n): from random import shuffle, random ; r = range(n); shuffle(r, lambda: 0); return len(r) == n and r != range(n)  and sorted(r) == range(n)" , 10**4,  shuffle2=[int])
+
+    def test_shuffle3(self):
+        self.run_test("def shuffle2(n): from random import shuffle, random ; r = range(n); shuffle(r, random; return len(r) == n and r != range(n)  and sorted(r) == range(n)" , 10**4,  shuffle2=[int])
```
pbrunet: a closing brace is missing.

serge-sans-paille:
@coyotte508 what's up with this one?

coyotte508:
@serge-sans-paille : it should be ok now, I hope

PR #451, opened 2015-05-11
Decl impl2

pbrunet:


pbrunet:
For review most commits are just "move" but from b2da3fd, I cherry-pick part of the numpy.dot PR.
I added commit  a9f2e90 to valid it on travis and change include directive generation in 1205b80

.clang-format
```diff
@@ -0,0 +1,65 @@
...
+SpaceAfterCStyleCast: false
+SpacesInContainerLiterals: true
+SpaceBeforeAssignmentOperators: true
+ContinuationIndentWidth: 4
+CommentPragmas:  '^ IWYU pragma:'
+ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
+SpaceBeforeParens: ControlStatements
+DisableFormat:   false
+...
+
```
serge-sans-paille: why not using one of the default setting?

serge-sans-paille:
Did you rebase this on against master?

.clang-format
pbrunet: It is the default I think. I dumped it so that it is stable for the future and easier to edit for our needs.

pbrunet:
Yes it is.

.clang-format
pbrunet: BTW, there are multiple default with clang-format. It could be: LLVM, google, chromium, mozilla or webkit. I choose LLVM I think.

PR #452, opened 2015-05-14
distutils

serge-sans-paille:


doc/CLI.rst
```diff
@@ -44,7 +44,7 @@ Pythran can also generate raw C++ code, using the ``-e`` switch::
 
   $> pythran -e cli_foo.py -o cli_foo.hpp
   $> printf '#include \"cli_foo.hpp\"\nusing namespace __pythran_cli_foo ; int main() { foo()(); return 0 ; }' > cli_foo.cpp
-  $> `pythran-config --compiler` cli_foo.cpp `pythran-config --cflags --libs` -o cli_foo
+  $> `pythran-config --compiler` -std=c++11 cli_foo.cpp `pythran-config --cflags --libs` -o cli_foo
```
pbrunet: Why do we need this? It is not the default in the configuration?

omp/__init__.py
```diff
@@ -7,10 +7,11 @@
 import sys
 
 try:
+    # there may be an environ modification when loading config
     from pythran import config
-    cxx = config.cfg.get('user', 'cxx')
 except ImportError:
-    cxx = 'c++'
+    pass
+cxx = os.environ.get('CXX', 'c++')
```
pbrunet: Is it the correct way to do?
I mean. Here is a "hidden" behavior on os.environ through the import. But why not just using the os.environ["CXX"] variable?
I would say that it is done that way because different compiler may use differente libgomp version but it is false as each compiler use its own OpenMP Implementation. (Clang is using libiomp5.so for example).
Here, we only want a path to an openmp library. We don't care about what is the version use. We don't care if it is the clang one or the gcc one, we just want an implementation to call it (without threads) in the Python world.
IMO, we should just use the os.environ without Pythran modification.

pythran/config.py
```diff
@@ -21,8 +24,48 @@ def init_cfg(sys_file, platform_file, user_file):
...
+        "extra_compile_args": cfg.get('compiler', 'cflags').split(),
+        "extra_link_args": cfg.get('compiler', 'ldflags').split(),
+    }
+
+    extension['define_macros'].append('ENABLE_PYTHON_MODULE')
+
+    here = os.path.dirname(os.path.dirname(__file__))
+    extension["include_dirs"].append(here + '/pythran')
+    for k, w in extra.items():
+        extension.setdefault(k, []).extend(w)
```
pbrunet: You may have not already present sections?

pythran/config.py
```diff
@@ -21,8 +24,48 @@ def init_cfg(sys_file, platform_file, user_file):
...
+        "library_dirs": cfg.get('compiler', 'library_dirs').split(),
+        "libraries": cfg.get('compiler', 'libs').split(),
+        "extra_compile_args": cfg.get('compiler', 'cflags').split(),
+        "extra_link_args": cfg.get('compiler', 'ldflags').split(),
+    }
+
+    extension['define_macros'].append('ENABLE_PYTHON_MODULE')
+
+    here = os.path.dirname(os.path.dirname(__file__))
+    extension["include_dirs"].append(here + '/pythran')
```
pbrunet: os.path would be better for future Windows support

pythran/config.py
```diff
@@ -21,8 +24,48 @@ def init_cfg(sys_file, platform_file, user_file):
...
+    for k, w in extra.items():
+        extension.setdefault(k, []).extend(w)
+    extension["define_macros"] = map(parse_define, extension["define_macros"])
+    if cfg.getboolean('pythran', 'complex_hook'):
+        # the patch is *not* portable
+        extension["include_dirs"].append(here + '/pythran/pythonic/patch')
```
pbrunet: os.path too

pythran/pythonic/types/slice.hpp
```diff
@@ -169,10 +169,10 @@ namespace pythonic {
...
                 if(upper.is_none) {
-                    assert(nt2::bitofsign(step) != nt2::bitofsign(lower));
+                    assert(nt2::bitofsign(step) != nt2::bitofsign((long)lower));
                     len = -(long)lower;
                 } else if(lower.is_none) {
-                    assert(nt2::bitofsign(step) = nt2::bitofsign(upper));
+                    assert(nt2::bitofsign(step) == nt2::bitofsign((long)upper));
                     len = upper;
                 } else
                     len = upper - lower;
```
pbrunet: Can you remove this commit?

pythran/tests/test_openmp.py
```diff
@@ -22,7 +22,7 @@ def interface(name, file=None):
     pythran.compile_cxxcode('#include <omp.h>', cxxflags=['-fopenmp'])
     TestOpenMP.populate(TestOpenMP)
     TestOpenMPLegacy.populate(TestOpenMPLegacy)
-except pythran.CompileError:
+except SystemExit:
```
pbrunet: SystemExit? It is raise by setup(..)?

pythran/toolchain.py
```diff
@@ -18,9 +18,10 @@
 from pythran.syntax import check_specs
 import pythran.frontend as frontend
 
-from numpy import get_include
+from numpy.distutils.core import setup
```
pbrunet: Why numpy.xxx?

scripts/pythran
```diff
@@ -132,7 +126,7 @@ parser.add_argument('-O', dest='extra_Oflags', metavar='level',
                     action='append',
                     help='any optimization level relevant to the underlying '
                     'C++ compiler',
-                    default=['2'])
+                    default=list())
```
pbrunet: Default is O0?

scripts/pythran-config
```diff
@@ -21,14 +26,28 @@ args = parser.parse_args(sys.argv[1:])
...
+    def fmt_define(define):
+        name, value = define
+        if value is None:
+            return '-D' + name
+        else:
+            return '-D' + name + '=' + value
+    output.extend(fmt_define(define) for define in extension['define_macros'])
+    output.extend(('-I' + include) for include in extension['include_dirs'])
+    output.append('-I' + numpy.get_include())
+    output.append('-I' + distutils.sysconfig.get_python_inc())
```
pbrunet: It is not the default for distutils.Extension?

scripts/pythran-config
```diff
@@ -21,14 +26,28 @@ args = parser.parse_args(sys.argv[1:])
...
 
 if args.libs:
-    output.extend(tc.ldflags())
+    output.extend(('-L' + include) for include in extension['library_dirs'])
+    output.append('-L' + distutils.sysconfig.get_config_var('LIBPL'))
+    output.extend(('-l' + include) for include in extension['libraries'])
+    output.append('-lpython' + distutils.sysconfig.get_config_var('VERSION'))
```
pbrunet: This one too.

setup.py
```diff
@@ -1,6 +1,8 @@
 from __future__ import print_function
 from distutils.command.build import build
+from distutils.command.install import install
 from distutils.core import setup, Command
+from distutils.dir_util import mkpath
```
pbrunet: These new import looks useless.

pythran/toolchain.py
```diff
@@ -299,31 +228,44 @@ def compile_cxxfile(cxxfile, module_so=None, **kwargs):
...
+          # fake CLI call
+          script_name='setup.py',
+          script_args=['--verbose'
+                       if logger.isEnabledFor(logging.INFO)
+                       else '--quiet',
+                       'build_ext',
+                       '--build-lib', builddir,
+                       '--build-temp', buildtmp,
+                       ]
+          )
```
pbrunet: I think it would be better to keep raison CompilationError in case of error here as it is more explicite than SystemExit.

doc/CLI.rst
serge-sans-paille: I removed it because I don't want to force the mode. What if the guy
wants std=c++14, or if its compiler uses different flag for c++11
support (like windows) --cflags only outputs -D, -I

That's the way pkg-config works, at least

omp/__init__.py
serge-sans-paille: I agree with your arguments but disagree on the conclusion ;-)

The problem is "how can the user state which compiler I can use to find an openmp runtime".
CXX is not always defined, so it's not an option
c++ is not always a valid symlink, so it's not an option

The best way I had was to rely on pythranrc, or CXX, or c++ in that
order.

pythran/config.py
serge-sans-paille: > You may have not already present sections?
No. I'll update this.

pythran/config.py
serge-sans-paille: > os.path would be better for future Windows support

No. Quoting distutils' documentation

  include_dirs : [string]
    list of directories to search for C/C++ header files (in Unix
    form for portability)

pythran/config.py
serge-sans-paille: http://epydoc.sourceforge.net/stdlib/distutils.extension.Extension-class.html

but I admit it's tricky

pythran/pythonic/types/slice.hpp
serge-sans-paille: done.

pythran/tests/test_openmp.py
serge-sans-paille: > SystemExit? It is raise by setup(..)?

yes :/
they are calling... os.system!

pythran/toolchain.py
serge-sans-paille: On Fri, May 22, 2015 at 07:10:23AM -0700, pbrunet wrote:
> > @@ -18,9 +18,10 @@
> >  from pythran.syntax import check_specs
> >  import pythran.frontend as frontend
> >  
> > -from numpy import get_include
> > +from numpy.distutils.core import setup
> 
> Why numpy.xxx?

because it fixes a c++ warning, and includes numpy.get_include()

scripts/pythran
serge-sans-paille: On Fri, May 22, 2015 at 07:14:34AM -0700, pbrunet wrote:
> > @@ -132,7 +126,7 @@ parser.add_argument('-O', dest='extra_Oflags', metavar='level',
> >                      action='append',
> >                      help='any optimization level relevant to the underlying '
> >                      'C++ compiler',
> > -                    default=['2'])
> > +                    default=list())
> 
> Default is O0?

default is set by distutils to O2

scripts/pythran-config
serge-sans-paille: On Fri, May 22, 2015 at 07:16:10AM -0700, pbrunet wrote:
> >  
> >  if args.cflags:
> > -    output.extend(tc.cppflags() + tc.cxxflags())
> > +    def fmt_define(define):
> > +        name, value = define
> > +        if value is None:
> > +            return '-D' + name
> > +        else:
> > +            return '-D' + name + '=' + value
> > +    output.extend(fmt_define(define) for define in extension['define_macros'])
> > +    output.extend(('-I' + include) for include in extension['include_dirs'])
> > +    output.append('-I' + numpy.get_include())
> > +    output.append('-I' + distutils.sysconfig.get_python_inc())
> 
> It is not the default for distutils.Extension?

It is, but it's internal to distutils and I cannot manage to extract the
info from the Extension (it's probably built in the build_ext somewhere)

I hate this piece of code, but cannot find a bettre way (I tried!)

scripts/pythran-config
serge-sans-paille: > This one too.

I double-hate it :-)

setup.py
serge-sans-paille: > These new import looks useless.

It is!

pythran/toolchain.py
serge-sans-paille: > I think it would be better to keep raison CompilationError in case of error here as it is more explicite than SystemExit.
you mean catching SytemError and raising CompilationError?

scripts/pythran-config
pbrunet: Ho ok, it is because pythran-config is not used in the Extension. I think the distutils.sysconfig is part of the build_ext so it looks good to me.

doc/CLI.rst
pbrunet: Looking at the pythran.cfg file, -std=c++11 is already in the cflags sections and pythran-config is loading this information. Here is a duplication.
Users can change it updating the .pythranrc file.

omp/__init__.py
pbrunet: Hum, OK I though that c++ was always valide.

I looked at the code using the cxx compiler and finally notice that it does ... nothing.
If the commande return useful information, it means the library is in the LIBRARY_PATH or default search (this is exactly what it done by find_library).
Shouldn't we just remove this part?

pythran/config.py
pbrunet:
```Python
import posixpath
posixpath.join("lala", "lili")
``` in this case?

File: pythran/toolchain.py (hunk 13 above)
Comment (pbrunet) [inline pythran/toolchain.py]: Yes, It was done that way before catching CalledProcessError

Comment (pbrunet) [inline pythran/toolchain.py]: You may use : distutils.errors.CompileError if you prefer.

File: doc/CLI.rst (hunk 1 above)
Comment (serge-sans-paille) [inline doc/CLI.rst]: no, pythran-config does noit load the extra_compile_args, only defines
and include_dirs

File: pythran/config.py (hunk 4 above)
Comment (serge-sans-paille) [inline pythran/config.py]: you're petty aren't you ;-)
 I propose to leave it like this and when I'll do the windows port,
 change it if needed.

File: pythran/toolchain.py (hunk 13 above)
Comment (serge-sans-paille) [inline pythran/toolchain.py]: > You may use : distutils.errors.CompileError if you prefer.
Great! Thanks!

File: pythran/config.py (hunk 4 above)
Comment (pbrunet) [inline pythran/config.py]: I admit it is pedantic. It is just because when I see code like this, I think "Programmer was lazy so I will fix it and use os.path.join" While if you use posixpath, I know it is intensional and that it should not be changed.

**Comment by serge-sans-paille** [comment]
@pbrunet I am ready. No more dev until I read the decl impl PR :-)

File: pythran/tests/test_openmp.py [hunk 14]
```diff
@@ -19,10 +19,10 @@ def interface(name, file=None):
 
 # only activate OpenMP tests if the underlying compiler supports OpenMP
 try:
-    pythran.compile_cxxcode('#include <omp.h>', cxxflags=['-fopenmp'])
+    pythran.compile_cxxcode('#include <omp.h>', extra_compile_args=['-fopenmp'], extra_link_args=['-fopenmp'])
     TestOpenMP.populate(TestOpenMP)
     TestOpenMPLegacy.populate(TestOpenMPLegacy)
-except pythran.CompileError:
+except SystemExit:
```
pbrunet: CompileError here?

serge-sans-paille:
@pbrunet : Travis loooves me! I'm rebased!

PR #453, opened 2015-05-14
Make setup.py more portable

serge-sans-paille:
- do not rely on git to fetch nt2
- do not rely on make to build nt2

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,13 @@ from __future__ import print_function
 from distutils.command.build import build
 from distutils.core import setup, Command
 from subprocess import check_call, check_output
+from urllib import urlopen
+from zipfile import ZipFile
+from StringIO import StringIO
+
 import logging
 import os
+import re
 import shutil
 import sys
 import time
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -54,39 +58,40 @@ def build_ply(self):
...
-            check_call(cmd.split())
-        else:
-            cmd = 'git fetch && git checkout origin/release'
-            check_call(cmd, shell=True, cwd=nt2_dir)
-            print('nt2 git repository updated')
+        cwd = os.getcwd()
+        nt2_src_dir = os.path.join(cwd, self.build_temp, nt2_dir + '_src')
+        if not os.path.isdir(nt2_src_dir):
+            print('nt2 archive needed, downloading it')
+            url = 'https://github.com/NumScale/nt2/archive/release.zip'
```
pbrunet: So, how can we handle the "cherry pick" for blas here?

pbrunet: And how do we handle updates for this zipfile?

setup.py
```diff
@@ -54,39 +58,40 @@ def build_ply(self):
...
+            print('nt2 not configured, configuring it')
 
-            cwd = os.getcwd()
-            abs_nt2_dir = os.path.join(cwd, nt2_dir)
             os.chdir(nt2_build_dir)
             build_cmd = ['cmake',
-                         abs_nt2_dir,
-                         '-DCMAKE_INSTALL_PREFIX=.',
-                         '-DNT2_FIND_REPOSITORIES='
-                         'git://github.com/MetaScale/nt2-modules.git']
```
pbrunet: This "sub repo" is already included?

setup.py
serge-sans-paille: You would just point to the URL of your branch, like

    https://github.com/pbrunet/nt2/archive/gemv_release.zip

right?

serge-sans-paille: > And how do we handle updates for this zipfile?

We don't... Just ask for a new build, which is much faster now because
of faster download.

setup.py
serge-sans-paille: > This "sub repo" is already included?
It apperas it's no longer needed ;-)

serge-sans-paille:
@pbrunet : ready to go!

PR #459, opened 2015-06-02
pythran: change 'l0', 'l1' to 'arr0', 'arr1'

hainm:
they are like '10', '11'


serge-sans-paille:
@hainm thanks for the doc improvement!

hainm:
@serge-sans-paille the only easy I  think help now. :P 

I am trying `pythran` and if there is any thing that is not clear in doc, I can help with this (as a part of my learning).

Hai

PR #462, opened 2015-06-07
Itertools refactor

pbrunet:
Travis is almost happy (only formating failed, I will update it only if everything else is right to avoid extra travis validation)


pbrunet:
this one is OK on my travis.

PR #463, opened 2015-06-15
Feature/winpython support

serge-sans-paille:


serge-sans-paille:
Made obsolete by #469

PR #464, opened 2015-06-16
Adding blas dependency for Debian in README

artas360:
It looks like installing numpy does not install blas on Debian testing. Can anybody else confirm?


serge-sans-paille:
@pbrunet I read you had a conversation with @artas360 on this subject on the IRC. This looks OK to me, feel free to accept the PR w/o Travis validation.

PR #465, opened 2015-06-16
Changing numpy random generator & adding numpy.random.binomial

artas360:
Any remarks?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,7 @@ env:
     #  'TESTCASE=pythran/tests/test_numpy_func1.py',
     #  'TESTCASE=pythran/tests/test_numpy_func2.py',
     #  'TESTCASE=pythran/tests/test_numpy_func3.py',
+    #  'TESTCASE=pythran/tests/test_numpy_random.py',
     #  'TESTCASE=pythran/tests/test_numpy_ufunc_unary.py',
     #  'TESTCASE=pythran/tests/test_numpy_ufunc_binary.py',
     #  'TESTCASE=pythran/tests/test_openmp.py',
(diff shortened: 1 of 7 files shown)


pbrunet:
Nice job Sares :-)

pythran/pythonic/include/numpy/random/generator.hpp
```diff
@@ -0,0 +1,24 @@
...
+  {
+    namespace random
+    {
+      namespace details
+      {
+        // Found by reading Numpy source code RNG values
+        // in (random/mtrand/randomkit.c),
+        // but the random generator type is not explicitly mentioned.
+        using default_numpy_generator_t = std::mt19937;
+        default_numpy_generator_t generator;
```
serge-sans-paille: Do we really need to have the same random generator? I mean, what's the added value?

pythran/pythonic/numpy/random/binomial.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "pythonic/include/types/exceptions.hpp"
+
+namespace pythonic
+{
+  namespace numpy
+  {
+    namespace random
+    {
+
+      inline void parameters_check(double n, double p)
```
serge-sans-paille: this function should go in a ``details`` namespace

serge-sans-paille:
appart from the two remarks, this LGTM

artas360:
Not yet, I have modifications coming up, but I have ... mispaced them ... :/

pythran/pythonic/include/numpy/random/generator.hpp
artas360: Hum we probably don't have too, but it doesn't cost much. And it might avoid bad surprises for numpy users, no?
And according to cppreference.com
> default_random_engine 	implementation-defined

So I guess it wouldn't hurt to impose the same random generator whatever the platform. What do you think?

pbrunet: Looking at numpy.random, all of them share the same generator with a sequence of value (I guess they are seed).
Also, it will avoid extra object creation cost no?

serge-sans-paille: > So I guess it wouldn't hurt to impose the same random generator whatever the platform. What do you think?

ok ;-)

pbrunet:
This looks good to me except for the missing .travis.yml check

.travis.yml
```diff
@@ -60,6 +60,7 @@ env:
     #  'TESTCASE=pythran/tests/test_numpy_func1.py',
     #  'TESTCASE=pythran/tests/test_numpy_func2.py',
     #  'TESTCASE=pythran/tests/test_numpy_func3.py',
+    #  'TESTCASE=pythran/tests/test_numpy_random.py',
```
pbrunet: And in the non commented part? :)

artas360: Arf I didn't even check how the file worked -_-

serge-sans-paille: @artas360 ping us when the tests have been updated :-)

artas360: I'm pretty sure it is, I did that a couple of days ago. I though you would get a mail :/

So unless I forgot sthg (again :/) it should be good to merge.

serge-sans-paille:
Level Up!

PR #466, opened 2015-06-17
Fixing nt2 DL in setup.py

artas360:
urllib.urlopen is deprecated and wasn't working on my setup, switching to urllib2 fixes it.

Additionally In some cases (in urllib) the file downloaded wasn't the zip file but an error page (http 404). Adding a check in case this happens.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from distutils.command.build import build
 from distutils.command.install import install
 from distutils.core import setup, Command
 from subprocess import check_call, check_output
-from urllib import urlopen
+from urllib2 import urlopen
 from zipfile import ZipFile
 from StringIO import StringIO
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -86,6 +86,8 @@ def build_nt2(self):
             print('nt2 archive needed, downloading it')
             url = 'https://github.com/pbrunet/nt2/archive/gemv_release.zip'
             location = urlopen(url)
+            http_code_prefix = location.getcode() / 100
+            assert http_code_prefix not in [4, 5], "Failed to download nt2."
```
serge-sans-paille: The fix is great. But is assert the right way to go?

artas360: I chose an assert because there was one a few lines after :p
(and because it's short)
But I'm all ears

serge-sans-paille: > I chose an assert because there was one a few lines after :p
> (and because it's short)
> But I'm all ears

This was a real question :-) I'm OK with an assert. I'll accept this on
once Travis is OK

PR #468, opened 2015-07-24
Add __pythran__ global variable

serge-sans-paille:
It holds some compilation informations


pbrunet:
LGTM but I am waiting for travis. Did you check it on windows?

(I read that http://www.appveyor.com/ could be use like travis with windows. I will certainly port it but I didn't have much time on this these last few weeks)

serge-sans-paille:
travis is ok! No windows check.

PR #469, opened 2015-07-26
Feature/winpython support

pbrunet:
It is WIP but you can see my modifications for the future PR


README.rst
```diff
@@ -75,6 +75,36 @@ Using `yaourt`::
...
+
+2. Install Boost (http://www.boost.org/users/download/): Once the archive is
+   extracted, use the *WinPython Command Prompt* and run the following::
+
+        cd boost_1_58
+        .\bootstrap.bat
+        .\b2 toolset=gcc link=shared --with-python
+        xcopy /E boost E:\WinPython-32bit-2.7.9.5\python-2.7.9\include\boost
+        copy stage\lib E:\WinPython-32bit-2.7.9.5\python-2.7.9\libs
+        copy stage\lib E:\WinPython-32bit-2.7.9.5\python-2.7.9\DLLs
```
serge-sans-paille: are the copy still needed now that we use BOOST_ROOT ?

extensions/pythranmagic.py
```diff
@@ -1,37 +1,42 @@
...
 """
-#-----------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 # Copyright (C) 2010-2011, IPython Development Team.
 #
 # Distributed under the terms of the Modified BSD License.
 #
 # The full license is in the file COPYING.txt, distributed with this software.
-#-----------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
```
serge-sans-paille: ??

pythran/backend.py
```diff
@@ -132,11 +132,11 @@ class Cxx(Backend):
         no jump are requiered.
         (else in loop means : don't execute if loop is terminated with a break)
 
-    >>> import ast, passmanager
+    >>> import ast, passmanager, os
     >>> node = ast.parse("def foo(): print 'hello world'")
     >>> pm = passmanager.PassManager('test')
     >>> r = pm.dump(Cxx, node)
-    >>> print r
+    >>> print str(r).replace(os.sep, '/')
```
serge-sans-paille: mmmh, / is portable for include path, maybe we should just use it instead of path.join for includes?

pythran/config.py
```diff
@@ -46,18 +46,19 @@ def parse_define(define):
...
+        "library_dirs": map(str, cfg.get('compiler', 'library_dirs').split()),
+        "libraries": map(str, cfg.get('compiler', 'libs').split()),
+        "extra_compile_args": map(str, cfg.get('compiler', 'cflags').split()),
+        "extra_link_args": map(str, cfg.get('compiler', 'ldflags').split()),
     }
 
     extension['define_macros'].append('ENABLE_PYTHON_MODULE')
 
-    here = os.path.dirname(os.path.dirname(__file__))
+    here = os.path.dirname(os.path.dirname(__file__)) or '.'
```
serge-sans-paille: ?

pythran/config.py
```diff
@@ -69,8 +70,8 @@ def parse_define(define):
 
     # blas dependency
     numpy_blas = numpy_sys.get_info("blas")
-    extension['libraries'].extend(numpy_blas['libraries'])
-    extension['library_dirs'].extend(numpy_blas['library_dirs'])
+    extension['libraries'].extend(numpy_blas.get('libraries', []))
+    extension['library_dirs'].extend(numpy_blas.get('library_dirs', []))
```
serge-sans-paille: so we still have the blas issue, right?

pythran/pythonic/__builtin__/id.hpp
```diff
@@ -6,6 +6,12 @@
 #include "pythonic/types/long.hpp"
 #include "pythonic/utils/proxy.hpp"
 
+/*
+ * We use uintptr_t conversion because on windows 64 bits, sizeof(void*) == 8
+ * and sizeof(long) == 4. Because of this, void* to long is forbidden but
+ * void* -> uintptr_t -> long is allowed
```
serge-sans-paille: add that there's a loss of accuracy here

pythran/pythonic/include/math/hypot.hpp
```diff
@@ -4,6 +4,10 @@
 #include "pythonic/utils/proxy.hpp"
 #include <cmath>
 
+#undef hypot
+// This is a windows defined macro that clash with std::hypot and our hypot
+// function
+
```
serge-sans-paille: crazy. How is it handled in their implen of the stdc++ lib

extensions/pythranmagic.py
pbrunet: pep8

pythran/backend.py
pbrunet: May be but I don't want to start to many modification in this PR, otherwise, it will never be merged.

pythran/config.py
pbrunet: I will try to solve it. Once others things are OK, I will try to find a procedure to install blas on windows and use it.

pythran/pythonic/include/math/hypot.hpp
pbrunet: I have absolutly no idea :-)

README.rst
pbrunet: For now, I don't really use BOOST_ROOT so yes, it is still needed.
I though about your idea to remplace BOOST_ROOT at install time but I think it is tricky for users as if they just use : python setup.py install, it will not work.

serge-sans-paille: On Mon, Jul 27, 2015 at 02:10:28PM -0700, pbrunet wrote:
> > +=======
> > +
> > +Using WinPython (http://winpython.github.io), installed in ``E:\WinPython-32bit-2.7.9.5``
> > +
> > +1. Install the binary version of CMake (http://www.cmake.org/download), and
> > +   make sure to check to check the box to add cmake to the PATH.
> > +
> > +2. Install Boost (http://www.boost.org/users/download/): Once the archive is
> > +   extracted, use the *WinPython Command Prompt* and run the following::
> > +
> > +        cd boost_1_58
> > +        .\bootstrap.bat
> > +        .\b2 toolset=gcc link=shared --with-python
> > +        xcopy /E boost E:\WinPython-32bit-2.7.9.5\python-2.7.9\include\boost
> > +        copy stage\lib E:\WinPython-32bit-2.7.9.5\python-2.7.9\libs
> > +        copy stage\lib E:\WinPython-32bit-2.7.9.5\python-2.7.9\DLLs
> 
> For now, I don't really use BOOST_ROOT so yes, it is still needed.
> I though about your idea to remplace BOOST_ROOT at install time but I think it is tricky for users as if they just use : python setup.py install, it will not work.

It will not work on windows because nt2 will not find boost install
anyway, so BOOST_ROOT needs to be set on windows?

pbrunet: For the first install but others install don't need it, do they?

serge-sans-paille: > For the first install but others install don't need it, do they?

Correct!

serge-sans-paille:
@pbrunet : this PR is too big. I'll start a new PR to incorporate the pip part I wrote (It's independent from Windows)

serge-sans-paille:
@pbrunet : I'll wait until you properly squash or remove my few commits to validate this one.

pbrunet:
Updated. I only squashed and remove the last commit about boost hack for windows installation to keep it for later.

pythran/pythonic/os/path/join.hpp
```diff
@@ -32,11 +38,15 @@ namespace pythonic
...
         if (head[0] == '/')
-          buffer = head;
-        else if (not buffer or *buffer.rbegin() == '/')
-          buffer += head;
-        else
-          buffer += "/" + head;
+          buffer = std::forward<T>(head);
+        else if (not buffer or *buffer.rbegin() == OS_SEP or
+                 *buffer.rbegin() == '/')
+          // SG->PB: I don't understand why you still need this last check (*buffer.rbegin() == '/')
```
serge-sans-paille: I thought you removed this comment and replaced it by the proper comment?

pythran/pythonic/os/path/join.hpp
```diff
@@ -32,11 +38,15 @@ namespace pythonic
       void _join(types::str &buffer, T &&head, Types &&... tail)
       {
         if (head[0] == '/')
```
serge-sans-paille: why is this char still here and not turned into another check? BTW, in addition of #defining OS_SEP, you should add a function is_path_sep that takes into account the fact that windows can use both / and \ as path sep.

pbrunet: Because for windows, we should handle the C:/ and not the / which is not supported here. I didn't want to spend too much time on it so I don't support all these possibilities. Because of this, the is_path_sep is not really interesting.

pbrunet:
It is OK now :-) !!!!

serge-sans-paille:
super great!

Issue #473, opened 2015-08-24
failing print


serge-sans-paille:
the following snippet fails to compile:

```python
#pythran export testPrint(int, float [])
def testPrint(n,a):
    i=0
    for i in xrange(0,n):
        for j in xrange(0,n):
            print(i,a[i])
#            print("i:",a[i])
```

serge-sans-paille:
Got it, the stream operator was declared in the wrong namespace, fix coming soon.

PR #474, opened 2015-08-24
Fix stream operator for std::tuple

serge-sans-paille:
the error lied in a bad namespace declaration, this fixes Issue #473.


serge-sans-paille:
@pbrunet this one is ready for merge!

PR #476, opened 2015-08-25
Fix bug in some numpy functions when passed scalar argument

### LINKED ISSUE #457: implement numpy.matrix, numpy.mat

serge-sans-paille:
This fixes #457

### DIFF
--- a/pythran/pythonic/include/numpy/all.hpp
+++ b/pythran/pythonic/include/numpy/all.hpp
@@ -12,13 +12,19 @@ namespace pythonic
   namespace numpy
   {
     template <class E>
-    bool _all(E begin, E end, utils::int_<1>);
-
-    template <class E, size_t N>
-    bool _all(E begin, E end, utils::int_<N>);
+    typename std::enable_if<types::is_numexpr_arg<E>::value, bool>::type
+    all(E const &expr, types::none_type _ = types::none_type());
 
     template <class E>
-    bool all(E const &expr, types::none_type _ = types::none_type());
+    typename std::enable_if<
+        std::is_scalar<E>::value or types::is_complex<E>::value, bool>::type
+    all(E const &expr, types::none_type _ = types::none_type());
+
+    template <class E>
+    auto all(E const &array, long axis) ->
+        typename std::enable_if<std::is_scalar<E>::value or
+                                    types::is_complex<E>::value,
+                                decltype(all(array))>::type;
 
     template <class E>
     auto all(E const &array, long axis) ->
--- a/pythran/pythonic/include/numpy/reduce.hpp
+++ b/pythran/pythonic/include/numpy/reduce.hpp
@@ -19,29 +19,6 @@ namespace pythonic
   namespace numpy
   {
 
-    template <class Op, size_t N, bool vector_form>
-    struct _reduce;
-
-    template <class Op, bool vector_form>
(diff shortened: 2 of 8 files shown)


pythran/pythonic/include/numpy/all.hpp
```diff
@@ -12,13 +12,19 @@ namespace pythonic
...
     template <class E>
-    bool _all(E begin, E end, utils::int_<1>);
+    typename std::enable_if<types::is_numexpr_arg<E>::value, bool>::type
+    all(E const &expr, types::none_type _ = types::none_type());
 
-    template <class E, size_t N>
-    bool _all(E begin, E end, utils::int_<N>);
+    template <class E>
+    typename std::enable_if<
+        std::is_scalar<E>::value or types::is_complex<E>::value, bool>::type
```
pbrunet: Don't we have a is_number for this?

pythran/pythonic/include/numpy/reduce.hpp
```diff
@@ -19,29 +19,6 @@ namespace pythonic
...
-      F operator()(E e, F acc);
-    };
-#endif
-
-    template <class Op, size_t N, bool vector_form>
-    struct _reduce {
-      template <class E, class F>
-      F operator()(E e, F acc);
-    };
-
```
pbrunet: Nice. For others classes, I moved this part in a "details" namespace. (Only available in the "src" file except for types)

pythran/pythonic/numpy/reduce.hpp
```diff
@@ -20,53 +20,61 @@ namespace pythonic
 
   namespace numpy
   {
+    template <class Op, size_t N, bool vector_form>
+    struct _reduce {
+      template <class E, class F>
+      F operator()(E e, F acc)
```
pbrunet: Why do we want a copy?

pythran/pythonic/include/numpy/all.hpp
serge-sans-paille: On Wed, Sep 02, 2015 at 11:24:03AM -0700, pbrunet wrote:
> >  
> > -    template <class E, size_t N>
> > -    bool _all(E begin, E end, utils::int_<N>);
> > +    template <class E>
> > +    typename std::enable_if<
> > +        std::is_scalar<E>::value or types::is_complex<E>::value, bool>::type
> 
> Don't we have a is_number for this?

This function does not exist, but it definitively should! I implemented
it as is_dtype

pythran/pythonic/numpy/reduce.hpp
serge-sans-paille: On Wed, Sep 02, 2015 at 11:29:49AM -0700, pbrunet wrote:
> > @@ -20,53 +20,61 @@ namespace pythonic
> >  
> >    namespace numpy
> >    {
> > +    template <class Op, size_t N, bool vector_form>
> > +    struct _reduce {
> > +      template <class E, class F>
> > +      F operator()(E e, F acc)
> 
> Why do we want a copy?
To enforce return by value and pass by value which makes it easier to
optimize by the compiler?

I recall a conversation with @joker-eph on the subject :-)

serge-sans-paille:
@pbrunet: I answered your question, it seems no change is needed (is_dtype is implemented in another PR)

PR #477, opened 2015-08-27
Python code cleaning

serge-sans-paille:
Some are serious:
- Use partial functions instead of lambdas referencing iterators in loop
  (that's probably a bug)
- Do not use list or sets as default parameters (that may lead to a bug)

Some are minors:
- Remove useless import
- Remove unused variables
- Remove lambda when the lambda body could be used instead

Some are cosmetic:
- Rename unused parameters
- Use isinstance instead of type

### DIFF
--- a/pythran/backend.py
+++ b/pythran/backend.py
@@ -239,7 +239,7 @@ class Cxx(Backend):
                 def __init__(self, s):
                     self.s = s
 
-                def generate(self, ctx):
+                def generate(self, _):
                     return self.s
 
             def __init__(self, other=None):
@@ -318,9 +318,11 @@ class Cxx(Backend):
                      if node in self.pure_expressions else EmptyStatement())
 
         def make_function_declaration(rtype, name, ftypes, fargs,
-                                      defaults=None, attributes=[]):
+                                      defaults=None, attributes=None):
             if defaults is None:
                 defaults = [None] * len(ftypes)
+            if attributes is None:
+                attributes = []
             arguments = list()
             for i, (t, a, d) in enumerate(zip(ftypes, fargs, defaults)):
                 if self.yields:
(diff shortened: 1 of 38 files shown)


pythran/config.py
```diff
@@ -1,7 +1,4 @@
-try:
-    import configparser
-except:
-    import ConfigParser as configparser
+import ConfigParser as configparser
```
pbrunet: No more Python3 compatibility? At least for the future?

pythran/optimizations/constant_folding.py
```diff
@@ -7,7 +7,6 @@
 from pythran.analyses.ast_matcher import DamnTooLongPattern
 
 import ast
-import numpy
```
pbrunet: I think numpy is required here. It is not in the code but eval may need to load it. (If I correctly remember... sorry for not documenting this ...)

pythran/backend.py
```diff
@@ -581,7 +583,7 @@ def visit_Return(self, node):
             stmt = ReturnStatement(self.visit(node.value))
             return self.process_omp_attachements(node, stmt)
 
-    def visit_Delete(self, node):
+    def visit_Delete(self, _):
```
pbrunet: It could be a staticmethod :-)

serge-sans-paille: > It could be a staticmethod :-)

Yes. let's not bother with this :-)

pythran/optimizations/constant_folding.py
serge-sans-paille: On Wed, Sep 02, 2015 at 11:39:51AM -0700, pbrunet wrote:
> I think numpy is required here. It is not in the code but eval may need to load it. (If I correctly remember... sorry for not documenting this ...)

I would say that it's already handled in the prepare method.
At least running the test_cases.py validation does not yield any
error...

pythran/config.py
serge-sans-paille: > No more Python3 compatibility? At least for the future?

I forgot why we had this one. I reverted the change and added a comment

serge-sans-paille:
@pbrunet: travis is ok now

PR #478, opened 2015-09-02
Bugfix/numpy improvements

serge-sans-paille:


serge-sans-paille:
@pbrunet: travis is ok now

pbrunet:
It is OK for me. If you want to change on my comment, do it :-) otherwise you can just merge it.

serge-sans-paille:
On Sat, Sep 05, 2015 at 11:30:24AM -0700, pbrunet wrote:
> It is OK for me. If you want to change on my comment, do it :-) otherwise you can just merge it.

I'll take into account your comment and use it in a few other places
too!

PR #479, opened 2015-09-02
Glimpses of Python3 support

serge-sans-paille:
Make it possible to compile pythran-generated code as a python3 native
module.


pythran/toolchain.py
```diff
@@ -169,8 +169,10 @@ def __str__(self):
                             )
         mod.add_to_includes(*map(Include, _extract_specs_dependencies(specs)))
         mod.add_to_includes(*content.body)
+
+        # the lambda is here to harmonize inconsistent return type between py2 and py3
         mod.add_to_init(
-            Line('#ifdef PYTHONIC_TYPES_NDARRAY_HPP\nimport_array()\n#endif'))
+            Line('#ifdef PYTHONIC_TYPES_NDARRAY_HPP\n[](){import_array(); return NUMPY_IMPORT_ARRAY_RETVAL;}();\n#endif'))
```
pbrunet: can't you just add a void cast before to say we ignore the result?

pythran/pythonic/types/long.hpp
```diff
@@ -87,14 +87,23 @@ namespace pythonic
     void *storage = ((boost::python::converter::rvalue_from_python_storage<
                          mpz_class> *)(data))->storage.bytes;
     auto s = PyObject_Str(obj_ptr);
+#if PY_MAJOR_VERSION >= 3
+    new (storage) mpz_class(PyUnicode_AS_DATA(s));
```
pbrunet: How can we still have mpz as input? You said there are no difference between int and long in Python3.

pythran/config.py
```diff
@@ -136,9 +135,6 @@ def fmt_define(define):
         output.append('-L' + distutils.sysconfig.get_config_var('LIBPL'))
         output.extend(('-l' + include)
                       for include in extension['libraries'])
-        output.extend(distutils.sysconfig.get_config_var('LIBS').split())
-        output.append('-lpython' +
-                      distutils.sysconfig.get_config_var('VERSION'))
```
pbrunet: You remove it here but you didn't add it in the default compilation. It won't compile no?

doc/CLI.rst
```diff
@@ -58,10 +71,9 @@ code can propagate constants using the Pythran ConstantFolding optimization::
...
   $> pythran -e cli_foo.py -p pythran.optimizations.ConstantFolding
 
-If you want to specify the path of generated file::
+If you're brave, you can generate native modules compatible with Python3. The
+hacky (and currently only) way to do so is by running::
 
-  $> pythran cli_foo.py -o /tmp/cli_foo.so -DNDEBUG
-  $> ls /tmp/cli_foo.so
-  /tmp/cli_foo.so
+  $> `pythran-config --compiler --cflags` `python3-config --cflags --libs` -lboost_python-py34 -lgmp -lgmpxx -std=c++11 cli_foo.cpp -shared -fPIC -o cli_foo -DNDEBUG
```
pbrunet: gmp is not in the pythran-config?

pythran/toolchain.py
serge-sans-paille: On Wed, Sep 02, 2015 at 11:54:53AM -0700, pbrunet wrote:
> >          mod.add_to_init(
> > -            Line('#ifdef PYTHONIC_TYPES_NDARRAY_HPP\nimport_array()\n#endif'))
> > +            Line('#ifdef PYTHONIC_TYPES_NDARRAY_HPP\n[](){import_array(); return NUMPY_IMPORT_ARRAY_RETVAL;}();\n#endif'))
> 
> can't you just add a void cast before to say we ignore the result?

No, because import_array is a macro

pythran/pythonic/types/long.hpp
serge-sans-paille: > How can we still have mpz as input? You said there are no difference between int and long in Python3.

When you turn a python2 code into a pythran3 native module (this patch
only allows generation of python3 module from python2 code)

pythran/config.py
serge-sans-paille: On Wed, Sep 02, 2015 at 11:57:43AM -0700, pbrunet wrote:
> > @@ -136,9 +135,6 @@ def fmt_define(define):
> >          output.append('-L' + distutils.sysconfig.get_config_var('LIBPL'))
> >          output.extend(('-l' + include)
> >                        for include in extension['libraries'])
> > -        output.extend(distutils.sysconfig.get_config_var('LIBS').split())
> > -        output.append('-lpython' +
> > -                      distutils.sysconfig.get_config_var('VERSION'))
> 
> You remove it here but you didn't add it in the default compilation. It won't compile no?

This code is only relevant to pythran-config. So any user of
pythran-config (=empty set) would need extra flags, but that's to cope
with boost python 2/3 difference

doc/CLI.rst
serge-sans-paille: On Wed, Sep 02, 2015 at 11:58:05AM -0700, pbrunet wrote:
> >  
> > -  $> pythran cli_foo.py -o /tmp/cli_foo.so -DNDEBUG
> > -  $> ls /tmp/cli_foo.so
> > -  /tmp/cli_foo.so
> > +  $> `pythran-config --compiler --cflags` `python3-config --cflags --libs` -lboost_python-py34 -lgmp -lgmpxx -std=c++11 cli_foo.cpp -shared -fPIC -o cli_foo -DNDEBUG
> 
> gmp is not in the pythran-config?
 pythran-config is called without --libs (on purpose)

serge-sans-paille:
green green green

hainm:
@serge-sans-paille I suppose 'Glimpses' mean I still need to wait, right? :P

serge-sans-paille:
@hainm it's actually far better than what I thought. You can generate C++ code with pythran2 and compile it into native code with pythran3, getting a python3 compatible native module.

hainm:
> @hainm it's actually far better than what I thought. 

:+1:

PR #480, opened 2015-09-10
Feature/numpy linalg norm

serge-sans-paille:


pbrunet:
I looked at numpy.linalg and the matrice case it not handled in the same path (that is why you see a wrong formula for norm). The correct path for matrice use `cond(...)` which almost always do : `norm(a, p) * norm(inv(a), p)`.
I don't know how good we are to compute `inv`. May be it is better than using the explicite formula?

pbrunet:
Fix : Forget my other message, the doc confuse me but it is the `norm` function which is called.

But, with axis=None (the default, axis is set to (0, 1) and then the path `if len(axis) == 2` is use which use the norm from the documentation.
I guess we should take some tests from this function documentation because tests pass for Pythran while formula looks incorrect.

serge-sans-paille:
On Tue, Sep 22, 2015 at 12:54:46PM -0700, pbrunet wrote:
> Fix : Forget my other message, the doc confuse me but it is the `norm` function which is called.
> 
> But, with axis=None (the default, axis is set to (0, 1) and then the path `if len(axis) == 2` is use which use the norm from the documentation.
> I guess we should take some tests from this function documentation because tests pass for Pythran while formula looks incorrect.

Currently, we simply don't support if len(axis) == 2, but I'll follow
your advice and add more test cases ;-)

serge-sans-paille:
ready for merging!

pbrunet:
Sorry, but I don't understand the last state of this MR. We saw that the implementation for ndarray::value == 2 is not correct but you don't want to fix it now. I totally agree with it as it required time for the new implementation but you don't raise any assert in case of 2D-ndarray so user may use this function and silently have an incorrect result and there are some tests with 2D-ndarray so if user look at test, he will really think 2D array are supported. I think it is really tricky!

PR #482, opened 2015-09-20
Fix/numpy gexpr and cie

serge-sans-paille:


serge-sans-paille:
ready for review!

serge-sans-paille:
travis is ok with this one too :-)

PR #485, opened 2015-09-25
Numpy random

pbrunet:


pythran/pythonic/include/numpy/random/choice.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+      template <class P>
+      types::ndarray<long, 1> choice(long max,
+                                     long size,
+                                     bool replace,
+                                     P && p);
+
+      template <class T>
+      auto choice(long max, T&& size) -> decltype(randint(0, max, size));
+
```
serge-sans-paille: missing std::forward<T>(size) here

pythran/pythonic/include/numpy/random/rand.hpp
```diff
@@ -0,0 +1,23 @@
...
+{
+  namespace numpy
+  {
+    namespace random
+    {
+      template <class... T>
+      types::ndarray<double, sizeof...(T)> rand(T... shape);
+
+      double rand();
+
```
serge-sans-paille: this conflicts with one of the commit of my previous PR... I'll prune it from my PR. (you knew it: you reviewed it ;-)

pythran/pythonic/include/numpy/random/standard_normal.hpp
```diff
@@ -0,0 +1,29 @@
...
+    namespace random
+    {
+      template <size_t N>
+      types::ndarray<double, N>
+      standard_normal(types::array<long, N> const &shape);
+
+      auto standard_normal(long size)
+          -> decltype(standard_normal(types::array<long, 1>{{size}}));
+
+      double standard_normal(types::none_type d = types::none_type());
```
serge-sans-paille: I think you can use the more concise notation ``types::none_type d = {}``

pythran/pythonic/numpy/random/bytes.hpp
```diff
@@ -0,0 +1,34 @@
...
+  namespace numpy
+  {
+    namespace random
+    {
+
+      types::str bytes(long length)
+      {
+        types::str result;
+        result.reserve(length);
+        std::uniform_int_distribution<char> distribution{0, 256};
```
serge-sans-paille: 256 -> 255! (bounds are included)

pythran/pythonic/numpy/random/bytes.hpp
```diff
@@ -0,0 +1,34 @@
...
+        for(long i=0; i<length; i++)
+          result.push_back(distribution(details::generator));
```
serge-sans-paille: std::generate ! (and it will be slightly faster than push_back), but you'll need to call resize instead of reserve.

pythran/pythonic/numpy/random/choice.hpp
```diff
@@ -0,0 +1,125 @@
...
+      {
+        return randint(max);
+      }
+
+      /***********************************************************
+       * Implementation with array as first argument
+       **********************************************************/
+
+      template <class T>
+      typename T::dtype choice(T const& a)
```
serge-sans-paille: This does not work with all numpy_expr, does it?

pythran/pythonic/numpy/random/choice.hpp
```diff
@@ -0,0 +1,125 @@
...
+        static_assert(T::value == 1, "ValueError: a must be 1-dimensional");
+
+        return a.fast(randint(a.size()));
+      }
+
+      template <class T, size_t S>
+      types::ndarray<typename T::dtype, S> choice(T const& a, types::array<long, S> const& shape)
+      {
+        types::ndarray<typename T::dtype, S> result{shape, types::none_type()};
+        std::uniform_int_distribution<long> distribution{0, a.size()};
```
serge-sans-paille: again, the bounds are inclusive -> ``a.size() - 1``

pythran/pythonic/types/str.hpp
```diff
@@ -447,6 +447,11 @@ namespace pythonic
       data->reserve(n);
     }
 
+    void str::push_back(char n)
+    {
+      data->push_back(n);
+    }
```
serge-sans-paille: If you're using std::generate, this one becomes useless

pythran/pythonic/numpy/random/bytes.hpp
pbrunet: Yes, So there will be an initialization for every values. I don't know what is the more efficient between one full useless initialisation and some push_back call with a reserve (and with this old computer, my measures are not very accurate). Did you do this measure?

pythran/pythonic/include/numpy/random/rand.hpp
pbrunet: Yes, I notice it after doing this implementation :-(

pythran/pythonic/numpy/random/choice.hpp
pbrunet: Yes, there is a test for it. What case doesn't work?

pythran/pythonic/numpy/random/bytes.hpp
serge-sans-paille: You're right, let's mesure!
```
    std::string s;
    s.resize(n + 1);
    for (int j = 0; j < n; ++j)
      s[j] = (j);
```
 runs almost three times faster than

```
    std::string s;
    s.reserve(n + 1);
    for (int j = 0; j < n; ++j)
      s.push_back(j);
```

on my laptop.

pythran/pythonic/numpy/random/choice.hpp
serge-sans-paille: Well, it will not compile for any numpy expression (including ndarray) that's not 1D, because ``a.fast(i)`` is somehow equivalent to ``a[i]`` without bound check.

pbrunet: This is not a problem as the parameter have to be a 1D array. (It is a requirement from Doc). I will add a static_assert.

pythran/pythonic/numpy/random/bytes.hpp
pbrunet: :) thx for this information. I didn't have such a big difference on my computer.

pythran/pythonic/numpy/random/bytes.hpp
```diff
@@ -0,0 +1,35 @@
...
+{
+  namespace numpy
+  {
+    namespace random
+    {
+
+      types::str bytes(long length)
+      {
+        // dummy init + rewrite is faster than reserve and push_back
+        types::str result(std::string(length, 0));
```
serge-sans-paille: great to keep the comment! In my memories, resizing is slightly different from initialization with a value as you do now, but I'm ok with this anyway.

serge-sans-paille:
Unfortunately clang-format fails.

PR #486, opened 2015-10-04
Remove boost.python dependency

serge-sans-paille:
- no more exported function returning void
- use direct calls to Python C API
- use a custom to_python/from_python interface (cleaner and faster!)
- move most of the dirty stuff to cxxgen.py

As a result, compilation runs faster, tests run faster... enjoy!


serge-sans-paille:
@pbrunet : enjoy!

serge-sans-paille:
travis is ok on this one ;-) ready 4 battle!

pythran/cxxgen.py
```diff
@@ -472,64 +476,237 @@ def generate(self):
...
+                static PyObject *
+                {ward}wrap_{name}(PyObject *self, PyObject *args)
+                {{
+                    return to_python({name}({args}));
+                }}'''
+
+        self.wrappers.append(
+            wrapper.format(name=func.fdecl.name,
+                           size=len(types),
+                           pyname=name,
```
pbrunet: pyname? It looks to be unused.

pythran/cxxgen.py
```diff
@@ -472,64 +476,237 @@ def generate(self):
...
+                           fmt="O" * len(types),
+                           objs=', '.join('&args_obj[%d]' % i
+                                          for i in range(len(types))),
+                           args=', '.join(args_unboxing),
+                           checks=' and '.join(args_checks),
+                           ward=pythran_ward,
+                           )
+        )
+
+        func_descriptor = pythran_ward + 'wrap_' + func.fdecl.name, types
```
pbrunet: If you define this name, you should use it in the previous template to be sure it is the same function name.

pythran/pythonic/__builtin__/id.hpp
```diff
@@ -25,11 +24,6 @@ namespace pythonic
       return t.id();
     }
 
-    long id(pythran_long_t const &t)
-    {
-      return reinterpret_cast<uintptr_t>(&t);
-    }
-
```
pbrunet: How do you handle gmp?

pythran/pythonic/__builtin__/set/difference_update.hpp
```diff
@@ -16,22 +16,27 @@ namespace pythonic
     {
 
       template <typename T, typename... Types>
-      void difference_update(types::set<T> &set, Types const &... others)
+      types::none_type difference_update(types::set<T> &set,
+                                         Types const &... others)
       {
         set.difference_update(others...);
+        return {};
```
pbrunet: Ha?

pythran/pythonic/include/utils/shared_ref.hpp
```diff
@@ -8,7 +8,6 @@
 #include <atomic>
 #endif
 #ifdef ENABLE_PYTHON_MODULE
-#include <boost/python/object.hpp>
```
pbrunet: ifdef are not needed anymore

pythran/pythonic/types/bool.hpp
```diff
@@ -3,4 +3,35 @@
...
+    }
+    static bool convert(PyObject *obj)
+    {
+      return PyInt_AsLong(obj);
+    }
+  };
+}
+
+#endif
+
```
pbrunet: Like for other, can you declare it in include files?

pythran/pythonic/types/complex.hpp
```diff
@@ -63,4 +63,38 @@ namespace pythonic
...
+      return PyComplex_Check(obj) or PyFloat_Check(obj) or PyInt_Check(obj);
+    }
+    static std::complex<T> convert(PyObject *obj)
+    {
+      if (PyComplex_Check(obj))
+        return {PyComplex_RealAsDouble(obj), PyComplex_ImagAsDouble(obj)};
+      else if (PyFloat_Check(obj))
+        return {PyFloat_AsDouble(obj), 0.};
+      else
+        return {(double)PyInt_AsLong(obj), 0.};
```
pbrunet: Here you check type twice. Once in `is_convertible` and another in `convert`. Why not using only `convert` and return nullptr when it fail?

pythran/pythonic/__builtin__/id.hpp
serge-sans-paille: > How do you handle gmp?

In the gmp -specific header, which removes a useless dependency

pythran/pythonic/types/dict.hpp
```diff
@@ -609,93 +609,58 @@ namespace std
...
-      registered = true;
-      boost::python::converter::registry::push_back(
-          &convertible, &construct,
-          boost::python::type_id<types::dict<K, V>>());
+  struct to_python<types::dict<K, V>> {
+    static PyObject *convert(types::dict<K, V> const &v)
+    {
+      PyObject *ret = PyDict_New();
+      for (auto kv = v.item_begin(); kv != v.item_end(); ++kv)
+        PyDict_SetItem(ret, ::to_python(kv->first), ::to_python(kv->second));
```
pbrunet: If it is thread-safe, it could be great to use OMP here :-)

pythran/pythonic/include/utils/shared_ref.hpp
serge-sans-paille: On Mon, Oct 05, 2015 at 12:50:02PM -0700, pbrunet wrote:
> > @@ -8,7 +8,6 @@
> >  #include <atomic>
> >  #endif
> >  #ifdef ENABLE_PYTHON_MODULE
> > -#include <boost/python/object.hpp>
> 
> ifdef are not needed anymore

They are needed for the (unused) pure C++ mode, where you d'ont want any
Python C API stuff

pythran/pythonic/types/dict.hpp
```diff
@@ -609,93 +609,58 @@ namespace std
...
-  }
+  struct from_python<types::dict<K, V>> {
 
-  PyObject *custom_empty_dict_to_dict::convert(types::empty_dict const &)
-  {
-    return PyDict_New();
-  }
+    static bool is_convertible(PyObject *obj)
+    {
+      if (PyDict_Check(obj)) {
```
pbrunet: You don't need the "__len__" check anymore?

pythran/pythonic/types/bool.hpp
serge-sans-paille: On Mon, Oct 05, 2015 at 12:54:00PM -0700, pbrunet wrote:
> > +  };
> > +  template <>
> > +  struct from_python<bool> {
> > +    static bool is_convertible(PyObject *obj)
> > +    {
> > +      return PyBool_Check(obj);
> > +    }
> > +    static bool convert(PyObject *obj)
> > +    {
> > +      return PyInt_AsLong(obj);
> > +    }
> > +  };
> > +}
> > +
> > +#endif
> > +
> 
> Like for other, can you declare it in include files?

The interface is not exported to other files (one always use the
to_python function instead) so... no?

pythran/pythonic/types/dict.hpp
```diff
@@ -609,93 +609,58 @@ namespace std
...
+        PyObject *key, *value;
+        Py_ssize_t pos = 0;
+        if (PyDict_Next(obj, &pos, &key, &value)) {
+          return ::is_convertible<K>(key) and ::is_convertible<V>(value);
```
pbrunet: You only check the first element and guess others have the same type?

pythran/pythonic/types/float.hpp
```diff
@@ -18,4 +18,57 @@ namespace pythonic
...
+      if (PyFloat_Check(obj))
+        return PyFloat_AsDouble(obj);
+      else
+        return PyInt_AsLong(obj);
+    }
+  };
+}
+
+#endif
+
```
pbrunet: Pas de support pour tous les types numpy?

pythran/pythonic/types/float.hpp
```diff
@@ -18,4 +18,57 @@ namespace pythonic
...
+    static PyObject *convert(float d)
+    {
+      return PyFloat_FromDouble(d);
+    }
+  };
+  template <>
+  struct from_python<double> {
+    static bool is_convertible(PyObject *obj)
+    {
+      return PyFloat_Check(obj) or PyInt_Check(obj);
```
pbrunet: Should we support complex with no imag part?

pythran/pythonic/types/int.hpp
```diff
@@ -15,4 +15,158 @@ long getattr<1>(long const &)
...
+      return PyInt_Check(obj);
+    }
+    static signed long convert(PyObject *obj)
+    {
+      return PyInt_AsLong(obj);
+    }
+  };
+}
+#endif
+
```
pbrunet: You could use some macro here?

pythran/pythonic/types/list.hpp
```diff
@@ -687,99 +687,47 @@ namespace std
...
-      registered = true;
-      boost::python::converter::registry::push_back(
-          &convertible, &construct, boost::python::type_id<types::list<T>>());
+  struct to_python<types::list<T>> {
+    static PyObject *convert(types::list<T> const &v)
+    {
+      Py_ssize_t n = v.size();
+      PyObject *ret = PyList_New(n);
+      for (Py_ssize_t i = 0; i < n; i++)
+        PyList_SET_ITEM(ret, i, ::to_python(v[i]));
```
pbrunet: no autofor here?

pythran/pythonic/types/set.hpp
```diff
@@ -493,7 +494,7 @@ namespace pythonic
     set<T> empty_set::operator^(set<T> const &s) { return s; }
 
     template <class... Types>
-    void empty_set::update(Types &&...)
+    none_type empty_set::update(Types &&...)
     {
```
pbrunet: no return stmt?

pythran/pythonic/types/set.hpp
```diff
@@ -519,91 +520,63 @@ namespace pythonic
...
-    static bool registered = false;
-    if (not registered) {
-      registered = true;
-      boost::python::converter::registry::push_back(
-          &convertible, &construct, boost::python::type_id<types::set<T>>());
+  struct to_python<types::set<T>> {
+    static PyObject *convert(types::set<T> const &v)
+    {
+      PyObject *obj = PySet_New(nullptr);
+      for (auto &&e : v)
```
pbrunet: This auto&& looks to be overkill :)

pythran/tests/cases/mandel.py
```diff
@@ -1,5 +1,5 @@
-#runas mandel(20,0,0, 8)
-#bench mandel(400,0,0, 75)
+#runas mandel(20,0.,0., 8)
+#bench mandel(400,0.,0., 75)
```
pbrunet: Typing is more strict?

pythran/toolchain.py
```diff
@@ -153,56 +153,21 @@ def __str__(self):
...
+                    'version': __version__,
+                    'date': datetime.now(),
+                    }
+
+        mod = PythonModule(module_name, docstrings, metainfo)
+        mod.add_to_preamble(Define("BOOST_SIMD_NO_STRICT_ALIASING", "1"))
         mod.add_to_includes(Include("pythonic/core.hpp"),
                             Include("pythonic/python/core.hpp"),
+                            Include("pythonic/types/bool.hpp"),
+                            Include("pythonic/types/int.hpp"),
```
pbrunet: Why are they required?

pythran/pythonic/__builtin__/id.hpp
pbrunet: I saw it later :-) great!!

pythran/pythonic/include/utils/shared_ref.hpp
pbrunet: I just mean that here, you have 
```
ifdef...
endif
```

WIth nothing inside

pythran/pythonic/types/complex.hpp
serge-sans-paille: > Here you check type twice. Once in `is_convertible` and another in `convert`. Why not using only `convert` and return nullptr when it fail?

I agree with the problem of the double check. But Your option would
imply either unnecessary conversions, or convoluted code, as in:

```
// we want to try two functions : f(set<int>, set<int>) or f (list<int>, list<int>)

if(set<int> arg0 = try_convert<set<int>>(py_arg0)) {
    if(set<int> arg1 = try_convert<set<int>>(py_arg1)) {
        return f0(arg0, arg1);
    }
    // arg0 will be destroyed here, so that's a potentially very costly
    // conversion that proved to be useless
}
else if (list<int> arg0 = try_convert<list<int>>(py_arg0)) {
    if(list<int> arg1 = try_convert<list<int>>(py_arg1)) {
        return f1(arg0, arg1);
    }  
    // arg0 will be destroyed here too
}
```

pythran/pythonic/types/dict.hpp
serge-sans-paille: > If it is thread-safe, it could be great to use OMP here :-)

:-) There's the GIL stuff

pythran/pythonic/types/dict.hpp
serge-sans-paille: > You only check the first element and guess others have the same type?

yes! Be brave :-)

pythran/pythonic/types/float.hpp
serge-sans-paille: > Pas de support pour tous les types numpy?

You mean int16 and the likes?

pythran/toolchain.py
serge-sans-paille: On Mon, Oct 05, 2015 at 01:29:32PM -0700, pbrunet wrote:
> >          mod.add_to_includes(Include("pythonic/core.hpp"),
> >                              Include("pythonic/python/core.hpp"),
> > +                            Include("pythonic/types/bool.hpp"),
> > +                            Include("pythonic/types/int.hpp"),
> 
> Why are they required?

bool: because it appeads as a result of comparison, as in

```
def foo(i, j):
    return i < j
```

which is not detected by our include generator. And because iwas lazy to
fix all the functions that return an int/bool without including this
headers...

pythran/tests/cases/mandel.py
serge-sans-paille: On Mon, Oct 05, 2015 at 01:26:39PM -0700, pbrunet wrote:
> > @@ -1,5 +1,5 @@
> > -#runas mandel(20,0,0, 8)
> > -#bench mandel(400,0,0, 75)
> > +#runas mandel(20,0.,0., 8)
> > +#bench mandel(400,0.,0., 75)
> 
> Typing is more strict?
 During my first tests yes, then no...

pythran/pythonic/types/dict.hpp
serge-sans-paille: Looks like it's ok, as validation pass. Seems like a boost-python specific issue?

pythran/pythonic/types/float.hpp
serge-sans-paille: That would lead to a lot of checks, I'd say no.

pythran/pythonic/types/dict.hpp
pbrunet: So, omp can improve the to_python conversion? Or there is the GIL everywhere (so it will be slow anyway...)

pythran/pythonic/types/float.hpp
pbrunet: int16 is "easy" but `numpy.int64` and `int` are sames in practice but have different types in Python.

pythran/toolchain.py
pbrunet: OK, can you just add a fixme? I will certainly continue on the include stuff so I may work on it once it is merged.

pythran/pythonic/types/complex.hpp
pbrunet: Hum OK, I understand the extra cost in these cases.

pythran/pythonic/types/dict.hpp
pbrunet: Great. It was not clear why we required it so it is better now.

pythran/pythonic/types/dict.hpp
serge-sans-paille: Will be slow (as far as I understand)

serge-sans-paille:
@pbrunet travis validation completed

serge-sans-paille:
@pbrunet I tokk into account your comment on headers, and updated .travis.yml to new architecture

pythran/pythonic/include/types/str.hpp
```diff
@@ -258,57 +258,4 @@ namespace std
...
-
-template <>
-struct __combined<char const *, pythonic::types::str> {
-  using type = pythonic::types::str;
-};
-
-template <>
-struct __combined<pythonic::types::str, char const *> {
-  using type = pythonic::types::str;
-};
```
pbrunet: There are no problems with these type inference removed?

pythran/pythonic/types/bool.hpp
pbrunet: I agree that it work that way for now but will it be always true?
I mean, if we correctly export sliced array. Will we use the generic `to_python/from_python` or will we need the specific one for ndarray in numpy_gexpr?
I just don't want to be the one doing this job next time.

pythran/pythonic/types/str.hpp
```diff
@@ -704,64 +704,63 @@ namespace std
...
+      return PyString_Check(obj);
+    }
+    static types::str convert(PyObject *obj)
+    {
+      return {PyString_AS_STRING(obj), (size_t)PyString_GET_SIZE(obj)};
+    }
+  };
+}
+/* type inference stuff  {*/
+#include "pythonic/include/types/combined.hpp"
```
pbrunet: type inference should be n includes.

pythran/pythonic/include/types/str.hpp
serge-sans-paille: seems not :-)

pythran/pythonic/types/bool.hpp
serge-sans-paille: This won't be an issue. The idea is that I export (in a pythonic/include/.../...hpp) the ``to_python()`` function and it gets specialized in the relevant header. If you want to call a specific ``to_python`` specialization, just call the good one and it will be automagically dispatched thanks to the template specialization mechanism. The only problem that could appear would be template use before specialization. We just need to respect the correct include order to by pass this.

The benfit ? cleaner include files as only the interface is exposed there. For instance sometimes the ``namespace impl`` stuff are also exported, which should be avoided (that's how I understand the ``impl`` namespace.)

pythran/pythonic/types/str.hpp
serge-sans-paille: Sorry I don't understand the comment.

pbrunet: This file is not in the include folder but type inference shoudl be there to compute correct function return types (Some use __combine in their types)

pythran/pythonic/types/bool.hpp
pbrunet: For namespace impl in header, It make sens to me when it is type related stuff (like return type for enumerate or other iterators) but I agree that others should not be in include files.

Also, I don't really like the idea of having a correct include order. We should include the dependencies in the file but not having an include order as it is the issue I tried to fix with the include split stuff. If it is not possible, I think the include declaration will have to be done (to_python and from_python are part of the interface as a converter interface. It is called from outside the class.)

pbrunet:
Looks good to me. I just wait for travis!!

serge-sans-paille:
nice, I split the implem / interface stuff, let's see if travis like this :-)

serge-sans-paille:
@pbrunet ok!

pbrunet:
Ready for a micro-release !!!

Issue #350, opened 2014-11-08
Support installation through conda


serge-sans-paille:
conda http://conda.pydata.org/docs/intro.html
is more and more used for non - developper scientific users. It would be nice to be able to install pythran through conda!

elyase:
+1

hainm:
@serge-sans-paille is there any progress with conda stuff? I had very hard time to install `pythran` from github or from pip. (pythran could not install `nt2` lib)

Hai

pbrunet:
It works for me with : `conda install -c https://conda.binstar.org/serge-sans-paille pythran`

I am not very familiar with conda but this looks to work for me.

hainm:
@pbrunet 

Hi,

I was able to install from `binstar` but still can not make `pythran` works. I got this kind of error
(tried `Basic Usage` from README file in `pythran`)

`fatal error: gmpxx.h: No such file or directory
 #include <gmpxx.h>`

And do you know why do we need to install more than 300 MB of `gcc_49-4.9.1`? It took really long time to download.

Hai

serge-sans-paille:
hi Hai,
the download of gcc_49 is here to make sure you have a recent compiler. The one bundled by conda by default (4.8) is far too old to support C++11 feature used by pythran.

Concerning gmp, that's not that much of a trouble, you can run ``conda install gmp`` that should install the dependencies.

Note that conda support for pythran is still in alpha stage, I'm very grateful you help to debug it! So if you still run into issues, please report them here !

I'll add gmp as a pythran dep in the coda description.

hainm:
Hi @serge-sans-paille 

I was able to pull `gmp` by `conda` but still getting the same error.

Below is the full error.

```
(python2)hnguyen@x~/programs/pythran/pythran/tests/euler (master) $ conda install gmp
Fetching package metadata: ........
Solving package specifications: .
Package plan for installation in environment /mnt/raidc2/haichit/anaconda3/envs/python2:

The following NEW packages will be INSTALLED:

    gmp: 5.1.2-2

Proceed ([y]/n)? y

Linking packages ...
[      COMPLETE      ]|####################################################################################################################| 100%
(python2)hnguyen@x~/programs/pythran/pythran/tests/euler (master) $ pythran euler40.py
WARNING: Disabling color, you really want to install colorlog.
CRITICAL: Leave the vessel! Women and children first!
C++ compiler failed to compile translated code.
E: Compile error!

******** Command line was: ********
[...]

PR #487, opened 2015-10-04
add conda build

hainm:
I am not sure if this https://github.com/serge-sans-paille/pythran/issues/350 is still valid, but I think install `pythran` from conda is the easiest way. 

I am myself not a fan of using 'sudo' since I am always working on shared cluster.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,6 +50,15 @@ Debian/Ubuntu
 		$> sudo apt-get update
 		$> sudo apt-get install pythran
 
+**or**
+
+1. Install ``conda``, following the instruction given in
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
nice! I'll do a few manual checks before merging though ;-)

hainm:
>  Are you using the conda packages extensively ?

@serge-sans-paille Yes, I am. For now, I only install from source code if I want to read the code and want to install the master branch. For the release, just conda.

I think conda is trending now. It's so easy to install numpy, scipy, cython, pandas, sklearn, matplotlib... with it. All within 1 minute.

PR #497, opened 2015-10-19
Fix/faster analyse

serge-sans-paille:


serge-sans-paille:
@pbrunet : travis likes me! merge me!

PR #498, opened 2015-10-30
Report a proper error when detecting compile time error like divide b…

serge-sans-paille:
…y zero

Commit inspired by the swift talk ;-)


serge-sans-paille:
@pbrunet : travis likes me! merge me!

PR #499, opened 2015-11-09
Simpler setup: bundle boost and nt2 deps

serge-sans-paille:
Bundle (part of) boost and full nt2 headers in third_parts directory for
easier install. No more trouble installing boost on windows etc, the
only absolute dependency of pythran is now a c++11 conforming compiler
\o/


kirbyfan64:
If you do this, my life will be complete. ;)

serge-sans-paille:
@kirbyfan64 bringing happiness to the world is our ultimate goal ;-)
What OS are you running?

kirbyfan64:
Me? Ubuntu 14.04 LTS. However, I seem to have the absolute worst luck on earth when trying to build stuff, *including* Pythran (which failed building nt2 with mysterious errors...).

hainm:
@serge-sans-paille 

I've tried your branch and seems a bit easier. This is what I did

$ conda --version
conda 3.18.4

-  make fresh env `conda create -n pythran python=2.7 numpy`
-  activate new env `source activate pythran`
- try pythran new's branch `git checkout feature/less-dependencies`
- install some packages `conda install decorator ply networkx`
- install colorlog `pip install colorlog` (seems that conda does not have colorlog)
- go back to pythran folder and `python setup.py install`
- update more packages for compiling.
    - gcc_499 `conda install -c serge-sans-paille gcc_49 gmp gmpxx`

The installation is much easier but the compiling to `.so` file is still difficult and I still need to spend much time to figure out why. The overall log is here

https://github.com/hainm/tipbag/issues/7

pbrunet:
It looks good to me. Can you provide your script to extract this part of boost? (for future update maybe?)

If we use std::unordered_map instead of boost::unordered_map, will it remove a lots of these file? Is it better?

serge-sans-paille:
On Wed, Nov 11, 2015 at 10:24:05AM -0800, pbrunet wrote:
> It looks good to me. Can you provide your script to extract this part of boost? (for future update maybe?)

It's already commited at the root of third_party/. You did not noticed
him among the 16k new files? How surprising!

> If we use std::unordered_map instead of boost::unordered_map, will it remove a lots of these file? Is it better?

The last benchmark I made (some 6+ month ago where in favor of
boost::unordered_map). The culprit here is the nt2 boost dependencies
(boost::fusion, boost::proto...)

serge-sans-paille:
On Tue, Nov 10, 2015 at 07:34:06AM -0800, Ryan Gonzalez wrote:
> Me? Ubuntu 14.04 LTS. However, I seem to have the absolute worst luck on earth when trying to build stuff, *including* Pythran (which failed building nt2 with mysterious errors...).

If current branch helps you, please tell us ;-) And more important, if
it doesn't, tell us too!

Issue #489, opened 2015-10-10
AttributeError: Distribution instance has no attribute 'exclude_package_data'


hainm:
```bash
$ python ./setup.py install
/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'entry_points'
  warnings.warn(msg)
/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'test_suite'
  warnings.warn(msg)
/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'install_requires'
  warnings.warn(msg)
/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'tests_require'
  warnings.warn(msg)
running install
running build
running build_py
Traceback (most recent call last):
  File "./setup.py", line 193, in <module>
    cmdclass={'build_py': BuildWithPly, 'test': PyTest})
  File "/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/command/install.py", line 563, in run
    self.run_command('build')
  File "/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/cmd.py", line 326, in run_command
    self.distribution.run_command(command)
  File "/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/command/build.py", line 127, in run
    self.run_command(cmd_name)
  File "/mnt/raidc2/haichit/anaconda3/envs/py2/lib/python2.7/distutils/cmd.py", line 326, in run_command
[...]

PR #508, opened 2015-11-29
Explicit setuptools dependency

serge-sans-paille:
Warn about missing setuptools during setup, see #489


serge-sans-paille:
Travis is ok with this one.

PR #511, opened 2015-12-03
Support multiline pythran spec

serge-sans-paille:
This makes it possible, in case of very long signature, to still follow
pep8 recommendation.

### DIFF
--- a/pythran/tests/test_spec_parser.py
+++ b/pythran/tests/test_spec_parser.py
@@ -50,3 +50,47 @@ class TestSpecParser(unittest.TestCase):
         code = '#pythran export bar(int)\ndef foo(): pass'
         with self.assertRaises(pythran.syntax.PythranSyntaxError):
             pythran.compile_pythrancode("dumber", code)
+
+    def test_multiline_spec0(self):
+        code = '''
+#pythran export foo(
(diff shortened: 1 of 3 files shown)


serge-sans-paille:
@pbrunet : this one is ok for review!

pythran/tests/test_spec_parser.py
```diff
@@ -50,3 +50,39 @@ def test_invalid_specs2(self):
...
+            '''
+        pythran.spec_parser(code)
+
+    def test_crappy_spec0(self):
+        code = '''
+#      pythran export foo(int) this is a pythran test
+def foo(i,j,k): return
+            '''
+        pythran.spec_parser(code)
+
```
pbrunet: Do we need a test with a comment on a next line? (I guess it is the same as crappy export)

serge-sans-paille: On Sat, Dec 05, 2015 at 10:41:09AM -0800, pbrunet wrote:
> Do we need a test with a comment on a next line? (I guess it is the same as crappy export)
Better safe than sorry, so let's add it :-)

serge-sans-paille:
ready!

PR #512, opened 2015-12-13
Feature/more syntax checks

serge-sans-paille:


serge-sans-paille:
I'm green badoudibadouda

PR #515, opened 2015-12-16
Install

pbrunet:


setup.py
```diff
@@ -126,22 +125,31 @@ def detect_gmp(self):
...
             objs = cc.compile(srcs)
             cc.link(ccompiler.CCompiler.EXECUTABLE,
                     objs, exe,
                     libraries=['gmp', 'gmpxx'])
-        except Exception:
+        except (LinkError, CompileError):
             # failure: remove the gmp dependency
             print('Failed to compile GMP source, disabling long support.')
-            for cfg in glob.glob(os.path.join(pythrandir, "pythran-*.cfg")):
+            for cfg in glob.glob(os.path.join("pythran", "pythran-*.cfg")):
```
lancelotsix: You should also updates the configurations that have already been copied in `os.path.join(self.build_lib, 'pythran')` (those that will be installed by `python setup.py install`)

PR #516, opened 2015-12-18
setup.py: fix error handeling when GMP is missing

lancelotsix:
For tests to run without gmp, original configuration files (i.e.
`pythran/pythran-*.cfg`) have to be modified, which is quite unsafe.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,6 +115,7 @@ class BuildWithThirdParty(build_py):
         print('Trying to compile GMP dependencies.')
 
         cc = ccompiler.new_compiler(verbose=False)
+        trash = list()
         # try to compile a code that requires gmp support
         with NamedTemporaryFile(suffix='.cpp', delete=False) as temp:
             temp.write('''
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
@lancelotsix : can you check if #515 fixes your issue? It has been reviewed and is almost ready to be merged!

lancelotsix:
OK, I'll do that

setup.py
```diff
@@ -125,23 +126,30 @@ def detect_gmp(self):
...
             print('Failed to compile GMP source, disabling long support.')
-            for cfg in glob.glob(os.path.join(pythrandir, "pythran-*.cfg")):
-                with open(cfg, "r+") as cfg:
-                    content = cfg.read()
-                    content = content.replace('USE_GMP', '')
-                    content = content.replace('gmp gmpxx', '')
-                    cfg.seek(0)
-                    cfg.write(content)
-        map(os.remove, objs + srcs + [exe])
+            for pythrandir in (os.path.join(self.build_lib, "pythran"),
```
serge-sans-paille: @pbrunet : I prefer this modification rather than modifying 'pythran'

setup.py
```diff
@@ -125,23 +126,30 @@ def detect_gmp(self):
...
+                               os.path.join(os.path.dirname(__file__),
+                                            "pythran")):
```
serge-sans-paille: @lancelotsix why do you modify the second version of the cfg. The one that lies in build should be enough, right?

lancelotsix: The configs in `self.build_lib` are staged for install while the one in `.` (the one updated in #515) are the ones that are loaded when you run `python setup.py test`. I think around here, you want to update both (even it is kind of a bad idea to update a version controlled file…).

PR #517, opened 2015-12-30
Fix enumerate over an ndarray corner case

serge-sans-paille:
As pointer out by a FIXME, the enumerate object was eventually holding a
dangling reference.

As a side effect remove an old iterator::reference artifact.

Added a test case from stack overflow pointing out the issue!

### DIFF
--- a/pythran/pythonic/__builtin__/enumerate.hpp
+++ b/pythran/pythonic/__builtin__/enumerate.hpp
@@ -30,8 +30,9 @@ namespace pythonic
       }
 
       template <class Iterator>
-      auto enumerate_iterator<Iterator>::operator*()
-          -> decltype(std::make_tuple(value, *iter))
+      typename enumerate_iterator_base<Iterator>::value_type
+          enumerate_iterator<Iterator>::
+          operator*()
       {
         return std::make_tuple(value, *iter);
       }
@@ -76,19 +77,15 @@ namespace pythonic
       }
 
       /// details::enumerate implementation
-
       template <class Iterable>
       enumerate<Iterable>::enumerate()
       {
       }
 
-      // FIXME : Here is a possible dangling reference.
-      // enumerate_iterator take seq.begin while the reference is kept with
-      // another seq.
       template <class Iterable>
       enumerate<Iterable>::enumerate(Iterable seq, long first)
-          : enumerate_iterator<typename Iterable::iterator>(seq.begin(), first),
-            seq(seq), end_iter(seq.end(), -1)
+          : Iterable(seq), iterator(Iterable::begin(), first),
+            end_iter(Iterable::end(), -1)
       {
       }
 
@@ -119,7 +116,6 @@ namespace pythonic
         typename std::remove_reference<Iterable>::type>::type>
     enumerate(Iterable &&seq, long first)
     {
-      // FIXME: Here we forward seq while enumerate signature required a copy
       return {std::forward<Iterable>(seq), first};
     }
 
--- a/pythran/pythonic/include/__builtin__/enumerate.hpp
+++ b/pythran/pythonic/include/__builtin__/enumerate.hpp
@@ -16,21 +16,18 @@ namespace pythonic
     {
       // FIXME return value may be a type::make_tuple
       template <class Iterator>
-      struct enumerate_iterator
-          : public std::iterator<
-                typename std::iterator_traits<Iterator>::iterator_category,
-                std::tuple<long,
-                           typename std::iterator_traits<Iterator>::value_type>,
-                ptrdiff_t, std::tuple<long, typename std::iterator_traits<
-                                                Iterator>::value_type> *,
-                std::tuple<long, typename std::iterator_traits<
-                                     Iterator>::value_type> /* not a ref */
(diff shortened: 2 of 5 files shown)


pythran/pythonic/include/types/nditerator.hpp
```diff
@@ -12,15 +12,9 @@ namespace pythonic
...
-              typename std::remove_reference<decltype(
-                  std::declval<E &>().fast(0))>::type,
-              ptrdiff_t, typename std::remove_reference<decltype(
-                             std::declval<E &>().fast(0))>::type *,
-              typename std::remove_reference<decltype(std::declval<E &>().fast(
-                  0))>::type /* no ref here, because we return a proxy object */
-              > {
+        : public std::iterator<std::random_access_iterator_tag,
+                               typename std::remove_reference<decltype(
+                                   std::declval<E &>().fast(0))>::type> {
```
pbrunet: It will not be a problem that the field "reference" is empty? In case the std do something with : A::iterator_trait::reference or something like this?

pythran/pythonic/include/__builtin__/enumerate.hpp
```diff
@@ -39,10 +35,14 @@ namespace pythonic
...
+      struct enumerate
+          : private Iterable, /* to hold a reference on the iterable */
+            public enumerate_iterator<
+                typename Iterable::iterator> /* to be compatible with
+                                                __builtin__.next*/
+            {
         using iterator = enumerate_iterator<typename Iterable::iterator>;
-        // we need to keep one ref over the enumerated sequence alive
-        Iterable seq;
+        using iterator::operator*;
```
pbrunet: Why do you need this operator?

pythran/pythonic/include/__builtin__/enumerate.hpp
```diff
@@ -14,23 +14,19 @@ namespace pythonic
 
     namespace details
     {
-      // FIXME return value may be a type::make_tuple
```
pbrunet: This one is still valid no?

pythran/pythonic/include/types/nditerator.hpp
serge-sans-paille: On Wed, Dec 30, 2015 at 12:08:42PM -0800, pbrunet wrote:
> It will not be a problem that the field "reference" is empty? In case the std do something with : A::iterator_trait::reference or something like this?

Yeah the default is to use value_type &

pythran/pythonic/include/__builtin__/enumerate.hpp
serge-sans-paille: On Wed, Dec 30, 2015 at 12:35:31PM -0800, pbrunet wrote:
> >          using iterator = enumerate_iterator<typename Iterable::iterator>;
> > -        // we need to keep one ref over the enumerated sequence alive
> > -        Iterable seq;
> > +        using iterator::operator*;
> 
> Why do you need this operator?

It's implicitly inherited from Iterable *and* enumerate_iterator. This
is the disambiguation.


We inherit from Iterable to store the original sequence and from
enumerate_iterator to be compatible with next(...)

pythran/pythonic/include/__builtin__/enumerate.hpp
serge-sans-paille: I thought the comment meant: beware, the return value may be different
from the type described by the  iterator trait. I fixed this somehow and
thought it was ok to remove the comment.

serge-sans-paille: Oh I got it! It's types::make_tuple vs. std::make_tuple! I'll leave the FIXME

serge-sans-paille:
@pbrunet everything should be ok now ;-)

pythran/pythonic/include/types/nditerator.hpp
pbrunet: Which is a dangling reference isn't it? As the iexpr is generated from the fast operator.

pythran/pythonic/__builtin__/enumerate.hpp
```diff
@@ -76,19 +77,15 @@ namespace pythonic
...
 
-      // FIXME : Here is a possible dangling reference.
-      // enumerate_iterator take seq.begin while the reference is kept with
-      // another seq.
       template <class Iterable>
       enumerate<Iterable>::enumerate(Iterable seq, long first)
-          : enumerate_iterator<typename Iterable::iterator>(seq.begin(), first),
-            seq(seq), end_iter(seq.end(), -1)
+          : Iterable(seq), iterator(Iterable::begin(), first),
+            end_iter(Iterable::end(), -1)
```
pbrunet: Nice trick to save value without attributs :-)

pbrunet:
There is just a question about "possible" dangling reference otherwise, LGTM.

pythran/pythonic/include/types/nditerator.hpp
serge-sans-paille: yes. the trick is that because we generate calls to auto && instead of auto now, they are no longer used

PR #518, opened 2016-01-05
0.7.4 - OpenDreamKit

serge-sans-paille:


serge-sans-paille:
@pbrunet this one is ok now, let's release!

PR #521, opened 2016-01-19
Support the astype method from numpy.ndarray

serge-sans-paille:


serge-sans-paille:
@pbrunet: it's to daddy to go!

PR #522, opened 2016-01-22
Support numpy.hstack and numpy.vstack

serge-sans-paille:
As a side effect, improve support for numpy.concatenate.


serge-sans-paille:
@pbrunet : I'm still working on this one!

serge-sans-paille:
@pbrunet ready for review now

serge-sans-paille:
@pbrunet updated, I took into account all your comment *but* the one on sets. We actually cannot create set of ndarray in pythran (because we use std::set and std::set rely on operator< that does not return a bool for ndarray)

serge-sans-paille:
@pbrunet : green is *the* color

PR #523, opened 2016-01-29
numpy.std implementation - not sa fast as I would like :-/

serge-sans-paille:


serge-sans-paille:
Still a WIP, there seems to be a performance issue on the axis version of numpy.mean / numpy.var and numpy.std as a consequence...

serge-sans-paille:
pbrunet : it's ok now

PR #524, opened 2016-01-29
Complete implementation of numpy.round and numpy.round_

serge-sans-paille:
They are explicitly described as alias to numpy.around, so why did we
reimplemented both versions o_O

### DIFF
--- a/pythran/pythonic/include/numpy/around.hpp
+++ b/pythran/pythonic/include/numpy/around.hpp
@@ -2,29 +2,35 @@
 #define PYTHONIC_INCLUDE_NUMPY_AROUND_HPP
 
 #include "pythonic/include/numpy/rint.hpp"
-#include <nt2/include/functions/pow.hpp>
+#include "pythonic/include/numpy/asarray.hpp"
 
 namespace pythonic
 {
 
   namespace numpy
   {
+    // fast path
+    template <class E>
+    auto around(E const &a) -> decltype(functor::rint{}(a));
 
     // generic floating point version, pure numpy_expr
     template <class E>
-    auto around(E const &a, long decimals = 0) -> typename std::enable_if<
-        !std::is_integral<typename E::dtype>::value,
-        decltype(functor::rint()(a *nt2::pow(typename E::dtype(10), decimals)) /
-                 nt2::pow(typename E::dtype(10), decimals))>::type;
+    auto around(E const &a, long decimals) -> typename std::enable_if<
+        !std::is_integral<typename types::dtype_of<E>::type>::value,
+        decltype(functor::rint{}(
+                     a *std::declval<typename types::dtype_of<E>::type>()) /
+                 std::declval<typename types::dtype_of<E>::type>())>::type;
 
     // the integer version is only relevant when decimals < 0
     template <class E>
-    auto around(E const &a, long decimals = 0) -> typename std::enable_if<
-        std::is_integral<typename E::dtype>::value,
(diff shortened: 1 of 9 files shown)


pbrunet:
LGTM, I just wonder why we can use around while others are in "non vectorizable" functions

serge-sans-paille:
On Sat, Jan 30, 2016 at 12:29:21AM -0800, pbrunet wrote:
> LGTM, I just wonder why we can use around while others are in "non vectorizable" functions
 Don't know :-)

serge-sans-paille:
@pbrunet ready!

pythran/pythonic/include/numpy/around.hpp
```diff
@@ -2,29 +2,48 @@
...
 {
 
   namespace numpy
   {
+    namespace details
+    {
+      template <class T>
+      struct is_integral_valued {
+        // what a trick :-) use sum to get the dtype even when T is a scalar
+        // type
```
pbrunet: Why?

serge-sans-paille: How do you now if a type, wich is a valid numexpr eargument, has an integral dtype.

there are two cases:
- it's a numpexpr or such, and in thatc ase it has a dtype
- it's a constant

The sum trick projects both types on a scalar, that's a funny trick

pbrunet: Ok, I get it. It is funny, but not really easy to understand IMHO.
Something like : 

```
using type = typename std::conditional<is_numexp<T>, T::dtype, T>::type;
static constexpr bool value = std::is_integral<type>::value;
```

Make it easier to me to understand this part.

serge-sans-paille: On Sun, Feb 14, 2016 at 01:38:30AM -0800, pbrunet wrote:
> >  
> >  namespace pythonic
> >  {
> >  
> >    namespace numpy
> >    {
> > +    namespace details
> > +    {
> > +      template <class T>
> > +      struct is_integral_valued {
> > +        // what a trick :-) use sum to get the dtype even when T is a scalar
> > +        // type
> 
> Ok, I get it. It is funny, but not really easy to understand IMHO.
> Something like : 
> 
> ```
> using type = typename std::conditional<is_numexp<T>, T::dtype, T>::type;
> static constexpr bool value = std::is_integral<type>::value;
> ```
> 
> Make it easier to me to understand this part.
This does not work, as both part of std::conditional must be valid and
int::dtype is not :-)

the right way to this would be to create a meta function dtype_of
specialized for scalar types, I'll go this way.

serge-sans-paille:
@pbrunet: dtype_of implemented!

pythran/pythonic/numpy/around.hpp
```diff
@@ -11,30 +13,42 @@ namespace pythonic
...
+    }
+    // list version
+    template <class E>
+    auto around(types::list<E> const &a, long decimals)
+        -> decltype(around(functor::asarray{}(a), decimals))
     {
-      return a /
-             (long)nt2::pow(typename E::dtype(10), std::max(0L, -decimals)) *
-             (long)nt2::pow(typename E::dtype(10), std::max(0L, -decimals));
+      return around(functor::asarray{}(a), decimals);
```
pbrunet: We can't do this for now but It could be greate to have an "asarray" function which do nothing for numpy expr so that we can call it in the "main" path

PR #525, opened 2016-02-03
Feature/variant functor

serge-sans-paille:


pythran/pythonic/types/variant_functor.hpp
```diff
@@ -0,0 +1,139 @@
...
+          -> decltype(std::declval<Type>()(std::forward<Args>(args)...))
+      {
+        return (*head)(std::forward<Args>(args)...);
+      }
+
+      template <class Type, class... Types>
+      template <class... OtherTypes>
+      variant_functor_impl<Type, Types...>::variant_functor_impl(
+          char mem[], OtherTypes const &... t)
+          : head(mem, t...), tail(mem, t...)
```
pbrunet: So a Type can never match multiple OtherTypes?

serge-sans-paille: You're dawn right! It could and I am unsure this breaks or not an invariant...

serge-sans-paille: I think in that case there will be two head that points to the functor, but that should be ok

PR #531, opened 2016-02-07
Fix distutils integration

serge-sans-paille:


LuisBL:
Just tested on a big project 
It's work like a charm !

PR #533, opened 2016-02-14
Extend numpy.lexsort to heterogeneous tuple

serge-sans-paille:


pythran/tests/test_numpy_func0.py
```diff
@@ -283,6 +283,9 @@ def test_lexsort0(self):
     def test_lexsort1(self):
         self.run_test("def np_lexsort1(a): from numpy import lexsort ; b = [9,4,0,4,0,2,1] ; return lexsort((a,b))", [1,5,1,4,3,4,4], np_lexsort1=[[int]])
 
+    def test_lexsort2(self):
+        self.run_test("def np_lexsort2(a): from numpy import lexsort ; return lexsort((a+1,a-1))", numpy.array([1,5,1,4,3,4,4]), np_lexsort2=[numpy.array([int])])
```
pbrunet: You should use the numpy example:

```
a = [1,5,1,4,3,4,4] # First column
b = [9,4,0,4,0,2,1] # Second column
ind = np.lexsort((b,a)) # Sort by a, then by b
```

Because with your values, the second-to-last order is the same as the last one.

serge-sans-paille: Done (for one of the examples)

PR #534, opened 2016-02-15
return alias draft support for user functions

serge-sans-paille:


pythran/analyses/aliases.py
```diff
@@ -128,6 +128,14 @@ def visit_Set(self, node):
         self.generic_visit(node)
         return self.add(node)  # not very accurate
 
+    def visit_Return(self, node):
+        if not node.value:
+            return
+        for alias in self.visit(node.value):
+            # FIXME: handle subscript
+            if type(alias) is ast.Name and type(alias.ctx) is ast.Param:
```
pbrunet: You should use "isinstance" to be consistent with other part of the code and use PEP8 recommendation.

pythran/analyses/aliases.py
```diff
@@ -245,6 +253,11 @@ def visit_FunctionDef(self, node):
                             for arg in node.args.args)
 
         self.generic_visit(node)
+        if '@' in self.aliases:
```
pbrunet: A global constant for this symbole may be better.

pythran/pythonic/__builtin__/list/append.hpp
```diff
@@ -23,6 +23,12 @@ namespace pythonic
         seq.push_back(std::forward<F>(value));
         return __builtin__::None;
       }
+      template <class T, class F>
+      types::none_type append(types::list<T> &&seq, F &&value)
+      {
+        seq.push_back(std::forward<F>(value));
+        return __builtin__::None;
+      }
```
pbrunet: Nice to have this feature. For information, I think some function doesn't have a correct implementation for rvalue (do nothing as we didn't plan this).

pythran/analyses/aliases.py
```diff
@@ -245,6 +253,11 @@ def visit_FunctionDef(self, node):
...
+            param_indices = {i for i in range(len(node.args.args))
+                             if node.args.args[i].id in self.aliases}
```
pbrunet: It looks like an enumerate :-)

pythran/analyses/aliases.py
```diff
@@ -245,6 +253,11 @@ def visit_FunctionDef(self, node):
...
+            node.return_alias = lambda node: {node.args[i]
+                                              for i in param_indices}
```
pbrunet: Just to be sure: You compute param_indices based on "node" but here, you use these indices based on another "node" (the lambda one, not the function one). Is it normal?

pythran/analyses/aliases.py
serge-sans-paille: 100% agree

pythran/analyses/aliases.py
serge-sans-paille: I used type because it's shorter and avoid more than 80 characters ;-)

You're definitevely right!

pythran/analyses/aliases.py
serge-sans-paille: Nice! (Brest! Saint Romain au mont d'or!)

pythran/analyses/aliases.py
serge-sans-paille: Yes, I changed the parameter name to avoid confusion

PR #541, opened 2016-02-29
Feature/better aliasing

serge-sans-paille:


serge-sans-paille:
@pbrunet ready for review ! enjoy ;-)

serge-sans-paille:
@pbrunet should be ok now

PR #542, opened 2016-03-05
Support cython-like array specification

serge-sans-paille:


serge-sans-paille:
@pbrunet this one is ready ;-)

pythran/spec.py
```diff
@@ -151,20 +152,19 @@ def p_type(self, p):
         '''type : term
                 | type LIST
                 | type SET
-                | type LARRAY RARRAY
-                | type LARRAY COLUMN COLUMN RARRAY
+                | type LARRAY array_indices RARRAY
                 | type COLUMN type DICT
                 | LPAREN types RPAREN'''
+        global _warned_obsolete_array_spec
```
pbrunet: What is it done for?

pbrunet:
LGTM, I just want to know what this global is done for :-)

pythran/spec.py
serge-sans-paille: nothing :-) At some point i added a warning to advocate for the cython
notation, then reverted everything but this line :-)

serge-sans-paille:
@pbrunet fixed

pbrunet:
Ok, thanks to this clarification, I notice that there is still an import too :-p. Do you prefer to merge anyway? :-)

serge-sans-paille:
On Thu, Mar 10, 2016 at 11:11:59AM -0800, pbrunet wrote:
> Ok, thanks to this clarification, I notice that there is still an import too :-p. Do you prefer to merge anyway? :-)

I'll remove it, no need to hurry :-)

PR #545, opened 2016-03-07
Ignore clang-format if not available

serge-sans-paille:


pbrunet:
Not sure we should  update this as it is still not in the stable debian release (OK, 3.6 is not either...)

serge-sans-paille:
ok.

PR #547, opened 2016-03-16
Feature/export global

serge-sans-paille:


serge-sans-paille:
@pbrunet this one is ready!

PR #549, opened 2016-03-19
None

pbrunet:


pythran/spec.py
```diff
@@ -221,6 +224,8 @@ def p_term(self, p):
         from numpy import uint8, uint16, uint32, uint64
 
         p[0] = eval(p[1])
+        if p[0] is None:
+            p[0] = type(None)
```
serge-sans-paille: you should update the doc about pythran annotation too

pythran/pythonic/__builtin__/dict/get.hpp
```diff
@@ -4,7 +4,7 @@
 #include "pythonic/include/__builtin__/dict/get.hpp"
 
 #include "pythonic/types/dict.hpp"
-#include "pythonic/types/none.hpp"
+#include "pythonic/types/NoneType.hpp"
```
serge-sans-paille: Why this change?

pbrunet: Because Pythran generate includes with ```pythonic/types/type(input_type).h```
Here , input_type for None is NoneType. I prefer have the same naming convention for every type than a special handling for None.

serge-sans-paille: ok great!

PR #551, opened 2016-03-20
Clean gen

pbrunet:
I don't really know if it should be integrated or not.

I didn't like the long generation of the same exception handling every time. It makes generated code harder to read (It is not the purpose of generated code but it is easier when we debug) and longer.

But while I changed this, I remove some ordering code for exception handling. It means ordering will have to be done manually when we will add new Exceptions (it is bad) but we don't add a lot of exception (not in the last 4 years)

### DIFF
--- a/pythran/cxxgen.py
+++ b/pythran/cxxgen.py
@@ -29,10 +29,7 @@ Generator for C/C++.
 
 from __future__ import division
 from textwrap import dedent
-import networkx as nx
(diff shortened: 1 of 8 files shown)


pbrunet:
@serge-sans-paille : Can you give me your opinion on this PR to know If I continue or not?

serge-sans-paille:
As long as the ordering stuff is documented, I'm ok with the change. I'll review it once it's green ;-)

PR #552, opened 2016-03-20
Add inlining optimisation.

pbrunet:
* Usefull to propagate more expression template.

I did some test with

```
#pythran export bar(float[])                                                     
import numpy                                                                     
                                                                                 
def foo(a):                                                                      
    return numpy.cos(a)                                                          
                                                                                 
def bar(c):                                                                      
    return foo(c) + c * foo(c)  
```

and I am 10% faster. (On my slow computer...)

### DIFF
--- a/pythran/analyses/inlinable.py
+++ b/pythran/analyses/inlinable.py
@@ -1,0 +1,29 @@
+""" Inlinable list function that may be inlined. """
+
+from pythran.passmanager import ModuleAnalysis
+from pythran.analyses import Identifiers
+
+import ast
+import copy
+
+
+class Inlinable(ModuleAnalysis):
+
+    """ Determine set of inlinable function.
+
+    A function can be inlined if it has only one statement and doesn't
+    recurse on itself.
+    """
+
+    def __init__(self):
+        self.result = dict()
+        super(Inlinable, self).__init__()
+
+    def visit_FunctionDef(self, node):
+        """ Determine this function definition can be inlined. """
+        if (len(node.body) == 1 and
+                isinstance(node.body[0], (ast.Call, ast.Return))):
+            ids = self.passmanager.gather(Identifiers, node.body[0], self.ctx)
+            # FIXME : It mark "not inlinable" def foo(foo): return foo
+            if node.name not in ids:
+                self.result[node.name] = copy.deepcopy(node)
--- a/pythran/optimizations/inlining.py
+++ b/pythran/optimizations/inlining.py
@@ -1,0 +1,108 @@
+""" Inlining inline functions body. """
+
+from pythran.analyses import Inlinable, Aliases
+from pythran.passmanager import Transformation
+
+import ast
+import copy
+
+
+class Inlining(Transformation):
+
+    """
+    Inline one line functions.
+
+    >>> import ast
+    >>> from pythran import passmanager, backend
+    >>> pm = passmanager.PassManager("test")
+    >>> node = ast.parse('''
+    ... def foo(a, b):
+    ...     return b + b * a
+    ... def bar(b):
+    ...     return foo(2 * b, b) * foo(b, b)''')
+    >>> _, node = pm.apply(Inlining, node)
+    >>> print pm.dump(backend.Python, node)
+    def foo(a, b):
+        return (b + (b * a))
+    def bar(b):
+        __pythran_inlinefooa0 = (2 * b)
+        __pythran_inlinefoob0 = b
+        __pythran_inlinefooa1 = b
+        __pythran_inlinefoob1 = b
+        return ((__pythran_inlinefoob0 + (__pythran_inlinefoob0 * \
+__pythran_inlinefooa0)) * (__pythran_inlinefoob1 + \
+(__pythran_inlinefoob1 * __pythran_inlinefooa1)))
+    """
+
+    def __init__(self):
+        """ fun : Function {name :body} for inlinable functions. """
+        self.update = False
+        self.defs = list()
+        self.call_count = 0
+        super(Inlining, self).__init__(Inlinable, Aliases)
+
+    def visit_Stmt(self, node):
+        """ Add new variable definition before the Statement. """
+        save_defs, self.defs = self.defs or list(), list()
+        self.generic_visit(node)
+        new_defs, self.defs = self.defs, save_defs
+        return new_defs + [node]
+
+    visit_Return = visit_Stmt
+    visit_Assign = visit_Stmt
+    visit_AugAssign = visit_Stmt
+    visit_Print = visit_Stmt
+    visit_For = visit_Stmt
+    visit_While = visit_Stmt
+    visit_If = visit_Stmt
+    visit_With = visit_Stmt
+    visit_Assert = visit_Stmt
+    visit_Expr = visit_Stmt
+
+    def visit_Call(self, node):
+        """
+        Replace function call by inlined function's body.
+
+        We can inline if it aliases on only one function.
+        """
+        func_aliases = self.aliases[node.func]
+        if len(func_aliases) == 1:
+            function_def = iter(func_aliases).next()
+            if (isinstance(function_def, ast.FunctionDef) and
+                    function_def.name in self.inlinable):
+                self.update = True
+                to_inline = copy.deepcopy(self.inlinable[function_def.name])
+                arg_to_value = dict()
+                values = node.args
+                values += to_inline.args.defaults[len(node.args) -
+                                                  len(to_inline.args.args):]
+                for arg_fun, arg_call in zip(to_inline.args.args, values):
+                    v_name = "__pythran_inline{}{}{}".format(function_def.name,
+                                                             arg_fun.id,
+                                                             self.call_count)
+                    new_var = ast.Name(id=v_name, ctx=ast.Store())
+                    self.defs.append(ast.Assign(targets=[new_var],
+                                                value=arg_call))
+                    arg_to_value[arg_fun.id] = ast.Name(id=v_name,
+                                                        ctx=ast.Load())
+                self.call_count += 1
+                return Inliner(arg_to_value).visit(to_inline.body[0])
+        return node
+
+
+class Inliner(ast.NodeTransformer):
+
+    """ Helper transform that perfomed inlined body transformation. """
+
+    def __init__(self, match):
+        """ match : {original_variable_name : Arguments use on call}. """
+        self.match = match
+        super(Inliner, self).__init__()
+
+    def visit_Name(self, node):
+        """ Transform name from match values if available. """
+        return self.match.get(node.id, node)
+
+    def visit_Return(self, node):
+        """ Remove return keyword after inline. """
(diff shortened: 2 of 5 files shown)


pythran/optimizations/inlining.py
```diff
@@ -0,0 +1,84 @@
...
+            * It is a one line function
+            * It is not recursive
+        """
+        for stmt in node.body:
+            if isinstance(stmt, ast.FunctionDef):
+                if len(stmt.body) == 1:
+                    ids = self.passmanager.gather(Identifiers, stmt.body[0],
+                                                  self.ctx)
+                    if stmt.name not in ids:
+                        self.fun[stmt.name] = stmt
```
serge-sans-paille: In the future, the notion of what is inlinable and what is not may evolve and grow more complex. I'll advocate for an Inlinable analysis separated from this transformation.

pythran/optimizations/inlining.py
```diff
@@ -0,0 +1,84 @@
...
+        return self.generic_visit(node)
+
+    def visit_Call(self, node):
+        """ Replace function call by inlined function's body. """
+        if isinstance(node.func, ast.Name) and node.func.id in self.fun:
```
serge-sans-paille: no no no, you should rely on aliasing, and not name id, to perform inlining. remember an identifier name has nothing to do with is actual value :-)

pythran/optimizations/inlining.py
```diff
@@ -0,0 +1,84 @@
...
+            self.update = True
+            to_inline = copy.deepcopy(self.fun[node.func.id])
+            arg_to_value = {a.id: b
+                            for a, b in zip(to_inline.args.args, node.args)}
+            return Inlineur(arg_to_value).visit(to_inline.body[0])
```
serge-sans-paille: what a frenchism « Inlineur » → Inliner ?

pythran/optimizations/inlining.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class Inlineur(ast.NodeTransformer):
+
+    """ Helper transform that perfomed inlined body transformation. """
+
+    def __init__(self, match):
+        """ match : {original_variable_name : Arguments use on call}. """
+        self.match = match
+        super(Inlineur, self).__init__()
```
serge-sans-paille: Ther's one additional issue here : what if the argument is used twice, and the effective argument is an expression , You should create a temporary variable and let forward substituion do the job

pythran/pythran.cfg
```diff
@@ -11,6 +11,7 @@ optimizations = pythran.optimizations.ForwardSubstitution
                 pythran.optimizations.PatternTransform
                 pythran.optimizations.Square
                 pythran.optimizations.RangeLoopUnfolding
+                pythran.optimizations.Inlining
```
serge-sans-paille: Maybe put it before forwrad substitution?

serge-sans-paille:
Great proposal! Inlining is a complex subject and you fell in a few booby traps ;-)
You're lacking a few test cases too?

serge-sans-paille:
btw, I foresee a cool article on this PR :-)

pythran/optimizations/inlining.py
pbrunet: I am pretty rusty.... :-)

pbrunet:
I am not sure I need more tests as recursion is already tested with many others cases and the basic transformation is tested with doctest.

I do the suggested modiciation but sadly ForwardSubstition doesn't really help here :s 

```
#pythran export bar(float[])                                                     
import numpy                                                                     

def foo(a):                                                                      
    return numpy.cos(a)                                                          

def bar(c):                                                                      
    return foo(c) + c * foo(c)  
```

In this case, bar will become  : 
```
def bar(c):
    _pythran_fooa0 = c
    _pythran_fooa1 = c
    return numpy.cos(_pythran_fooa0) + c * numpy.cos(_pythran_fooa1) 
```

ForwardSubstition can't substitute here as pythran_fooa0 alias on `c` so it is use 4 times (no FS if it is use more than once)
May be we should improve FS to foward with more than one "use" in the case of "simple variable assignment"?

serge-sans-paille:
I reply point-by-point


> I am not sure I need more tests as recursion is already tested with many others cases and the basic transformation is tested with doctest.

Right now your tests are very conservative. We may want to extend them
later on (eg to support more complex inlining situations)

> I do the suggested modiciation but sadly ForwardSubstition doesn't really help here :s 
> 
> ```
> #pythran export bar(float[])                                                     
> import numpy                                                                     
> 
> def foo(a):                                                                      
>     return numpy.cos(a)                                                          
> 
> def bar(c):                                                                      
>     return foo(c) + c * foo(c)  
> ```
> 
> In this case, bar will become  : 
> ```
> def bar(c):
>     _pythran_fooa0 = c
>     _pythran_fooa1 = c
>     return numpy.cos(_pythran_fooa0) + c * numpy.cos(_pythran_fooa1) 
> ```


The primary goal was to better handle

```
def foo(a):
    return coas(a) + sin(a)
```

when the call site is

```
foo(b +c)
```

And your proposal should already improve the situation.

> ForwardSubstition can't substitute here as pythran_fooa0 alias on `c` so it is use 4 times (no FS if it is use more than once)
> May be we should improve FS to foward with more than one "use" in the case of "simple variable assignment"?

Isn't this an error in our FS implementation? I see only one use of _pythran_fooa0...

pbrunet:
Indead it is not an implementation error but Lazyness analysis was done to
determine "lazy" types. In the given case, if pythran_fooa0 is a lazy type,
as it alias on c, c is a lazy type and they will be evaluated 4 times (it
is bad for lazy type like numpy_expr). So we assume that it is not lazy.
But it is based on lazy type which is stricter than FS. We should certainly
looks only for variable name instead of aliasing in this case.

2016-03-22 20:19 GMT+01:00 serge-sans-paille <notifications@github.com>:

pythran/analyses/inlinable.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self):
+        self.result = dict()
+        super(Inlinable, self).__init__()
+
+    def visit_FunctionDef(self, node):
+        """ Determine this function definition can be inlined. """
+        if (len(node.body) == 1 and
+                isinstance(node.body[0], (ast.Call, ast.Return))):
+            ids = self.passmanager.gather(Identifiers, node.body[0], self.ctx)
```
serge-sans-paille: This does not work in the (strange) case where
```
def foo(foo):
    return foo
```

pythran/optimizations/inlining.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __init__(self):
+        """ fun : Function {name :body} for inlinable functions. """
+        self.update = False
+        self.defs = list()
+        self.call_count = 0
+        super(Inlining, self).__init__(Inlinable, Aliases)
+
+    def visit_Stmt(self, node):
+        """ Add new variable definision before the Statement. """
```
serge-sans-paille: defini*t*ion

pythran/optimizations/inlining.py
```diff
@@ -0,0 +1,111 @@
...
+    visit_Return = visit_Stmt
+    visit_Assign = visit_Stmt
+    visit_AugAssign = visit_Stmt
+    visit_Print = visit_Stmt
+    visit_For = visit_Stmt
+    visit_While = visit_Stmt
+    visit_If = visit_Stmt
+    visit_With = visit_Stmt
+    visit_Assert = visit_Stmt
+    visit_Expr = visit_Stmt
```
serge-sans-paille: you could write `visit_Return = visit_Assign = ... = visit_Stmt`

pythran/optimizations/inlining.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def visit_Call(self, node):
+        """
+        Replace function call by inlined function's body.
+
+        We can inline if it alias on only one function.
```
serge-sans-paille: alias*es*

pythran/optimizations/inlining.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        func_aliases = self.aliases[node.func].aliases
+        if len(func_aliases) == 1:
+            function_def = iter(func_aliases).next()
+            if (isinstance(function_def, ast.FunctionDef) and
+                    function_def.name in self.inlinable):
```
serge-sans-paille: is the first check meaningful? i mean if we get, say, NewMem, it cannot be in self.inlinable, right? Add an assert if you're not sure !

pythran/optimizations/inlining.py
```diff
@@ -0,0 +1,111 @@
...
+                self.update = True
+                to_inline = copy.deepcopy(self.inlinable[function_def.name])
+                arg_to_value = dict()
+                values = node.args
+                values += to_inline.args.defaults[len(node.args) -
+                                                  len(to_inline.args.args):]
+                for arg_fun, arg_call in zip(to_inline.args.args, values):
+                    v_name = "_pythran_inline_{}{}{}".format(function_def.name,
```
serge-sans-paille: we generally use double underscore as a prefix, as it's *reserved for the compiler*

pythran/optimizations/inlining.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __init__(self, match):
+        """ match : {original_variable_name : Arguments use on call}. """
+        self.match = match
+        super(Inliner, self).__init__()
+
+    def visit_Name(self, node):
+        """ Transform name from match values if available. """
+        if node.id in self.match:
+            return self.match[node.id]
```
serge-sans-paille: `return self.match.get(node.id, node)`

pythran/analyses/inlinable.py
pbrunet: Yes, but it is still "safe". Should I check for name not in identifiers except if in args? Does it looks enough to you?

pythran/optimizations/inlining.py
pbrunet: I could but I am not sure it is "better", is it?

pythran/optimizations/inlining.py
pbrunet: My code say : If you have only one alias and it is inlinable , then... and you say : OK but with one alias, you can get not inlinable. It is true, that's why I check both. If you have multiple alias and all of them are inlinable, you can't inline. If you have one alias but not inlinable, you can't inline.

pythran/optimizations/inlining.py
pbrunet: yeah !!

pythran/analyses/inlinable.py
serge-sans-paille: Yes your solution is safe, just add a FIXME ;-)

pythran/optimizations/inlining.py
serge-sans-paille: I don't know. Do as you want!

serge-sans-paille:
@pbrunet still a test issue :-/

PR #553, opened 2016-03-20
Proper Syntax error over global redefinition

serge-sans-paille:


pythran/transformations/expand_globals.py
```diff
@@ -46,6 +48,10 @@ def visit_Module(self, node):
                 continue
             for target in stmt.targets:
                 assert isinstance(target, ast.Name)
+                if target.id in self.to_expand:
+                    raise PythranSyntaxError(
```
pbrunet: You raise an error? Not just a warning?

pbrunet: BTW, should we also check it with function with the same name?

serge-sans-paille: yes, taht's what we already do for function redefinition anyway.

serge-sans-paille: That's already done. in fact the code already rasied an error, but
blaming a function redefinition as globals are turned into functions.

PR #555, opened 2016-03-25
Fix/various fixes

serge-sans-paille:


serge-sans-paille:
Ready 4 review! I plan to make a blog article on the compiler flags part

PR #556, opened 2016-03-27
Better interprocedural typing

serge-sans-paille:
take into account returning multiple possible alias

### DIFF
--- a/pythran/analyses/aliases.py
+++ b/pythran/analyses/aliases.py
@@ -205,7 +205,9 @@ class Aliases(ModuleAnalysis):
             # expand collected aliases
             all_aliases = set()
             for value in return_alias:
-                if isinstance(value, NewMem):
+                # no translation
+                if isinstance(value, (NewMem, ContainerOf,
+                                      ast.FunctionDef, Intrinsic)):
                     all_aliases.add(value)
                 elif value in self.result:
                     all_aliases.update(self.result[value].aliases)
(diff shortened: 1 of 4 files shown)


serge-sans-paille:
This one is ready 4 review too :-)

pythran/analyses/aliases.py
```diff
@@ -207,6 +207,10 @@ def visit_Call(self, node):
             for value in return_alias:
                 if isinstance(value, NewMem):
                     all_aliases.add(value)
+                elif isinstance(value, ContainerOf):
+                    all_aliases.add(value)
+                if isinstance(value, (ast.FunctionDef, Intrinsic)):
```
pbrunet: elif?

pythran/analyses/aliases.py
```diff
@@ -207,6 +207,10 @@ def visit_Call(self, node):
...
+                    all_aliases.add(value)
```
pbrunet: There is a lot of case with the same code. Can't we factorize it?

pythran/tests/test_typing.py
```diff
@@ -66,6 +66,44 @@ def typing_aliasing_and_update():
...
         self.run_test(code, typing_aliasing_and_update=[])
 
+    def test_typing_aliasing_and_update_and_globals(self):
+        code = '''
+def f(l): return len(l)
+def g(l): l.append(1) ; return 0
+foo=[f,g]
+def typing_aliasing_and_update_and_globals(i):
+    h = []
+    return foo[i](h), h'''
```
pbrunet: There are no issue with order of evaluation for comma separated instruction?

pythran/analyses/aliases.py
serge-sans-paille: yes ^^!

pythran/analyses/aliases.py
serge-sans-paille: I thought so first, then I decided that handling may change in the
future. But you're right, the code will be cleaner.

pythran/tests/test_typing.py
serge-sans-paille: > There are no issue with order of evaluation for comma separated instruction?

gotcha! very smart one :-)

It turns out it's ok here because we're
handling references, so whatever the evaluation order, in the end we get an int and a reference to a list.

If we evaluate from left to right, we first update the list then return
a reference to this list.

if we evaluate from right to left we first return a reference then
evaluate then update this reference.

From a call site point of view, we get an int and a reference, and the
reference ultimately points to the updated list ;-)

serge-sans-paille:
ready to merge!

serge-sans-paille:
Included in merge PR #557

PR #566, opened 2016-05-10
Better integration with jupyter/ipython

serge-sans-paille:
The following sequence now works:

        %import pythran
        %load_ext pythran.magic
        %%pythran -O3 -fopenmp
        #pythran export foo(in)
        def foo(n):
                ...


serge-sans-paille:
fixed and ready to be merged!

PR #568, opened 2016-05-19
Take into account remarks on range by « Ron »

serge-sans-paille:


serge-sans-paille:
@pbrunet: ready to merge!

PR #570, opened 2016-05-25
Fix broadcasted iterators

serge-sans-paille:


pythran/pythonic/include/types/numpy_expr.hpp
```diff
@@ -107,7 +107,8 @@ namespace pythonic
       template <size_t I>
       bool _neq(numpy_expr_iterator const &other, utils::int_<I>) const
       {
-        return (std::get<I - 1>(iters_) != std::get<I - 1>(other.iters_)) ||
+        return (steps_[I - 1] &&
```
pbrunet: You don't have to check with other?

serge-sans-paille: > You don't have to check with other?
Isn't it recursive?

pbrunet: If I correctly understant, thow iterator differ if they "(first don't have step (broadcast) and value differ) or another sub iterator differ"

In the case where there are no step (it broadcast), why you don't have to check that broadcasted value differ with  second one?

serge-sans-paille: > If I correctly understant, thow iterator differ if they "(first don't have step (broadcast) and value differ) or another sub iterator differ"

Almost Correct! Let's call the underlying iterators the iterators on
each operand of the numpy_expr. Two numpy expr iterators differ if any
of their non-broadcasted underlying iterator differ. Values are not
involved.

> In the case where there are no step (it broadcast), why you don't have to check that broadcasted value differ with  second one?
See above : That's an iterator check, not a value check.

serge-sans-paille:
Fix are now ok!

PR #573, opened 2016-05-31
Fix type inference when assigning to a slice

serge-sans-paille:


serge-sans-paille:
@pbrunet ready to merge!

PR #574, opened 2016-06-01
Use std::abs for scalar version and not nt2's

serge-sans-paille:
NT2's is more accurate but soooo slow on complex


serge-sans-paille:
@pbrunet this one is ok!

PR #577, opened 2016-06-04
Fix import warning.

pbrunet:
* It was looking for builtin modules based on alias name which is incorrect.


serge-sans-paille:
Great!

PR #579, opened 2016-06-07
Numpy sum imap

pbrunet:
WIP : I have to add some tests and do the itertools to numpy_expr transformation for others iterators like ifilter

### DIFF
--- a/pythran/intrinsic.py
+++ b/pythran/intrinsic.py
@@ -156,9 +156,10 @@ class MethodIntr(FunctionIntr):
 
 class ConstMethodIntr(MethodIntr):
     def __init__(self, *combiners, **kwargs):
+        kwargs.setdefault('argument_effects',
+                          (ReadEffect(),) * 12)
         super(ConstMethodIntr, self).__init__(
-            *combiners,
-            argument_effects=(ReadEffect(),) * 12, **kwargs)
+            *combiners, **kwargs)
 
 
 class AttributeIntr(Intrinsic):
--- a/pythran/pythonic/include/__builtin__/xrange.hpp
+++ b/pythran/pythonic/include/__builtin__/xrange.hpp
@@ -51,6 +51,9 @@ namespace pythonic
       using const_iterator = xrange_iterator;
       using reverse_iterator = xrange_riterator;
       using const_reverse_iterator = xrange_riterator;
+      using dtype = value_type;
+      static constexpr bool is_vectorizable = false;
+      static constexpr size_t value = 1;
 
       long _begin;
       long _end;
--- a/pythran/pythonic/itertools/ifilter.hpp
+++ b/pythran/pythonic/itertools/ifilter.hpp
@@ -95,7 +95,8 @@ namespace pythonic
       template <typename Operator, typename List0>
       ifilter<Operator, List0>::ifilter(Operator _op, List0 const &_seq)
           : utils::iterator_reminder<false, List0>(_seq),
-            iterator(_op, this->value), end_iter(npos(), _op, this->value)
+            iterator(_op, utils::iterator_reminder<false, List0>::value),
+            end_iter(npos(), _op, utils::iterator_reminder<false, List0>::value)
       {
       }
 
--- a/pythran/tables.py
+++ b/pythran/tables.py
@@ -13,7 +13,7 @@ from pythran.intrinsic import ClassWithConstConstructor, ExceptionClass
 from pythran.intrinsic import ClassWithReadOnceConstructor
 from pythran.intrinsic import ConstFunctionIntr, FunctionIntr, UpdateEffect
 from pythran.intrinsic import ConstMethodIntr, MethodIntr, AttributeIntr
-from pythran.intrinsic import ReadEffect, ConstantIntr
+from pythran.intrinsic import ReadEffect, ConstantIntr, ReadOnceEffect
 from pythran.intrinsic import ReadOnceFunctionIntr, ConstExceptionIntr
 from pythran.types.conversion import PYTYPE_TO_CTYPE_TABLE
 from pythran import range as prange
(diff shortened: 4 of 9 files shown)


pythran/intrinsic.py
```diff
@@ -156,9 +156,10 @@ def isstaticfunction(self):
 
 class ConstMethodIntr(MethodIntr):
     def __init__(self, *combiners, **kwargs):
+        kwargs.setdefault('argument_effects',
+                          (ReadEffect(),) * 12)
         super(ConstMethodIntr, self).__init__(
-            *combiners,
-            argument_effects=(ReadEffect(),) * 12, **kwargs)
+            *combiners, **kwargs)
```
serge-sans-paille: why?

pythran/pythonic/include/__builtin__/xrange.hpp
```diff
@@ -51,6 +51,9 @@ namespace pythonic
       using const_iterator = xrange_iterator;
       using reverse_iterator = xrange_riterator;
       using const_reverse_iterator = xrange_riterator;
+      using dtype = value_type;
+      static constexpr bool is_vectorizable = false;
```
serge-sans-paille: could be true in the future

pythran/pythonic/itertools/ifilter.hpp
```diff
@@ -95,7 +95,8 @@ namespace pythonic
       template <typename Operator, typename List0>
       ifilter<Operator, List0>::ifilter(Operator _op, List0 const &_seq)
           : utils::iterator_reminder<false, List0>(_seq),
-            iterator(_op, this->value), end_iter(npos(), _op, this->value)
+            iterator(_op, utils::iterator_reminder<false, List0>::value),
```
serge-sans-paille: instead of true/false you could use two different empty structures with explicit names, that would make the review easier

pythran/tables.py
```diff
@@ -626,7 +626,8 @@ def update_effects(self, node):
         "std_": ConstMethodIntr(args=('a', 'axis', 'dtype'),
                                 defaults=(None, None)),
         "subtract": ConstFunctionIntr(),
-        "sum": ConstMethodIntr(),
+        "sum": ConstMethodIntr(argument_effects=(ReadOnceEffect(),
```
serge-sans-paille: why twice?

pythran/pythonic/itertools/ifilter.hpp
serge-sans-paille: ok it's not your code/fault :-) still a good idea if you're brave enough ;-)

pythran/tests/test_numpy_func1.py
```diff
@@ -42,6 +42,37 @@ def test_sum10_(self):
...
+                import numpy as np
+                from itertools import izip
+                return np.sum(izip(a, a))
+                      """, range(10), np_sum_izip=[[int]])
+
+    def test_sum_xrange(self):
+        self.run_test("""
+            def np_sum_xrange(a):
+                import numpy as np
+                return np.sum(xrange(a))
```
serge-sans-paille: BTW I add the case of arange being used instead of range, we could make arange lazy :-)

pythran/intrinsic.py
pbrunet: Because we don't want to overwrite the value if it is explicitly set

pythran/pythonic/itertools/ifilter.hpp
pbrunet: haha indeed I was the one introducing this struct :) but not in this commit.

pythran/tables.py
pbrunet: Because of the array value and axis value (the only arguments we support for sum)

pythran/tests/test_numpy_func1.py
pbrunet: Is it OK to keep this "not same as Pythoné i(map/filter/zip) as numpy_expr behavior?

pbrunet:
There is an issue with this PR as numpy.sum on generator return the same generator. Reason is that numpy expression are not Lazy in numpy as numpy need to know the size of the generator to allocate memory. For this reason, replacing list with generator will never work on numpy functions. Of course, we may still want to replace some numpy call on list to numpy call to generator to avoid memory allocation. 

I see 2 solution:
1) The filter-> ifilter transformation is perform after others (so after constant folding) and we accept to have non complient python at this stage. It could be done 
2) We may add an attributes to builtin like ReadOnlyNDArray and in this case, the filter can be transform to -> numpy.fromiter(ifilter(...)) but it supports only 1D array (which is most of the cases but not all) and we don't have the size information which may be required to create a numpy_expr.

Any better idea? Or should I just leave this PR as it can't be really applied?

serge-sans-paille:
On Mon, Jun 27, 2016 at 12:55:06PM -0700, pbrunet wrote:
> There is an issue with this PR as numpy.sum on generator return the same generator. Reason is that numpy expression are not Lazy in numpy as numpy need to know the size of the generator to allocate memory. For this reason, replacing list with generator will never work on numpy functions. Of course, we may still want to replace some numpy call on list to numpy call to generator to avoid memory allocation. 
> 
> I see 2 solution:
> 1) The filter-> ifilter transformation is perform after others (so after constant folding) and we accept to have non complient python at this stage. It could be done 
> 2) We may add an attributes to builtin like ReadOnlyNDArray and in this case, the filter can be transform to -> numpy.fromiter(ifilter(...)) but it supports only 1D array (which is most of the cases but not all) and we don't have the size information which may be required to create a numpy_expr.

Isn't this similar to saying that if we have a 1D array, we cann turn
numpy.sum into sum? Eitherway, we have no way to know at compile time
the dimension of an array. This looks dommed to me and I'b better say «
do nothing ». Do yiu have a test case that exhibits a ral use for this
one?

PR #580, opened 2016-06-08
Import-free setuptools integration

nils-werner:
This pull request implements an import free setuptools integration, [inspired by CFFI](http://cffi.readthedocs.io/en/latest/cdef.html#distutils-setuptools).

Regard the following example `setup.py`:

```
import setuptools
from pythran.dist import PythranExtension

if __name__ == "__main__":
    setuptools.setup(
        name='matmul',
        version='0.1',
        packages=setuptools.find_packages(exclude=['tests']),

        install_requires=[
            'numpy>=1.6',
            'scipy>=0.9',
            'pythran',
        ],

        zip_safe=False,
        ext_modules=[PythranExtension('matmul', ['matmul.py'])]
    )
```

because of the requirement to `import` `pythran` into `setup.py`, it cannot be installed using `pip` if `pythran` wasn't installed first.

However `setuptools` allows libraries to define their own keyword arguments, in this case `pythran_modules`.

```
import setuptools
from distutils.core import Extension

if __name__ == "__main__":
    setuptools.setup(
        name='matmul',
        version='0.1',
        packages=setuptools.find_packages(exclude=['tests']),

        setup_requires=[
            'pythran'
        ],
        install_requires=[
            'numpy>=1.6',
            'scipy>=0.9',
            'pythran',
        ],

        zip_safe=False,
        pythran_modules=[Extension('matmul', ['matmul.py'])],
    )
```

The keyword argument will be ignored if it is not defined and `pythran` will be installed as per `setup_requires`. After that is done, the installation continues and `matmul.py` will be compiled by pythran.

Unfortunately, for this to work we need to modify `Extension` after it has been created. Before we could use `PythranExtension` which modified its content before calling the `Extension` constructor.

Now, this behaviour is turned on its head using a `pythranize()` function, which modifies an already existing `Extension` and quickly compiles its sources before continuing.

### DIFF
--- a/doc/MANUAL.rst
+++ b/doc/MANUAL.rst
@@ -278,12 +278,12 @@ When distributing a Python application with Pythran modules, you can either:
 
 * declare the module as a regular Python module. After all, they are 100% Python compatible.
 
-* declare them as a ``PythranExtension`` and Pythran will compile them::
+* declare them as an ``Extension`` in ``pythran_modules`` and Pythran will compile them::
 
-    from distutils.core import setup
-    from pythran.dist import PythranExtension
+    from distutils.core import setup, Extension
     setup(...,
-          ext_modules=[PythranExtension("mymodule", ["mymodule.py"])])
+          setup_requires=['pythran'],
+          pythran_modules=[Extension("mymodule", ["mymodule.py"])])
 
 Advanced Usage
 --------------
--- a/pythran/dist.py
+++ b/pythran/dist.py
@@ -21,41 +21,58 @@ class PythranExtension(Extension):
     The compilation process ends up in a native Python module.
     '''
     def __init__(self, name, sources, *args, **kwargs):
-        # the goal is to rely on original Extension
-        # to do so we convert the .py to .cpp with pythran
-        # and register the .cpp in place of the .py
-        # That's stage 0, and it's enough if you get the source
-        # from github and `python setup.py install it`
-        #
-        # *But* if you want to distribute the source through
-        # `python setup.py sdist` then the .py no longer exists
-        # and only the .cpp is distributed. That's stage 1
 
-        cxx_sources = []
-        for source in sources:
-            base, _ = os.path.splitext(source)
-            output_file = base + '.cpp'  # target name
+        Extension.__init__(self, name, sources, *args, **kwargs)
+        pythranize(self)
 
-            # stage 0 when we have the .py
-            if os.path.exists(source):
-                stage = 0
-            # stage 1 otherwise. `.cpp' should already be there
-            # as generated upon stage 0
+
+def pythranize(ext):
+    # the goal is to rely on original Extension
+    # to do so we convert the .py to .cpp with pythran
+    # and register the .cpp in place of the .py
+    # That's stage 0, and it's enough if you get the source
+    # from github and `python setup.py install it`
+    #
+    # *But* if you want to distribute the source through
+    # `python setup.py sdist` then the .py no longer exists
+    # and only the .cpp is distributed. That's stage 1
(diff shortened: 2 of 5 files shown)


serge-sans-paille:
Looks great! I was not aware of this feature, you will probably need to update 
https://github.com/serge-sans-paille/pythran/blob/master/pythran/tests/test_distutils/setup.py and https://github.com/serge-sans-paille/pythran/blob/master/pythran/tests/test_distutils_packaged/setup.py to test your change (please use ``--amend`` to keep the PR clean)

serge-sans-paille:
And the Distutils Integration section in https://github.com/serge-sans-paille/pythran/blob/master/doc/MANUAL.rst too :-)

nils-werner:
I have updated the places you mentioned.

I am pretty certain this introduces problems I am not yet aware of, so a few more tests testing stage 0 and stage 1 implementations may be needed.

serge-sans-paille:
> I have updated the places you mentioned.

Great! Thanks a lot!

> I am pretty certain this introduces problems I am not yet aware of, so a few more tests testing stage 0 and stage 1 implementations may be needed.

That's a good start though. As soon as we get feedbacks from other users
having issues, we'll add more tests ;-)

serge-sans-paille:
@nils-werner looks like your solution does not work for bdist target, see

https://travis-ci.org/serge-sans-paille/pythran/jobs/136360673

serge-sans-paille:
@nils-werner I've been digging through your issue, what do you think of the following:

```python
from distutils.core import Extension
from setuptools import setup

class ProxyExtension(Extension):

    def __init__(self, *args, **kwargs):
        self.proxy = None
        self.proxy_args = args
        self.proxy_kwargs = kwargs


    def __getattr__(self, name):
        if not self.proxy:
            from pythran import PythranExtension
            self.proxy = PythranExtension(*args, **kwargs)
        return getattr(self.proxy, name)

setup(name = 'demo',
      version = '1.0',
      description = 'This is a demo package',
      setup_requires=['pythran'],
      ext_modules = [ProxyExtension('demo', sources = ['a.py'])])
```

serge-sans-paille:
Another alternative : include the following line before importing pythran:

```python
setuptools.dist.Distribution(dict(setup_requires='pythran'))
```

pbrunet:
Looking at your code serge, It looks like you make Pythran required for pythran modules while it didn't for nils extension (and I prefer this)
I didn't look at the problem, can you explain why does it fails? If I correctly remember, bdist used to import .cpp files while now, it should just do the "classic" job and keep .py file. (so it should be just the default behavior).

serge-sans-paille:
@pbrunet : in both case, you need pythran as a dep (see the ``setup_required`` field in @nils-werner solution). You need it because of the header dependencies, at least. The problem pointed by @nils-werner  (and he's 100% right) is that current solution **requires** pythran to be installed to work, but does not enforce this in the ``setup.py`` file.

I propose to solve the problem by adding a line in the documentation stating that:

```python
import setuptools
setuptools.dist.Distribution(dict(setup_requires='pythran'))
```

is requited at the beginning of the ``setup.py`` file, before ``import pythran``. @nils-werner solution looks nice, but I cannot find a way to make it work in all the configuration we test.

serge-sans-paille:
@nils-werner if you're ok with this, I'll close the PR and update the documentation and tests to enforce usage of

```python
import setuptools
setuptools.dist.Distribution(dict(setup_requires='pythran'))
```

Thanks a lot for pointing out the issue. Even if your PR does not fix it the way we expected, we now have a way to fix it, somehow!

serge-sans-paille:
See #600 for a hacky but working fix

PR #582, opened 2016-06-20
Fix config file without GMP support.

pbrunet:
* a+ replace content but smaller content will keep garbage at the end
without manual truncate


serge-sans-paille:
great fix! LGTM

PR #587, opened 2016-06-27
Fix/nicer global assignment error reporting

serge-sans-paille:


serge-sans-paille:
@pbrunet various fixes related to pythran days, travis is gr33n

pythran/pythonic/include/types/combined.hpp
```diff
@@ -214,4 +214,21 @@ struct __combined<pythonic::types::variant_functor<Types0...>,
...
+SCALAR_COMBINER(bool)
+SCALAR_COMBINER(uint8_t)
+SCALAR_COMBINER(int8_t)
+SCALAR_COMBINER(uint16_t)
+SCALAR_COMBINER(int16_t)
+SCALAR_COMBINER(uint32_t)
+SCALAR_COMBINER(int32_t)
+SCALAR_COMBINER(uint64_t)
+SCALAR_COMBINER(int64_t)
+#undef SCALAR_COMBINER
```
pbrunet: Can we remove the part with same combination on operators?

pythran/pythonic/include/types/list.hpp
```diff
@@ -329,6 +331,12 @@ namespace pythonic
   struct assignable<types::list<T>> {
     typedef types::list<typename assignable<T>::type> type;
   };
+
+  // to cope with std::vector<bool> specialization
+  template <>
+  struct returnable<types::list<bool>::reference> {
```
pbrunet: I don't get it. Why a list here?

serge-sans-paille: > > +  // to cope with std::vector<bool> specialization
> > +  template <>
> > +  struct returnable<types::list<bool>::reference> {
> 
> I don't get it. Why a list here?

because types::list<bool> uses std::vector<bool> for its storage and
std::vector<bool> is specialized in a way that does not interacts well
with pythran

pythran/pythonic/include/types/combined.hpp
serge-sans-paille: I don't think so. We need both :/

PR #593, opened 2016-07-21
Customizable blas linking

serge-sans-paille:


pythran/config.py
```diff
@@ -78,9 +78,13 @@ def parse_define(define):
         extension["include_dirs"].append(here + '/pythran/pythonic/patch')
 
     # blas dependency
-    numpy_blas = numpy_sys.get_info("blas")
-    extension['libraries'].extend(numpy_blas.get('libraries', []))
-    extension['library_dirs'].extend(numpy_blas.get('library_dirs', []))
+    user_blas = cfg.get('compiler', 'blas')
+    if user_blas:
+        extension['libraries'].extend(user_blas)
```
pbrunet: What about library_dirs?

pbrunet: Just ignore, there is a CC flags arguments for this...

pbrunet: Why using a specific blas attribute instead of ldflags?

serge-sans-paille: On Thu, Jul 21, 2016 at 11:53:45AM -0700, pbrunet wrote:
> > @@ -78,9 +78,13 @@ def parse_define(define):
> >          extension["include_dirs"].append(here + '/pythran/pythonic/patch')
> >  
> >      # blas dependency
> > -    numpy_blas = numpy_sys.get_info("blas")
> > -    extension['libraries'].extend(numpy_blas.get('libraries', []))
> > -    extension['library_dirs'].extend(numpy_blas.get('library_dirs', []))
> > +    user_blas = cfg.get('compiler', 'blas')
> > +    if user_blas:
> > +        extension['libraries'].extend(user_blas)
> 
> Why using a specific blas attribute instead of ldflags?

if numpy decides that blas comes from cblas, there's no way to say it :
no use openblas instead. that's what this patch is for. playing with
ldflags cannot change this behavior, unfortunately

PR #595, opened 2016-07-21
Corrects explanatory prose about tuple unpacking

jfinkels:


doc/TUTORIAL.rst
```diff
@@ -101,7 +101,7 @@ transformation::
       a = __tuple0[0]
       b = __tuple0[1]
 
-Note that Pythran wraps the sequence of assignment into a dummy if condition.
+Pythran transforms the tuple unpacking into an intermediate tuple assignment.
 This ensures that a single instruction from the input code maps to a single
```
serge-sans-paille: Maybe just transform ``This ensures`` here into ``The if statement ensures``

pbrunet: It looks like no more If are generated

jfinkels: @serge-sans-paille I don't see an if statement in either the original code or the transformed code.

serge-sans-paille: On Thu, Jul 21, 2016 at 12:36:58PM -0700, jfinkels wrote:
> > @@ -101,7 +101,7 @@ transformation::
> >        a = __tuple0[0]
> >        b = __tuple0[1]
> >  
> > -Note that Pythran wraps the sequence of assignment into a dummy if condition.
> > +Pythran transforms the tuple unpacking into an intermediate tuple assignment.
> >  This ensures that a single instruction from the input code maps to a single
> 
> @serge-sans-paille I don't see an if statement in either the original code or the transformed code.

Correct!
@pbrunet: the whole story is that the if got removed in favor of a
temporary assignment which was still correct with respect to openmp,
then the temporaty got removed, and that's incorrect. i'll accept
@jfinkels PR and open  #596 to keep track of this.

PR #597, opened 2016-07-25
Up-to-date tuple unpacking doc w/ OpenMP

serge-sans-paille:


serge-sans-paille:
@pbrunet ready4review

PR #598, opened 2016-08-01
Uses object instead of None as node in CFG.

jfinkels:
The *[Creating a graph][1]* section of the NetworkX documentation states, "Note: Python’s None object should not be used as a node as it determines whether optional function arguments have been assigned in many functions." For example, if I want to get the dictionary of node degrees, normally I would do `dict(G.degree())` but that causes an error due to the `None` node. This commit changes the sink node to be a special sentinel object, `NIL`.

[1]: http://networkx.readthedocs.io/en/latest/tutorial/tutorial.html#creating-a-graph

### DIFF
--- a/pythran/analyses/cfg.py
+++ b/pythran/analyses/cfg.py
@@ -14,6 +14,13 @@ class CFG(FunctionAnalysis):
     * the OUT nodes, to be linked with the IN nodes of the successor
     * the RAISE nodes, nodes that stop the control flow (exception/break/...)
     """
+
+    #: The sink node in the control flow graph.
+    #:
+    #: The predecessors of this node are those AST nodes that terminate
+    #: control flow without a return statement.
+    NIL = object()
+
     def __init__(self):
         self.result = nx.DiGraph()
         super(CFG, self).__init__()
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
Great! Thanks!
You should update transformations/normalize_return.py:33 to reflect your change (that's the only transformation that uses this analyse)

serge-sans-paille:
And btw @jfinkels how / what for are you using pythran?

jfinkels:
I see, I've updated `normalize_return.py`. I'm experimenting with control flow analysis of Python programs.

pythran/analyses/cfg.py
```diff
@@ -6,6 +6,10 @@
 import networkx as nx
 
 
+#: A singleton object to use as the sink node in a graph.
+NIL = object()
+
+
 class CFG(FunctionAnalysis):
     """
```
serge-sans-paille: That's picky, but I think you should make ``NIL`` a (static) member of ``CFG``, that way you would not need to import it globally at the same level as other analysis (using ``CFG.NIL`` looks better)

jfinkels: Okay, I'll make that change.

jfinkels:
This has been updated with the `NIL` node a class attribute of the `CFG` class.

PR #601, opened 2016-08-21
Ufunc

pbrunet:


pythran/pythonic/core.hpp
```diff
@@ -1,6 +1,11 @@
 #ifndef PYTHONIC_CORE_HPP
 #define PYTHONIC_CORE_HPP
 
+#define QUOTEME(M) #M
```
serge-sans-paille: there's already a similar macro in sources/pythran/pythran/pythonic/include/types/numpy_nary_expr.hpp

```
#define STR_(a) #a
```

pythran/pythonic/include/numpy/mod.hpp
```diff
@@ -15,8 +15,7 @@ namespace pythonic
      * operator% which is correctly overloaded
      */
 
-    using operator_::mod;
-    DECLARE_FUNCTOR(pythonic::numpy, mod);
+    USING_FUNCTOR(mod, operator_::functor::mod);
```
serge-sans-paille: nice! brest!

pythran/pythonic/numpy/mod.hpp
```diff
@@ -2,24 +2,6 @@
...
-
-    /* this is still a numpy_expr, because operator::mod_ forwards to
-     * operator% which is correctly overloaded
-     */
-
-    using operator_::mod;
-    DEFINE_FUNCTOR(pythonic::numpy, mod);
-  }
-}
-
```
serge-sans-paille: plouzané!

pythran/tests/test_numpy_ufunc_binary.py
```diff
@@ -3,39 +3,32 @@
...
-    "logical_and", "logical_or", "logical_xor",
-    'maximum', 'minimum', 'mod', 'multiply',
-    'not_equal',  # 'nextafter', TODO : Why?
-    'power',
-    'remainder', 'right_shift',
-    'subtract',
-    'true_divide',
-)
+#automatic generation of basic test cases for ufunc
+binary_ufunc = (k for k, v in MODULES["numpy"].items() if isinstance(v, UFunc))
```
serge-sans-paille: smart!

pythran/pythonic/numpy/absolute/accumulate.hpp
```diff
@@ -0,0 +1,7 @@
+#ifndef PYTHONIC_NUMPY_ABSOLUTE_ACCUMULATE_HPP
```
serge-sans-paille: how does it work if `abs` is used instead of `absolute`?

pythran/pythonic/include/numpy/partial_sum.hpp
```diff
@@ -9,39 +9,44 @@ namespace pythonic
...
+      template <class E, class F>
       A operator()(E e, F &o, A acc);
     };
 
-    template <class Op, class E,
-              class dtype = types::dtype_t<typename E::dtype>>
+    template <class Op, class E>
+    using result_dtype = types::dtype_t<decltype(std::declval<Op>()(
+        std::declval<typename std::remove_reference<E>::type::dtype>(),
+        std::declval<typename std::remove_reference<E>::type::dtype>()))>;
```
serge-sans-paille: That's much more generic, the assumption that this is equal to `E::dtype` does not always hold?

serge-sans-paille: you could have used a `class... E` for more genericity and laess typing :-)

pythran/pythonic/include/numpy/ufunc_accumulate.hpp
```diff
@@ -0,0 +1,26 @@
+#ifndef UFUNC_NAME
+#error missing UFUNC_NAME
+#endif
+
+// clang-format off
+#include INCLUDE_FILE(pythonic/include/numpy,UFUNC_NAME)
+// clang-format on
```
serge-sans-paille: out of curiosity: what was clang doing otherwise?

pythran/pythonic/numpy/partial_sum.hpp
```diff
@@ -13,24 +12,53 @@ namespace pythonic
...
-    template <class Op>
-    template <class E, class F, class A>
-    A _partial_sum<Op, 1>::operator()(E e, F &o, A acc)
+    template <class Op, class A>
+    template <class E, class F>
+    A _partial_sum<Op, 1, A>::operator()(E e, F &o, A acc)
     {
       for (auto const &value : e) {
-        Op{}(acc, value);
+        acc = Op{}(acc, (A)value);
```
serge-sans-paille: This line look suspicious: why do you need the cast? It seem to prevent complex situation when accumulating over a numpy_expression.

pythran/pythonic/numpy/partial_sum.hpp
```diff
@@ -13,24 +12,53 @@ namespace pythonic
...
+    {
+      auto it_begin = e.begin();
+      A acc = *it_begin;
+      *o = acc;
+      ++o;
+      ++it_begin;
+      for (; it_begin < e.end(); ++it_begin) {
+        acc = Op{}(acc, (A)*it_begin);
+        *o = acc;
+        ++o;
```
serge-sans-paille: I'd rather put the `++o` inside the loop header `++it_begin, ++o` make things look more explicit to me.

pythran/pythonic/numpy/partial_sum.hpp
```diff
@@ -40,9 +68,8 @@ namespace pythonic
       const long count = expr.flat_size();
       types::ndarray<typename dtype::type, 1> the_partial_sum{
           types::make_tuple(count), __builtin__::None};
-      auto const p = utils::neutral<Op, typename E::dtype>::value;
-      auto iter = the_partial_sum.begin();
-      _partial_sum<Op, E::value>{}(expr, iter, p);
```
serge-sans-paille: if you no longer use neutral, then remove its definition (it was the only place it was used?)

pythran/pythonic/include/types/ndarray.hpp
```diff
@@ -491,10 +492,9 @@ namespace pythonic
       template <class T>
       struct dtype_helper;
 
-      // FIXME: use another type?
       template <>
       struct dtype_helper<bool> {
-        using type = pythonic::numpy::functor::uint8;
+        using type = pythonic::numpy::functor::bool_;
```
serge-sans-paille: I think you'll have issue with std::vector specialization...

pythran/pythonic/include/numpy/partial_sum.hpp
pbrunet: First I though it was enough but I had some bugs. You can look at equal. Inputs may be two int arrays but the result is a bool array.

pythran/pythonic/include/numpy/ufunc_accumulate.hpp
pbrunet: It doesn't detect we are performing an include and add spaces around "\" which give a wrong path

pythran/pythonic/numpy/partial_sum.hpp
pbrunet: Still an issue with boolean operations. With "equal" again, ccc may be [True, True] and value [1, 2], then equal(acc, value) return [True, True] while it will return [False, False] withotut the cast. I first though it was numpy.equal which behave this way but indead it is only for accumulation in numpy.

pythran/pythonic/numpy/partial_sum.hpp
pbrunet: It is still present for reduct but it will be my next step as reduction is also a ufunc method.

pythran/pythonic/include/types/ndarray.hpp
pbrunet: I didn't have issue with bool result on accumulate. :-s

pythran/pythonic/numpy/absolute/accumulate.hpp
pbrunet: Hum, I missed it, I should add it

pythran/pythonic/numpy/partial_sum.hpp
serge-sans-paille: On Mon, Aug 22, 2016 at 01:23:54PM -0700, pbrunet wrote:
> >      {
> >        for (auto const &value : e) {
> > -        Op{}(acc, value);
> > +        acc = Op{}(acc, (A)value);
> 
> Still an issue with boolean operations. With "equal" again, ccc may be [True, True] and value [1, 2], then equal(acc, value) return [True, True] while it will return [False, False] withotut the cast. I first though it was numpy.equal which behave this way but indead it is only for accumulation in numpy.

What's the signature of Op{} in that case ? If it's ool (bool, bool)
then there's no issue, right ?


Looks like a bad signature of Op...

pbrunet: In this case, Op{} signature is : (A a, B b) -> decltype(a == b).
So with A = bool and B = int, it will return false. With the cast, A = bool and B = bool and it return True

PR #603, opened 2016-08-26
Use new gast version for ExceptHandler.

pbrunet:
* name attribut is now an identifier for Python2


pbrunet:
gast is updated with Python2 attribute

PR #605, opened 2016-09-21
Fix subtle performance issue with vectorisation & broadcasting

serge-sans-paille:
The issue is that to achieve broadcasting, we need a dynamic test in the
inner expression template loop. No way around I could find.

So let's add a preliminary check on the ET, and if the ET appear to
*not* need any broadcasting, then another vectorization loop, condition
free, is selected at runtime (which means both version are actually
compiled).

### DIFF
--- a/pythran/pythonic/types/numpy_expr.hpp
+++ b/pythran/pythonic/types/numpy_expr.hpp
@@ -74,6 +74,20 @@ namespace pythonic
 
     template <class Op, class... Args>
     template <int... I>
+    bool numpy_expr<Op, Args...>::_no_broadcast(utils::seq<I...>) const
+    {
+      std::initializer_list<long> nb = {std::get<I>(args).shape()[0]...};
+      return std::all_of(nb.begin(), nb.end(),
+                         [this](long n) { return n == size() || n == 0; });
+    }
+    template <class Op, class... Args>
+    bool numpy_expr<Op, Args...>::no_broadcast() const
+    {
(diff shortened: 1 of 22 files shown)


serge-sans-paille:
@pbrunet: ready 4 review!

pythran/pythonic/types/numpy_expr.hpp
```diff
@@ -74,6 +74,21 @@ namespace pythonic
 
     template <class Op, class... Args>
     template <int... I>
+    bool numpy_expr<Op, Args...>::_no_broadcast(utils::seq<I...>) const
+    {
+      std::initializer_list<bool> nb = {
+          ((size() == std::get<I>(args).shape()[0]) ||
+           (0 == std::get<I>(args).shape()[0]))...};
+      return std::all_of(nb.begin(), nb.end(), [](bool b) { return b; });
```
pbrunet: It is details but if you perform your comparisons in the boolean test of all_of instead of doing this in the std::initialiser_list, you "might" compute less comparisons.

pbrunet:
I have to admit I don't really see what happened (neither what happen) as it imply a lot of modification in multiple places but it LGTM. If you want to care about my comment, go one, otherwise, we can merge it.

serge-sans-paille:
> It is details but if you perform your comparisons in the boolean test of all_of instead of doing this in the std::initialiser_list, you "might" compute less comparisons.

 Yes! Will do!

PR #607, opened 2016-09-30
Fix various leaks due to poor API design

serge-sans-paille:


serge-sans-paille:
@pbrunet : ready4review

serge-sans-paille:
On Tue, Oct 04, 2016 at 12:56:40PM -0700, pbrunet wrote:
> LGTM, Is there any way to test it? (I already saw some code with valgrind header but I don't know if it can be apply for pythran as 1) Valgrind on Python show a lots of errors 2) It certainly doesn't fit in the "kind of test" we run.

The current test suite asserts we are not freeing too much (that would
segfault).

I ran some tests with

    valgrind --log-file=valgrind.log --error-limit=no --leak-check=full --show-leak-kinds=all py.test pythran/tests/test_cases.py

and checked the reference to pythonic within :-/

pbrunet:
Ok, it looks hard to automated this.

PR #608, opened 2016-09-30
Fix python3

pbrunet:


.travis.yml
```diff
@@ -29,22 +29,22 @@ env:
     # Test with both clang and gcc
     # Only gcc can test openMP, and travis does not support gcc when gcc uses too much memory
     # So some tests have been commented out, as it is useless to compile them with clang
-    - PYTHON_VERSION=2 PEP8=false TESTCASE=pythran/tests/test_advanced.py CC=clang-3.5 CXX=clang++-3.5 CXXFLAGS="-g"
```
serge-sans-paille: I though that we had to force the python version, but your usage seems to lead toward : if it's both python2 and Python3 compatible, then no PYTHON_VERSION=XX ?

serge-sans-paille:
Travis spot failures in test_advanced.py

pythran/optimizations/iter_transformation.py
```diff
@@ -59,15 +59,21 @@ def find_matching_builtin(self, node):
...
-        """ Replace function call by its correct iterator if it is possible."""
+        """Replace function call by its correct iterator if it is possible."""
         if node in self.potential_iterator:
             match_keyword = self.find_matching_builtin(node)
+            # Special handling for map which can't be turn to imap with None as
+            # a parameter as map(None, [1, 2]) == [1, 2] while
+            # list(imap(None, [1, 2])) == [(1,), (2,)]
+            if match_keyword == "map":
+                if MODULES["__builtin__"]["None"] in self.aliases[node.args[0]]:
+                    return self.generic_visit(node)
```
serge-sans-paille: I think it's safer to check that aliases are functiondef. One could alias to a parameter that could be none, and in that case better be conservative.

pythran/pythonic/__builtin__/enumerate.hpp
```diff
@@ -55,23 +55,33 @@ namespace pythonic
         return *this;
       }
 
+      // Comparison operators can't use value as end() doesn't have a valid value content
+      // du to the lake of size information for generator
```
serge-sans-paille: du*e*

serge-sans-paille: *lack*

pythran/transformations/remove_comprehension.py
```diff
@@ -147,11 +147,10 @@ def visit_GeneratorExp(self, node):
                       ast.Expr(ast.Yield(node.elt))
                       )
 
-        sargs = sorted(ast.Name(arg, ast.Param(), None) for arg in args)
```
serge-sans-paille: I wonder why we were sorting... looks better like this!

pbrunet:
FInally, I am green !! (there are still some issues but let's go for this version, others will be fix later as they are already present).

About the PYTHON_VERSION, it also works this way. (It doesn't really like this but it works. At least, this variable should be remove as soon as we care about others python3 errors)

pythran/analyses/constant_expressions.py
```diff
@@ -25,32 +25,32 @@ def add(self, node):
         return True
 
     def visit_BoolOp(self, node):
-        return all(map(self.visit, node.values)) and self.add(node)
+        return all([self.visit(x) for x in node.values]) and self.add(node)
```
serge-sans-paille: you could even remove the hooks and use a generator expression, right?

pythran/analyses/parallel_maps.py
```diff
@@ -18,8 +18,8 @@ def __init__(self):
     def visit_Call(self, node):
         if all(alias == MODULES['__builtin__']['map']
                for alias in self.aliases[node.func]):
-            if all(self.pure_expressions.__contains__(f)
-                    for f in self.aliases[node.args[0]]):
+            if all(f in self.pure_expressions
+                   for f in self.aliases[node.args[0]]):
                 self.result.add(node)
```
serge-sans-paille: wow, did I really wrote this o_O

pythran/analyses/use_def_chain.py
```diff
@@ -229,14 +234,16 @@ def visit_For(self, node):
         self.in_loop = True
         old_node = {i: set(j) for i, j in self.current_node.items()}
         self.visit(node.target)
-        list(map(self.visit, node.body))
+        for n in node.body:
```
serge-sans-paille: +1, that's more pythonic this way, after all. And more portable across 2/3

pythran/backend.py
```diff
@@ -1075,7 +1075,7 @@ def visit_Raise(self, node):
 
     def visit_Assert(self, node):
         params = [self.visit(node.test), node.msg and self.visit(node.msg)]
-        sparams = ", ".join(map(strip_exp, [_f for _f in params if _f]))
+        sparams = ", ".join((strip_exp(_f) for _f in params if _f))
```
serge-sans-paille: are the parenthesis useful here?

pythran/config.py
```diff
@@ -91,13 +91,15 @@ def parse_define(define):
...
 def compiler():
+    """Get compiler to use for C++ to binary process."""
     return os.environ.get('CXX', 'c++')
 
 
 def have_gmp_support(**extra):
-    """ Check if the USE_GMP macro is defined. """
-    return any("USE_GMP" == name
-               for name, _ in make_extension(**extra)["define_macros"])
+    """Check if the USE_GMP macro is defined."""
```
serge-sans-paille: why? No OpenMP support for python3 ?

pythran/intrinsic.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, **kwargs):
             [to_ast(d) for d in kwargs.get('defaults', [])])
         self.return_range = kwargs.get("return_range",
                                        lambda call: UNKNOWN_RANGE)
-        self.return_range_content = kwargs.get("return_ange_content",
+        self.return_range_content = kwargs.get("return_range_content",
```
serge-sans-paille: \o/

pythran/optimizations/iter_transformation.py
```diff
@@ -59,15 +59,22 @@ def find_matching_builtin(self, node):
...
     def visit_Call(self, node):
-        """ Replace function call by its correct iterator if it is possible."""
+        """Replace function call by its correct iterator if it is possible."""
         if node in self.potential_iterator:
             match_keyword = self.find_matching_builtin(node)
+            # Special handling for map which can't be turn to imap with None as
+            # a parameter as map(None, [1, 2]) == [1, 2] while
+            # list(imap(None, [1, 2])) == [(1,), (2,)]
+            if (match_keyword == "map" and
+                    MODULES["__builtin__"]["None"] in
```
serge-sans-paille: correct, thanks for spotting this.

pythran/pythonic/include/numpy/complex.hpp
```diff
@@ -3,16 +3,26 @@
...
 namespace pythonic
 {
 
   namespace numpy
   {
+    namespace details
+    {
 
-    std::complex<double> complex();
+      std::complex<double> complex();
```
serge-sans-paille: is the first declaration still useful that way?

serge-sans-paille:
@pbrunet once the genexpr vs listcomp issue is clarified, and some squashing happens, I'm ok with this impressive PR!

pythran/analyses/constant_expressions.py
pbrunet: I can as all() stop evaluating generator as soon as it evaluate a value to false and I want to evaluate everything.

pythran/analyses/pure_expressions.py
```diff
@@ -27,14 +27,14 @@ def visit_FunctionDef(self, node):
...
-        is_pure = all(map(self.visit, ast.iter_child_nodes(node)))
+        is_pure = all([self.visit(x) for x in ast.iter_child_nodes(node)])
         if is_pure:
             self.result.add(node)
         return is_pure
 
     def visit_Call(self, node):
         # check if all arguments are Pures
-        is_pure = all(self.visit(arg) for arg in node.args)
+        is_pure = all([self.visit(arg) for arg in node.args])
```
pbrunet: For a case like:
```python
foo(a, 2 + 4)
```
a is not pure but I want to keep going and looks other argument. Otherwise, `2 + 4` will not be marked pure and constant folding will not happen.

pythran/config.py
pbrunet: Not OPENMP but GMP ;-)

serge-sans-paille:
Thanks for the clarifications on the lazyness of all/any! Maybe add them in the form of comments in the code so that the question does not arise later, do some squashing and I'll merge :-)

serge-sans-paille:
Impressive job @pbrunet ;-)

PR #612, opened 2016-10-16
Fix bug when copying numpy expressions

aguinet:
The move operator wasn't declared, so the case where we had:

numpy_gexpr e;
e = ...;

didn't work because the buffer member variable was set to an
uninitialized value.

Morever, add asserts in copy functions to ease futur debuggings!

### DIFF
--- a/pythran/pythonic/types/numpy_iexpr.hpp
+++ b/pythran/pythonic/types/numpy_iexpr.hpp
@@ -15,7 +15,8 @@ namespace pythonic
   {
 
     template <class Arg>
-    numpy_iexpr<Arg>::numpy_iexpr()
+    numpy_iexpr<Arg>::numpy_iexpr():
+      buffer(nullptr)
     {
     }
 
@@ -54,6 +55,7 @@ namespace pythonic
     template <class E>
     numpy_iexpr<Arg> &numpy_iexpr<Arg>::operator=(E const &expr)
     {
+      assert(buffer);
       return utils::broadcast_copy<numpy_iexpr &, E, value,
                                    value - utils::dim_of<E>::value,
                                    false /*NIY*/>(*this, expr);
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
please rebase on master, we just caught up with numpy versipon update

pythran/pythonic/include/types/numpy_expr.hpp
```diff
@@ -428,6 +428,9 @@ namespace pythonic
 
       numpy_expr(Args const &... args);
 
+      numpy_expr& operator=(numpy_expr const&) = default;
+      numpy_expr& operator=(numpy_expr&&) = default;
```
serge-sans-paille: I'm dubious about these ones. numpy_expr sometimes have a reference as member variable, which make the move operator a non sense.
The situation you're meeting, with a forward declaration of the type, is rather tricky. the only pythranic way to do this would be to force the type to be assignable, so that you get an ndarray to hold the result of the expression.

pythran/pythonic/types/numpy_iexpr.hpp
```diff
@@ -15,7 +15,8 @@ namespace pythonic
   {
 
     template <class Arg>
-    numpy_iexpr<Arg>::numpy_iexpr()
+    numpy_iexpr<Arg>::numpy_iexpr():
+      buffer(nullptr)
```
serge-sans-paille: yep, that looks much more robust :-)

pythran/pythonic/include/types/numpy_expr.hpp
aguinet: I just need the move operator. I changed them for an optional type so I think we can remove them.

serge-sans-paille:
You mean = delete?

aguinet:
> You mean = delete?
I didn't understand, delete what :) ?

serge-sans-paille:
Forget about it :-)

There seems to be a segfault in 

https://travis-ci.org/serge-sans-paille/pythran/jobs/169718629

can you reproduce it?

aguinet:
I can reproduce ! Currently investigating... !

aguinet:
It seemed like the declaration of the move version of operator= caused an infinite recursion. I didn't investigate more than this, since we don't need them anymore in the cython pythran backend!

serge-sans-paille:
@aguinet this one needs a rebase too, due to the numpy update.

aguinet:
Rebase done! Thx serge :)

serge-sans-paille:
you just need a correct call to clang-3.6-format!

serge-sans-paille:
@aguinet up!

aguinet:
Fixed and rebased on master!

serge-sans-paille:
obsolteted by #632

PR #613, opened 2016-10-16
Add to_python_from_expr API

aguinet:
This force a returnable type if a numpy expression is passed!

### DIFF
--- a/pythran/pythonic/python/core.hpp
+++ b/pythran/pythonic/python/core.hpp
@@ -25,6 +25,15 @@ auto to_python(T &&value) -> decltype(pythonic::to_python<
       typename std::remove_cv<typename std::remove_reference<T>::type>::type>::
       convert(std::forward<T>(value));
 }
+
+template <class T>
+auto to_python_from_expr(T &&value) -> decltype(to_python(
+      typename pythonic::returnable<typename std::remove_cv<typename std::remove_reference<T>::type>::type>::type{std::forward<T>(value)}))
+{
+  using returnable_type = typename pythonic::returnable<typename std::remove_cv<typename std::remove_reference<T>::type>::type>::type;
+  return to_python(returnable_type{std::forward<T>(value)});
+}
+
 template <class T>
 T from_python(PyObject *obj)
 {


pythran/pythonic/python/core.hpp
```diff
@@ -25,6 +25,15 @@ auto to_python(T &&value) -> decltype(pythonic::to_python<
...
 }
+
+template <class T>
+auto to_python_from_expr(T &&value) -> decltype(to_python(
+      typename pythonic::returnable<typename std::remove_cv<typename std::remove_reference<T>::type>::type>::type{std::forward<T>(value)}))
+{
+  using returnable_type = typename pythonic::returnable<typename std::remove_cv<typename std::remove_reference<T>::type>::type>::type;
+  return to_python(returnable_type{std::forward<T>(value)});
+}
+
```
pbrunet: expr is really specific to ndarray. I guess it should be with `to_python<types::numpy_iexpr<Arg>>` in ndarray.hpp or do you have another usage?

aguinet: I actually use it w/ Cython to convert any kind of expression with the same API (even if it's a long for instance). Maybe I should change the name :)

serge-sans-paille:
please rebase on master, we just caught up with numpy versipon update

pythran/pythonic/python/core.hpp
aguinet: It's a bit hard because I let the C++ compiler defines the types and don't process this information at the cython level (at least for now). I can inject this function inside the generated cython code if necessary.

serge-sans-paille:
> I can inject this function inside the generated cython code if necessary.

Yes, I think that's a proper solution.

serge-sans-paille:
@aguinet feel free to close this one if it's no longer needed

aguinet:
@serge-sans-paille for now I've been included this code straight into the cython generated code. I still think it would be nice to have a "universal" to_python API, but I think I'll do that once I've finished the other things ;)

aguinet:
I'm closing this one for now!

PR #614, opened 2016-10-16
Do not disable deprecated Numpy API when compiling from Cython

aguinet:
Cython still uses deprecated Numpy APIs!

### DIFF
--- a/pythran/pythonic/types/ndarray.hpp
+++ b/pythran/pythonic/types/ndarray.hpp
@@ -56,11 +56,6 @@
 #include <initializer_list>
 #include <numeric>
 
-#ifdef ENABLE_PYTHON_MODULE
-#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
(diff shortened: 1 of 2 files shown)


pythran/pythonic/types/ndarray.hpp
```diff
@@ -57,7 +57,11 @@
 #include <numeric>
 
 #ifdef ENABLE_PYTHON_MODULE
+// Cython still uses the deprecated API, so we can't set this macro in this
+// case!
+#ifndef CYTHON_ABI
 #define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
+#endif
 #include "numpy/arrayobject.h"
 #endif
```
serge-sans-paille: It's a good opportunity to remove this duplicate include :-)

aguinet: Yes I thought about that! Will update the PR :)

aguinet:
Just pushed --force without the duplicate header include

PR #615, opened 2016-10-18
Fill as much argument as possible from introspection for named argume…

pbrunet:
…nts.


serge-sans-paille:
Thxs a lot @pbrunet !

PR #617, opened 2016-10-22
Mission Numpy type

aguinet:


pythran/conversion.py
```diff
@@ -82,7 +82,7 @@ def to_ast(value):
                   numpy.float_, numpy.uint8, numpy.uint16, numpy.uint32,
                   numpy.uint64, numpy.int8, numpy.int16, numpy.int32,
                   numpy.int64, numpy.intp, numpy.intc, numpy.int_,
-                  numpy.bool_)
+                  numpy.bool_, numpy._globals._NoValue)
```
serge-sans-paille: What's that ? Please add a test that illustrates this type :-)

aguinet:
My pythran installation wouldn't even run without this type added :)

aguinet:
Well, now it works... I don't really know why. I think I can close this one..

PR #622, opened 2016-11-29
Feature/type checking

serge-sans-paille:
WIP


serge-sans-paille:
@pbrunet all tests except pep8 pass. Python2 *and* Python3! I'll add some type checking test later on, but if you can have a look at current state.

Good luck btw :-)

PR #623, opened 2016-12-16
Release the GIL when we don't need it anymore

serge-sans-paille:


serge-sans-paille:
@pbrunet Travis is almost finished, this one should be ok.

serge-sans-paille:
travis is ok :-)

pbrunet:
Looks nice. I just wonder if we should run test twices: once with one threads, once with 2 or 4 and check we are faster?

serge-sans-paille:
On Tue, Dec 20, 2016 at 12:15:42AM -0800, pbrunet wrote:
> Looks nice. I just wonder if we should run test twices: once with one threads, once with 2 or 4 and check we are faster?

We could do this with a call to sleep because it does no consume CPU. We
can't do this otherwise because we have no control over the number of
available threads. I'll add this test case.

serge-sans-paille:
@pbrunet turns out setting up the test will be quite complicated. I did it manually and it's ok, don't wanant to bother more, you can merge if that's ok with you too.

PR #628, opened 2017-01-17
Fix ndarray contiguous type checking

aguinet:
For instance, nd.array([[1]]) has C_CONTIGUOUS *and* F_CONTIGUOUS at
True. This is actually right because the memory representation is the
same with both conventions.

### DIFF
--- a/pythran/tests/cases/diffusion_pure_python.py
+++ b/pythran/tests/cases/diffusion_pure_python.py
@@ -1,7 +1,7 @@
 # Reference: http://continuum.io/blog/the-python-and-the-complied-python
 #pythran export diffusePurePython(float [][], float [][], int)
-#runas import numpy as np;lx,ly=(2**7,2**7);u=np.zeros([lx,ly],dtype=np.double);u[lx/2,ly/2]=1000.0;tempU=np.zeros([lx,ly],dtype=np.double);diffusePurePython(u,tempU,500)
-#bench import numpy as np;lx,ly=(2**6,2**6);u=np.zeros([lx,ly],dtype=np.double);u[lx/2,ly/2]=1000.0;tempU=np.zeros([lx,ly],dtype=np.double);diffusePurePython(u,tempU,55)
+#runas import numpy as np;lx,ly=(2**7,2**7);u=np.zeros([lx,ly],dtype=np.double);u[int(lx/2),int(ly/2)]=1000.0;tempU=np.zeros([lx,ly],dtype=np.double);diffusePurePython(u,tempU,500)
+#bench import numpy as np;lx,ly=(2**6,2**6);u=np.zeros([lx,ly],dtype=np.double);u[int(lx/2),int(ly/2)]=1000.0;tempU=np.zeros([lx,ly],dtype=np.double);diffusePurePython(u,tempU,55)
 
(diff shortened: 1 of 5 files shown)


serge-sans-paille:
@aguinet everthing is ok according to travis, except PEP8 and clang-format checks, thanks a lot!

pythran/tests/cases/diffusion_pure_python.py
```diff
@@ -1,7 +1,7 @@
 # Reference: http://continuum.io/blog/the-python-and-the-complied-python
 #pythran export diffusePurePython(float [][], float [][], int)
-#runas import numpy as np;lx,ly=(2**7,2**7);u=np.zeros([lx,ly],dtype=np.double);u[lx/2,ly/2]=1000.0;tempU=np.zeros([lx,ly],dtype=np.double);diffusePurePython(u,tempU,500)
-#bench import numpy as np;lx,ly=(2**6,2**6);u=np.zeros([lx,ly],dtype=np.double);u[lx/2,ly/2]=1000.0;tempU=np.zeros([lx,ly],dtype=np.double);diffusePurePython(u,tempU,55)
+#runas import numpy as np;lx,ly=(2**7,2**7);u=np.zeros([lx,ly],dtype=np.double);u[int(lx/2),int(ly/2)]=1000.0;tempU=np.zeros([lx,ly],dtype=np.double);diffusePurePython(u,tempU,500)
```
serge-sans-paille: could be lx//2

pythran/tests/cases/diffusion_pure_python.py
```diff
@@ -1,7 +1,7 @@
...
+#bench import numpy as np;lx,ly=(2**6,2**6);u=np.zeros([lx,ly],dtype=np.double);u[int(lx/2),int(ly/2)]=1000.0;tempU=np.zeros([lx,ly],dtype=np.double);diffusePurePython(u,tempU,55)
```
serge-sans-paille: same here

pythran/tests/cases/grayscott.py
```diff
@@ -10,8 +10,9 @@ def GrayScott(counts, Du, Dv, F, k):
 
     r = 20
     u[:] = 1.0
-    U[n/2-r:n/2+r,n/2-r:n/2+r] = 0.50
-    V[n/2-r:n/2+r,n/2-r:n/2+r] = 0.25
+    nd2 = int(n/2)
```
serge-sans-paille: and there :-)

setup.cfg
```diff
@@ -5,4 +5,4 @@ test=pytest
 pep8ignore = 
     pythran/tests/*.py ALL
     papers/* ALL
-addopts=--pep8
+addopts=
```
serge-sans-paille: ??

aguinet: mouarf that sould have never landed here... (my py.test installation is not happy with this option!)

pythran/tests/cases/diffusion_pure_python.py
aguinet: I thought this wouldn't work under Python2, but it actually do work!

serge-sans-paille:
Moved to #631

PR #630, opened 2017-01-22
Fix bitwise operator output type for uint8

serge-sans-paille:
Inferred type was incorrect, which resulted in extra conversions.


serge-sans-paille:
@pbrunet this commit is likely to be ok for travis, if you have any comments feel free!

PR #634, opened 2017-02-13
Add support of numpy.fill_diagonal

serge-sans-paille:


serge-sans-paille:
@pbrunet ready 4 review

PR #647, opened 2017-03-16
Pythran file

diorcety:


pythran/spec.py
```diff
@@ -254,10 +254,13 @@ def __call__(self, path_or_text):
...
-        pythran_data = (re.sub(r'#\s*pythran', '\_o< pythran >o_/', raw)
-                        .replace('#', '')
-                        .replace('\_o< pythran >o_/', '#pythran'))
+        if python_file:
+            raw = "\n".join(SpecParser.FILTER.findall(data))
+            pythran_data = (re.sub(r'#\s*pythran', '\_o< pythran >o_/', raw)
+                            .replace('#', '')
+                            .replace('\_o< pythran >o_/', '#pythran'))
+        else:
+            pythran_data = "\n".join(['#pythran ' + str(a) for a in data.split("\n")])
```
serge-sans-paille: you're not filtering out empty lines, so
```
foo(int)

foo(str)
```

will cause an error

pythran/spec.py
```diff
@@ -316,5 +319,5 @@ def specs_to_docstrings(specs, docstrings):
         )
 
 
-def spec_parser(path):
-    return SpecParser()(path)
+def spec_parser(path, python_file=True):
```
serge-sans-paille: I would feel more comfortable with a subclass of SpecParser that just overrides the right method, putting shared code where it's meant to be, instead of this ugly boolean

pythran/toolchain.py
```diff
@@ -311,7 +311,13 @@ def compile_pythrancode(module_name, pythrancode, specs=None,
...
     from pythran.spec import spec_parser
     if specs is None:
-        specs = spec_parser(pythrancode)
+        specs = spec_parser(path_or_text)
+
+    if os.path.isfile(path_or_text):
+        with open(path_or_text) as fd:
+            pythrancode = fd.read()
+    else:
+        pythrancode = path_or_text
```
serge-sans-paille: if you do it here, maybe you can remove the file_or_path  from the spec.py file

pythran/toolchain.py
```diff
@@ -343,18 +349,24 @@ def compile_pythranfile(file_path, output_file=None, module_name=None,
...
 
     # Add compiled module path to search for imported modules
     sys.path.append(os.path.dirname(file_path))
 
-    output_file = compile_pythrancode(module_name, open(file_path).read(),
+    specs = None
+    spec_file = os.path.join(dirname, module_name + ".pythran")
+    if os.path.isfile(spec_file):
+        from pythran.spec import spec_parser
+        specs = spec_parser(spec_file, False)
```
serge-sans-paille: here, instead of the bool, you would use an `aux_spec_parser` that crates the deriving object etc

pythran/toolchain.py
```diff
@@ -343,18 +349,24 @@ def compile_pythranfile(file_path, output_file=None, module_name=None,
...
+
+    output_file = compile_pythrancode(module_name, file_path, specs,
                                       output_file=output_file,
                                       cpponly=cpponly,
                                       **kwargs)
```
serge-sans-paille: I don't see a single test case :-)

pythran/toolchain.py
diorcety: The point is to have the filepath inside spec.py in order to have the file path in the error message

diorcety: BTW, shouldn't we replace non #pythran line by empty line in order to avoid wrong line information?

pythran/spec.py
```diff
@@ -254,18 +247,46 @@ def __call__(self, path_or_text):
...
+            pythran_data += '\n' * data[i:begin].count(b'\n') + data[begin:end]
+            i = end
+        pythran_data += '\n' * data[i:len(data)].count(b'\n')
+
+        # Remove #pythran python prefix
+        pythran_data = (re.sub(r'#\s*pythran', '', pythran_data)
+                        .replace('#', ''))
+
+        SpecParser._parse(self, pythran_data)
+
```
serge-sans-paille: I really prefer tis version \o/

pythran/toolchain.py
serge-sans-paille: I'd say yes.

pythran/toolchain.py
```diff
@@ -343,18 +349,24 @@ def compile_pythranfile(file_path, output_file=None, module_name=None,
     """
     if not output_file:
         # derive module name from input file name
-        _, basename = os.path.split(file_path)
+        dirname, basename = os.path.split(file_path)
```
serge-sans-paille: dirname is unsued later, so better keep _?

pythran/toolchain.py
serge-sans-paille: still no test case;-)

serge-sans-paille:
Moved to #656

PR #649, opened 2017-03-20
Optional

diorcety:


pythran/pythonic/types/NoneType.hpp
```diff
@@ -340,6 +316,23 @@ namespace pythonic
   }
 
   template <class T>
+  bool from_python<types::none<T>>::is_convertible(PyObject *obj)
```
serge-sans-paille: why do you need this? If I remember well, you either have a None, and then it's a none_type, or an object and then it's a plain object?

pythran/spec.py
```diff
@@ -53,16 +53,17 @@ class SpecParser:
...
     # token <> regexp binding
-    t_CRAP = r'[^,:\(\)\[\]]'
+    t_CRAP = r'[^,:\(\)\[\]\*]'
     t_COMMA = r','
     t_COLUMN = r':'
     t_LPAREN = r'\('
     t_RPAREN = r'\)'
     t_RARRAY = r'\]'
     t_LARRAY = r'\['
+    t_OPTIONAL = r'\*'
```
serge-sans-paille: I'd rather use either a `?` or the keyword `option` or `optional`, why the `*`?

pythran/tests/test_none.py
```diff
@@ -36,3 +37,15 @@ def none_input(a):
...
+    def test_use_optional(self):
+        code = """
+            def none_input(a):
+                if a is None:
+                  return 0.0
+                else:
+                  return a
+            """
+        self.run_test(code, 2.0, none_input=[Optional(float64)])
+        self.run_test(code, None, none_input=[Optional(float64)])
```
serge-sans-paille: ?? the syntax should be `Optional[float64]` I'm surprised this works.

diorcety: I copy past the previous test :D

serge-sans-paille:
Obsoleted by #659

PR #650, opened 2017-03-20
Multitype

diorcety:


pythran/spec.py
```diff
@@ -29,6 +29,7 @@ class SpecParser:
     reserved = {
         '#pythran': 'PYTHRAN',
         'export': 'EXPORT',
+        'or': 'PIPE',
```
serge-sans-paille: That's misleading, just use OR rather than PIPE

pythran/spec.py
```diff
@@ -151,8 +153,13 @@ def p_opt_types(self, p):
         p[0] = p[1] if len(p) == 2 else []
 
     def p_types(self, p):
-        '''types : type
-                 | type COMMA types'''
+        '''types : multitype
+                 | multitype COMMA types'''
```
serge-sans-paille: It looks like this commit supersedes the optional stuff, why should we propose two ways to achieve the same goal?

pythran/spec.py
```diff
@@ -300,6 +308,25 @@ def spec_expander(args):
     return all_specs
 
 
+def multiexports_to_exports(exports):
+    def decompose(value, arr, new_array=[]):
```
serge-sans-paille: you have a side effect on new_array *and* it's a default parameter. That's dangerous because of that http://docs.python-guide.org/en/latest/writing/gotchas/ even if that looks ok in your case. Please don't.

pythran/spec.py
```diff
@@ -300,6 +308,25 @@ def spec_expander(args):
...
+        if len(new_array) == len(value):
+            arr.append(new_array)
+            return
+        for elem in value[len(new_array)]:
+            a_array = list(new_array)
+            a_array.append(elem)
+            decompose(value, arr, a_array)
```
serge-sans-paille: This looks like itertools.product to me

pythran/tests/cases/multitype.py
```diff
@@ -0,0 +1,6 @@
+#pythran export square(int or float)
+#runas square(200)
+#runas square(200.200)
+
+def square(n):
+    return n*n
```
serge-sans-paille: maybe you could take an existing test case that could be rewritten using your notation?

pythran/spec.py
diorcety: indeed I will fix that part

pythran/spec.py
diorcety: indeed I will fix that part

serge-sans-paille:
@diorcety there a re a few design issues in your grammar, I'm working on it :-)

serge-sans-paille:
Obsoleted by #654

PR #654, opened 2017-03-28
Simplify spec declaration through the or operator

serge-sans-paille:
Make it possible to declare union type to declare multiple overload in a
convenient manner.

Original idea from Yann Diorcet, revisited by serge-sans-paille.


diorcety:
👍

PR #659, opened 2017-04-05
Feature/denone

serge-sans-paille:


serge-sans-paille:
@diorcety can you check this solves your issues with None?

serge-sans-paille:
@diorcety thanks for the live feedback. This should be okay now?

diorcety:
Yes, it's 👍

serge-sans-paille:
This is definitievely not the way too go, fat too hacky, I'll put this one in pause and rework it later.

serge-sans-paille:
Obsoleted by #939

PR #661, opened 2017-04-07
Pythran file: Skip last directive if there is no eol

diorcety:
Here multiples commits for expose and fix the issue.
The issue can't be triggered with the tests because we didn't take pythran file as directives source.
If there is no directive there is no test.
There is still a "WARNING:root:No pythran specification, no function will be exported" in the file /pythran/tests/cases/monte_carlo_pricer.py
(no pythran directive)


serge-sans-paille:
The regexp can be improved: https://travis-ci.org/serge-sans-paille/pythran/jobs/219740097

diorcety:
Sure. It was a quick demo more than a PR

serge-sans-paille:
Obsoleted by #662

PR #684, opened 2017-05-31
Fix/xrange/python3

serge-sans-paille:


serge-sans-paille:
@gouarin this is the PR I was talking about

Issue #687, opened 2017-06-14
Unable to configure CC and CXX in pythranrc and gcc 7 bug with pythran+numpy


ashwinvis:
## System configuration
- Arch Linux
- gcc-7.1.1 and gcc-5.4.0
- numpy==1.13.0
- pythran==0.8.0
- python==2.7.13

First of all, allow me to rant that it should be mentioned somewhere in the README that pythran only works in a medium version gcc > 4.9 and gcc < 6(??). I lost quite a lot of time trying to pythranize with gcc==4.7.2 and gcc==7.1.1 (super old and super new) - does not work at all (for traceback, see https://paste.pound-python.org/show/bA0TjMrAWP54PcKOU375/ ).Thanks to folks at IRC channel #pythran I was able to figure this out. 
 
## Issue

Pythran works when I try

    $ CC=gcc-5 CXX=g++-5 pythran nd_local_maxima.py

on the test case.

However when I try to make this choice of compiler permanent by modifying your pythran-linux2.cfg template and putting it in $HOME/.pythranrc as follows:
````
[compiler]
defines=USE_GMP
undefs=
include_dirs=
libs=gmp gmpxx
library_dirs=
cflags=-std=c++11 -fno-math-errno -w -fwhole-program -fvisibility=hidden
ldflags=-fvisibility=hidden -Wl,-strip-all
blas=cblas
CC=gcc-5
CXX=g++-5
````
Note that, leaving `blas` option empty gave me some other weird error. On running:

    $ pythran nd_local_maxima.py -v

yields a type error.

````python
running build_ext
running build_src
build_src
building extension "nd_local_maxima" sources
build_src: building npy-pkg config files
new_compiler returns distutils.unixccompiler.UnixCCompiler
INFO     customize UnixCCompiler
customize UnixCCompiler using build_ext
********************************************************************************
distutils.unixccompiler.UnixCCompiler
linker_exe    = u'gcc-5'
compiler_so   = u'gcc-5 -fno-strict-aliasing -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -DNDEBUG -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fPIC'
archiver      = ['ar', 'rc']
preprocessor  = u'gcc-5 -E'
linker_so     = ['gcc', '-pthread', '-shared', '-Wl,-O1,--sort-common,--as-needed,-z,relro']
compiler_cxx  = u'g++-5'
[...]

PR #688, opened 2017-06-15
Fix for issue #687

ashwinvis:
It seems like environment variables have to be set as ASCII, while configparser reads them as unicode. I stumbled over this when simply importing `PythranExtension` was affecting Cython extensions as well.


serge-sans-paille:
great, can you just squash theose in a single commit?

ashwinvis:
Done!

serge-sans-paille:
ok, letting travis validate that and I'll commit

serge-sans-paille:
thanks!

PR #689, opened 2017-06-22
Boost simd upgrade - 4.17.6.0

serge-sans-paille:


serge-sans-paille:
cc @jfalcou btw (no issue spotted)

PR #690, opened 2017-06-24
Delay typing until an error is found

serge-sans-paille:
This makes correct compilation much faster, without giving up with
typing error.

### DIFF
--- a/pythran/toolchain.py
+++ b/pythran/toolchain.py
@@ -112,7 +112,10 @@ class HasArgument(ast.NodeVisitor):
 
 def generate_cxx(module_name, code, specs=None, optimizations=None):
     '''python + pythran spec -> c++ code
-    returns a PythonModule object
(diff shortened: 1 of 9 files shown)


serge-sans-paille:
@pbrunet thanks for inspiring me this one!

pythran/toolchain.py
```diff
@@ -254,21 +263,47 @@ def compile_cxxfile(module_name, cxxfile, output_binary=None, **kwargs):
...
+                               if logger.isEnabledFor(logging.INFO)
+                               else '--quiet',
+                               'build_ext',
+                               '--build-lib', builddir,
+                               '--build-temp', buildtmp,
+                               ]
+                  )
+        except SystemExit as e:
+            raise CompileError(str(e))
+
```
serge-sans-paille: @pbrunet if you have time and a ready windows machine, could you try this is supported on Windows ?

Issue #691, opened 2017-07-05
Mention numpy in setup.py extras_require to allow using pythran in a tox test


ashwinvis:
The absence of `numpy` as a dependency is preventing us from testing `pythran` in our `tox` test. See travis build:

https://travis-ci.org/fluiddyn/fluidimage/jobs/250268847

Our tox configuration looks like this now:

```
[tox]
envlist =
    py{27,34,35,36}
    py{27,36}-pythran
    codecov

[testenv]
sitepackages = True
whitelist_externals = make
usedevelop = True
deps =
    coverage
    numpy
    matplotlib
    scipy
    psutil
    future
    pyfftw
    h5py
    ipython
    pillow
    scikit-image
    pythran: pythran
commands =
make tests_coverage
```

I propose adding an `extras_require` line in `setup` function call in pythran `setup.py` like this:
```python
 extras_require=dict(
          deps=['numpy'])
```

which should allow us to write the tox test as follows and should in practice `pip install numpy` before pythran:
```
    pythran: pythran[deps]
```

ashwinvis:
Also change `raise` in https://github.com/serge-sans-paille/pythran/blob/master/setup.py#L25 to a warning, perhaps?

ashwinvis:
At least pythran installs now:

https://travis-ci.org/fluiddyn/fluidimage/jobs/250296924

PR #695, opened 2017-07-05
setup.py: Numpy as an extra dependency. Change raise to warn if numpy

ashwinvis:
To fix issue #691

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,12 +17,11 @@ except ImportError:
 try:
     import numpy
 except ImportError:
-    print()
-    print("******************************************************************")
-    print("* Numpy must be installed before running setup, sorry about this *")
-    print("******************************************************************")
-    print()
-    raise
+    from warnings import warn
+    warn(("\n******************************************************************"
+          "\n* Numpy must be installed before running setup, sorry about this *"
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -192,5 +191,6 @@ def run(self, *args, **kwargs):
                      'pythran-config = pythran.config:run']},
       setup_requires=["pytest-runner"],
       tests_require=['pytest', 'pytest-pep8'],
+      extras_require={'deps': ['numpy']},
```
serge-sans-paille: would it work if it was part of setup_requires instead?

ashwinvis: It would work:
- if the user were to run `pip install -e .` which should install `numpy` using pip

and would not:
- if the user runs `python setup.py install` or `python setup.py develop` where easy_install is used.

I added it as an extra dep. because I assumed [you made this decision with good intentions](
https://github.com/ashwinvis/pythran/blob/7df4daa8011d4832bafbd54cc0b53b19ac052fb1/setup.py#L16). The problem arises when numpy is installed with `easy_install` yes?

Alternatively numpy can be added to the `install_requires` argument, which is processed when `pip` runs.

PR #701, opened 2017-07-24
Remove obsolete reference to nt2 files

serge-sans-paille:
This include a file that was only included in debug mode, and debug mode
is never tested...


scoder:
Fixes the NDEBUG compilation problem for me.

PR #706, opened 2017-08-28
Implement numpy.flip

diorcety:


serge-sans-paille:
@diorcety there's only one validation issue: you should apply clang-format-3.6 to you cpp files.

serge-sans-paille:
Thxs! Did you test the perf compared to numpy.flip?

serge-sans-paille:
ok, it turns out flip is implemented as an array view so O(1) in numpy. We are much slower....

diorcety:
Oh, I didn't think to that. I started from the roll source, i didn't know that was done by view and how to do this in pythran

Issue #709, opened 2017-08-29
cumsum not same behaviour


diorcety:
Here the code for reproduce
```
#pythran export f(float[])
#pythran export f(float[][])
def f(x):
    return np.cumsum(x, axis=0)
```

```
import diff
import numpy as np
from math import factorial

ROI_size = 3 
N = ROI_size**2

sample_spacing = 0.0000001
p = np.arange(0.,1.,sample_spacing)

sample_size = len(p)

proba = np.zeros( (N+1, sample_size) )
for i in xrange(N+1):
   x = 2 
   binomial_coef = 1. * factorial(N) / factorial(i) / factorial(N-i)
   proba[i, x] = binomial_coef * p[x]**i * (1-p[x])**(N-i)

print(["%.32f" % x for x in proba[:,2][0:10]])
cumulated = diff.f(proba)
print(["%.32f" % x for x in cumulated[:,2][0:10]])
```

Python output

> ['0.99999820000143990750274269885267', '0.00000179999712000201578593274084', '0.00000000000143999798400120935055', '0.00000000000000000067199919360040', '0.00000000000000000000000020159980', '0.00000000000000000000000000000004', '0.00000000000000000000000000000000', '0.00000000000000000000000000000000', '0.00000000000000000000000000000000', '0.00000000000000000000000000000000']
> ['0.99999820000143990750274269885267', '0.99999999999855992971475870945142', '0.99999999999999988897769753748435', '0.99999999999999988897769753748435', '0.99999999999999988897769753748435', '0.99999999999999988897769753748435', '0.99999999999999988897769753748435', '0.99999999999999988897769753748435', '0.99999999999999988897769753748435', '0.99999999999999988897769753748435']

Pythran output

> ['0.99999820000143990750274269885267', '0.00000179999712000201578593274084', '0.00000000000143999798400120935055', '0.00000000000000000067199919360040', '0.00000000000000000000000020159980', '0.00000000000000000000000000000004', '0.00000000000000000000000000000000', '0.00000000000000000000000000000000', '0.00000000000000000000000000000000', '0.00000000000000000000000000000000']
[...]

PR #712, opened 2017-09-06
Incorrect cumsum operation

diorcety:
Fix the bug reported in #709


serge-sans-paille:
LGTM, except clang-format is not happy :-/

PR #721, opened 2017-09-18
Fix leak when creating tuple

### LINKED ISSUE #720: Memory leak?
I encounter a problem which seems to be related to memory when I use this function

```
# pythran export myfunc(float64[][])

def myfunc(field):
    return 2*field, 1
```

with this code:

```
import numpy as np
a = np.ones((1024,)*2, dtype=np.float64)

for i in range(2000):
    myfunc(a)
```

The memory consumption increases a lot and the process is stopped by the kernel with a nice explanation:

"Killed"

pythran --version
0.8.2

Same thing with python 2.7, 3.5 and 3.6.

serge-sans-paille:
PyTuple_New does *not* steal references.

Fix #720

### DIFF
--- a/pythran/pythonic/include/types/ndarray.hpp
+++ b/pythran/pythonic/include/types/ndarray.hpp
@@ -236,9 +236,10 @@ namespace pythonic
       array<long, N> _strides; // strides
 
       /* mem management */
-      void mark_memory_external()
+      void mark_memory_external(extern_type obj)
       {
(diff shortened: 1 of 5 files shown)


serge-sans-paille:
@paugier I did a few more fixes related to leaks, can you confoirm this one is OK too?

paugier:
Yes I confirm that it works fine with the new fixes.

PR #725, opened 2017-09-30
Improved cython support

serge-sans-paille:
better to_python/from_python signatures


serge-sans-paille:
according to @aguinet this now compiles but the result is incorrect. I need to update the tests suite to showcase that.

serge-sans-paille:
@aguinet Just found the issue: you're converting extended slices calls to ``a[b,c]`` which in C++ calls the ``operator[]`` with **one** argument, using the comma as the sequence operator and not the argument separator. Yopu should generate ``a(b,c)`` instead. Can you confirm it works in that setup?

PR #737, opened 2017-11-05
Export native function pointer through __natives__

serge-sans-paille:
This can be used by third-party library like scipy.LowLevelCallable to
avoid conversion costs.


serge-sans-paille:
@maartenbreddels can you confirm it works for you test case? I still need to implement proper support for raw pointers as parameters in Pythran but that's for another PR.

PR #741, opened 2017-11-14
Support for mix/max reduction in OpenMP reduction

serge-sans-paille:


serge-sans-paille:
cc @diorcety this should fix the issue we had during the live session.

PR #744, opened 2017-11-24
Fix pythran file loading

diorcety:
load_specfile doesn't return a dictionnary anymore which produces an issue here


serge-sans-paille:
@diorcety can you add a test case (maybe in the docstring using doctest ?) so that we don't fall in that trap again?

serge-sans-paille:
Obsoleted by #748

PR #770, opened 2018-01-20
Feature/test scipy kernel

serge-sans-paille:


serge-sans-paille:
@aguinet : the test cases imported from scipy also provide a good set of test cases for cython integration! The original .pyx files can be found in the scipy official repo, I ported (some of) them to numpy-style.

PR #792, opened 2018-02-15
Correct type inference when combining two lists

### LINKED ISSUE #791: List concatenation
The following simple example doesn't compile using pythran

```
import itertools

# pythran export efgh(int32[][] list, int32[][] list)
def efgh(a, b):
  #return itertools.chain(a, b)
  return a + b
```

serge-sans-paille:
Fix #791


serge-sans-paille:
Thanks for reporting!

Issue #786, opened 2018-02-07
jupyter notebook magic: is it possible to hide warnings generated by Pythran?


flothesof:
Hi @serge-sans-paille,

thanks for continuing to put a lot of work into Pythran.
I'm mostly using Pythran from the Jupyter Notebook, using the nice %%pythran.magic command.
However, this always leads to a lot of output that I don't want to see in my notebook, that is, mostly warnings from the compiler. 
An example here can be found here on my blog: [http://flothesof.github.io/optimizing-python-code-numpy-cython-pythran-numba.html](http://flothesof.github.io/optimizing-python-code-numpy-cython-pythran-numba.html).
The output is full with stuff like: 
```
In file included from /var/folders/r9/s1k2nd110xzfxyfym7jmyb8r0000gn/T/tmp89u4ysag.cpp:9:
In file included from /Users/kappamaki/anaconda/lib/python3.6/site-packages/pythran/pythonic/include/types/ndarray.hpp:12:
In file included from /Users/kappamaki/anaconda/lib/python3.6/site-packages/pythran/pythonic/include/utils/broadcast_copy.hpp:4:
/Users/kappamaki/anaconda/lib/python3.6/site-packages/pythran/pythonic/include/types/tuple.hpp:63:5: warning: struct template 'numpy_gexpr' was previously declared as a class template [-Wmismatched-tags]
    struct numpy_gexpr;
    ^
/Users/kappamaki/anaconda/lib/python3.6/site-packages/pythran/pythonic/include/utils/numpy_traits.hpp:19:11: note: previous use is here
    class numpy_gexpr;
          ^
In file included from /var/folders/r9/s1k2nd110xzfxyfym7jmyb8r0000gn/T/tmp89u4ysag.cpp:9:
In file included from /Users/kappamaki/anaconda/lib/python3.6/site-packages/pythran/pythonic/include/types/ndarray.hpp:12:
In file included from /Users/kappamaki/anaconda/lib/python3.6/site-packages/pythran/pythonic/include/utils/broadcast_copy.hpp:4:
/Users/kappamaki/anaconda/lib/python3.6/site-packages/pythran/pythonic/include/types/tuple.hpp:373:3: warning: 'tuple_element' defined as a struct template here but previously declared as a class template [-Wmismatched-tags]
  struct tuple_element<I, pythonic::types::array<T, N>> {
  ^
[...]

PR #795, opened 2018-02-19
Correct support of develop target for setup.py

### LINKED ISSUE #783: `python setup.py develop` to work on pythran code?
Hi,

I'd like to work on pythran code and I try to use `python setup.py develop`. It would be nice to be able to use this instead of `python setup.py install`.

It does not work because g++ does not find header files.

I remember that I read something about this issue in the documentation but I can not find it anymore. Did I dream?

What is the recommend method to install pythran when one wants to modify its code?

serge-sans-paille:
Fix #783 and improves #786.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,8 @@ except ImportError:
          UserWarning)
 
 from setuptools.command.build_py import build_py
+from setuptools.command.develop import develop
+
 from setuptools import setup
 from distutils import ccompiler
 from distutils.errors import CompileError, LinkError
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
@paugier and @flothesof can you confirm this fixes the ``python setup.py develop`` target? Thanks!

paugier:
I confirm that `python setup.py develop` works well for me. Very convenient!

flothesof:
I just tested this from a clean conda environment where I checked out this branch. It worked fine on the dprod example from the Readme. So I can confirm it works on Mac. 

Thanks!

serge-sans-paille:
Thanks @flothesof and @paugier for the feedback!

Issue #800, opened 2018-03-02
issue with axis argument to argmax/argmin


cycomanic:
Not sure if I'm doing something wrong but I'm trying to do some work where I need to get the argmax/argmin values along one axis of a 2D array. However I'm getting compilation errors. The simplest test case is:
```python
import numpy as np

#pythran export testargmax(float[][])
def testargmax(x):
    return x.argmax(axis=1)
```
which results in a long list of  `template argument deduction/substitution failed:` errors with the following sections:
```
/home/jschrod/.local/lib/python3.6/site-packages/pythran/pythonic/utils/functor.hpp:13:22: error: no matching function for call to ‘argmax({anonymous}::pythonic::types::numpy_texpr<{anonymous}::pythonic::types::ndarray<double, 2> >, long int)’
         -> decltype(f(std::forward<Types>(types)...))                          \
```
if I understand this correctly pythran assumes it should get a long int not an array from argmax? Same thing happens for argmin.

serge-sans-paille:
Thanks for the bug report, I can reproduce it, it's just that we don't
support argmax along an axis. I'll implemment that!

serge-sans-paille:
Can you confirm #803 fixes your issue?

cycomanic:
Great thanks!

cycomanic:
Sorry, my answer might be not clear enough, yes #803 fixes my issue

PR #803, opened 2018-03-06
Feature/argminmax axis

serge-sans-paille:


serge-sans-paille:
Should fix #800.

serge-sans-paille:
@cycomanic can you confirm that's ok for you, so that I can merge?

PR #808, opened 2018-03-09
Feature/dump optimized python

paugier:
Add a not very useful -P option to dump optimized Python code.

It is mainly to start to contribute and see how it works.

Serge, you told me that I could do something simpler by simply change the backend. I'm not sure to understand because I would have to use the Python backend in the function generate_cxx, which is maybe not so great.

I didn't find the tests for the command line. Should I add a test for this -P option? Where?

### DIFF
--- a/pythran/toolchain.py
+++ b/pythran/toolchain.py
@@ -3,7 +3,7 @@ This module contains all the stuff to make your way from python code to
 a dynamic library, see __init__.py for exported interfaces.
 '''
 
-from pythran.backend import Cxx
+from pythran.backend import Cxx, Python
 from pythran.config import cfg, make_extension
 from pythran.cxxgen import PythonModule, Define, Include, Line, Statement
 from pythran.cxxgen import FunctionBody, FunctionDeclaration, Value, Block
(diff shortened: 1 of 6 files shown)


pythran/toolchain.py
```diff
@@ -391,7 +419,7 @@ def compile_pythrancode(module_name, pythrancode, specs=None,
 
 
 def compile_pythranfile(file_path, output_file=None, module_name=None,
-                        cpponly=False, **kwargs):
+                        cpponly=False, pyonly=False, **kwargs):
```
serge-sans-paille: seeing that change, I see a pattern emerging :-) I'll formalize that and issue a PR against your PR!

serge-sans-paille:
squashed, slightly reworked (but keeping the original design and authorship) and commited as 798db7c557c080b280695c5438f102bb5043b69d

serge-sans-paille:
Thanks @paugier ! I removed the .gitignore and test_exception diff as they were not related to the PR, and kept the remaining. I need to find the right abstraction for the different compilation stages.

PR #816, opened 2018-03-14
Fix/cleanup toolchain

serge-sans-paille:


serge-sans-paille:
@paugier FYI, this updates your previous commit

PR #817, opened 2018-03-16
Add optimization for replacing identities of binary operators.

sbacchio:
Hi,

I've casually find out pythran and it's a very interesting project. Thank you for all the effort you have put in it. I was experimenting a bit and trying to understand if it can be useful for the project I have in mind.
The structure is nice and easy to develop with it.

I've tried to implement an optimization which replace identities of the operators (BinOp at the moment). I tried to give a generic structure such that many identities can be implemented easily. Similarly in the optimization Square you have the function expand_pow which replace x**0 and x**1. This can be moved here for instance.

The pull request is not complete yet, but I wanted first to ask if it's something needed and the implementation is up-to-standard.

Cheers,
Simone

### DIFF
--- a/pythran/optimizations/__init__.py
+++ b/pythran/optimizations/__init__.py
@@ -22,6 +22,7 @@ from .pattern_transform import PatternTransform
 from .range_loop_unfolding import RangeLoopUnfolding
 from .range_based_simplify import RangeBasedSimplify
 from .square import Square
+from .binop_identities import BinOpIdentities
 from .inlining import Inlining
 from .inline_builtins import InlineBuiltins
 from .list_to_tuple import ListToTuple
--- a/pythran/optimizations/binop_identities.py
+++ b/pythran/optimizations/binop_identities.py
@@ -1,0 +1,79 @@
+""" Replaces identities of binary operators """
+
+from pythran.passmanager import Transformation
+from pythran.analyses.ast_matcher import ASTMatcher, AST_any
+
+import gast as ast
+
+
+class BinOpIdentities(Transformation):
+
+    """
+    Replaces identities of binary operators
+
+    >>> import gast as ast
+    >>> from pythran import passmanager, backend
+    >>> from pythran.optimizations import BinOpIdentities
+    >>> node = ast.parse('b = a * 0')
+    >>> pm = passmanager.PassManager("test")
+    >>> _, node = pm.apply(BinOpIdentities, node)
+    >>> print(pm.dump(backend.Python, node))
+    b = 0
+    >>> node = ast.parse('b = a * 1')
+    >>> pm = passmanager.PassManager("test")
+    >>> _, node = pm.apply(BinOpIdentities, node)
+    >>> print(pm.dump(backend.Python, node))
+    b = a
+    """
+
+    """ 
+    Identites for binary operators are implemented as:
+      IDs[ast.OPERATION.__name__] = ( ( A, B, C) )
+    i.e
+      ast.BinOp(left=A, op=OPERATION(), right=B) is replaced with C
+
+    A or B can be AST_any(). 
+    If C is AST_any(), then the node will be replaced with node.left
+    or node.right depending if A or B was AST_any(), respectively.
+
+    The implemented identities are
+       x * 0 = 0
+       0 * x = 0
+       x * 1 = x
+       1 * x = x
+       x + 0 = x
+       0 + x = x
+       x - 0 = x
+    """
+    IDs = {}
+    IDs[ast.Mult.__name__] = ( (ast.Num(0), AST_any(), ast.Num(0)),
+                               (AST_any(), ast.Num(0), ast.Num(0)),
+                               (AST_any(), ast.Num(1), AST_any()),
+                               (ast.Num(1), AST_any(), AST_any()), )
+
+    IDs[ast.Add.__name__] = ( (ast.Num(0), AST_any(), AST_any()),
+                              (AST_any(), ast.Num(0), AST_any()), )
+
+    IDs[ast.Sub.__name__] = ( (AST_any(), ast.Num(0), AST_any()), )
+
+    def __init__(self):
+        Transformation.__init__(self)
+        
+    def visit_BinOp(self, node):
+        self.generic_visit(node)
+        if node.op.__class__.__name__ in self.IDs:
+            for identity in self.IDs[node.op.__class__.__name__]:
+                if ASTMatcher(ast.BinOp(identity[0], node.op, identity[1])).search(node):
+                    replace = identity[2]
+                    if isinstance(replace, AST_any):
+                        if isinstance(identity[0], AST_any) and isinstance(identity[1], AST_any):
+                            print("Error: identity (AST_any(), AST_any(), AST_any()) not allowed.")
+                            replace = node
+                        elif isinstance(identity[0], AST_any):
+                            replace = node.left
+                        elif isinstance(identity[1], AST_any):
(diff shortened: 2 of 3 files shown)


serge-sans-paille:
Hi sabbachio and thanks for the on-going PR.
The idea is legitimate, but a few stuff could be improved / changed.

First your proposal is a ntural extension of

```
pythran.optimizations.PatternTransform
```

So you' be better register tour patterns here.

Then considering the validity of the transformations :

```
x * 0 != 0
```
think of x == [1,2] or x == 'er'.

but you could workaround that using
pythran/pythonic/include/utils/neutral.hpp and the netural element ofr the addition.

remember pythran optimizations are type agnostic, as the generated program must keep polymorphism.

serge-sans-paille:
Multiplication by 1 looks always foldable to me, and it does not interfere with type inference

```
In [2]: numpy.array([1,2], dtype=numpy.uint8) * 1
Out[2]: array([1, 2], dtype=uint8)
```

sbacchio:
Ok, I missed PatternTransform.
Shall I simply add the cases one by one explicitely, as it is done now, e.g.
```
    # X * 1 => X                                                                                              
    (ast.BinOp(left=Placeholder(0), op=ast.Mult(), right=ast.Num(n=1)),
     lambda: Placeholder(0)),
```
or try construct a more compact method as I did, but based on Placeholder?
For instance something which uses:
```
    known_patterns_BinOp[ ast.Mult() ] = [ ( Placeholder(0), ast.Num(n=1), Placeholder(0) ),
                                                                  ( ast.Num(n=1), Placeholder(0), Placeholder(0) ),
                                                                  etc ]
```

sbacchio:
For clarity sake, the implementation I would like to carry out is an optimization of matrix products with a given sparse matrix (e.g transformation).

For instance the code:
```
A = [[0, 0, 1, 0],
     [0, 0, 0, 1],
     [1, 0, 0, 0],
     [0, 1, 0, 0]]

#pythran export apply_A(float[ :, 4])
def apply_A ( vector ):
    return vector[:].dot(A)
```
Should give something like
```
def apply_A ( vector ):
    tmp = numpy.empty_like(vector)
    for i in range(len(vector)):
       tmp[i,0] = vector[i,2]
       tmp[i,1] = vector[i,3]
       tmp[i,2] = vector[i,0]
       tmp[i,3] = vector[i,1]

    return tmp
```
which then the compiler optimize easily.
Now.. there is quite a work to do, but I think the result is feasible with your code.

serge-sans-paille:
Yes for the palceholders!

Your test case is very interesting :-) part of the job should feasible, you alreay have constant forlding, dead code elimination and loop unrolling. 

We don't have support for constant array shape (in the pythran export) but that's something I should be able to do.

pythran/optimizations/pattern_transform.py
```diff
@@ -66,27 +66,37 @@
...
-         keywords=[])),
 ]
 
+# Dictionary with ast operator name as keys for each a list of tuple of
+#   (left, right, replacement) is defined.
+# replacement have to be a lambda function to have a new ast to replace when
+# replacement is inserted in main ast
+know_pattern_BinOp = {
+    ast.Mult.__name__ : [
+        (Placeholder(0), ast.Num(1), lambda: Placeholder(0)), # X * 1 => X
```
serge-sans-paille: Unfortunately, I relaized that even this simple transformation is not always correct:

```
x = [1,2]
y = x * 1
y[0] = 2
print(x, y)
```

pythran/optimizations/pattern_transform.py
```diff
@@ -66,27 +66,37 @@
...
+        (ast.Num(1), Placeholder(0), lambda: Placeholder(0)), # 1 * X => X
+        (Placeholder(0), Placeholder(0), lambda: ast.BinOp(left=Placeholder(0), op=ast.Pow(), right=ast.Num(n=2))), # X * X => X ** 2
```
serge-sans-paille: This case is already handled by ``pythran.optimization.square``

pythran/optimizations/pattern_transform.py
```diff
@@ -66,27 +66,37 @@
...
+    ],
+    ast.Add.__name__ : [
+        (Placeholder(0), ast.Num(0), lambda: Placeholder(0)), # X + 0 => X
+        (ast.Num(0), Placeholder(0), lambda: Placeholder(0)), # 0 + X => X
```
serge-sans-paille: same copy issue as above, when one multiplies by one.

pythran/optimizations/pattern_transform.py
```diff
@@ -66,27 +66,37 @@
...
+        (   # a + "..." + b => "...".join((a, b))
```
serge-sans-paille: This one looks correct :-)

pythran/optimizations/pattern_transform.py
```diff
@@ -66,27 +66,37 @@
...
+            lambda: ast.Call(func=ast.Attribute(
+                ast.Attribute(ast.Name('__builtin__', ast.Load(), None),'str',ast.Load()),'join', ast.Load()),
+                args=[ast.Str(Placeholder(1)),ast.Tuple([Placeholder(0), Placeholder(2)], ast.Load())],
+                             keywords=[])
+        ),
+    
+    ],
+    ast.Sub.__name__ : [
+        (Placeholder(0), ast.Num(0), lambda: Placeholder(0)), # X - 0 => X
+        (ast.Num(0), Placeholder(0), lambda: ast.UnaryOp(op=ast.USub(), operand=Placeholder(0))), # 0 - X => -X
```
serge-sans-paille: Same copy issue here.

pythran/optimizations/pattern_transform.py
```diff
@@ -125,3 +135,21 @@ def visit(self, node):
                 node = PlaceholderReplace(check.placeholders).visit(replace())
                 self.update = True
         return super(PatternTransform, self).visit(node)
+
+    def visit_BinOp(self, node):
+        """ 
+        Special method for BinOp.
+        Try to replace if node match the given pattern. 
+        """
```
serge-sans-paille: why do you need that specialization?

sbacchio: I thought would be good for speeding-up the optimization.
As soon as I find a way to implement safely the patterns I have in mind, we may have order of  100 patterns to check for replacement.  
Then, as it was,  pattern transform would check for any node any given pattern careless of the structure.
In this way, instead, it goes trough BinOp just when needed and select from the dictionary the patterns relative to the used operator.

pythran/optimizations/pattern_transform.py
sbacchio: Arg, damn python.. I need to find a way to make it safe...
For my test case I need somehow the replacements 
```
y += x * 0 => y += 0 => pass
```

serge-sans-paille:
> y += x * 0 => y += 0 => pass

I'll think about it but capturing the first pattern as a whole looks safe to me.

```
y += x * 0 => pass
```

serge-sans-paille:
ok, your identity is false for floating point numbers:

```
float('nan') * 0
```
evaluates to ``nan``.

My guess is that you should give up with the pattern matching stuff and implement a new pattern in ``optimizations/inline_builtin`` that inlinies the np.dot call when one of the parameter is a constant. The remaining optimizations will be taken care of by the backend compiler, if you activate ``-ffast-math``.

I don't know if I'm clear...

sbacchio:
Sure sure, it's clear. I will look at that.

PR #818, opened 2018-03-16
Fix Windows compilation

diorcety:
Without these casts, VS2017 fails with a C2397 and C2398.
Indeed pointer difference is not natively a long on Windows X64.
We need to do a static_cast in order to avoid these issues.


serge-sans-paille:
clang-formtaed and merged as cbd8faeb91119eb85be65dff3894c2930e7b4fcb. Thanks!

PR #835, opened 2018-03-30
Support from .X import Y

diorcety:
Work as-is with setup.py mechanism, for command line invocation a package argument can be provided in order to specify the current package of the "compiled" python file

### DIFF
--- a/pythran/run.py
+++ b/pythran/run.py
@@ -38,6 +38,8 @@ def compile_flags(args):
         'library_dirs': args.libraries_dir,
         'extra_link_args': args.extra_flags,
     }
+    if args.package:
+        compiler_options['package'] = args.package
     if args.opts:
         compiler_options['opts'] = args.opts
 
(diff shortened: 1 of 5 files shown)


pythran/run.py
```diff
@@ -82,6 +84,9 @@ def run():
                         'generation',
                         default=list())
 
+    parser.add_argument('-C', dest='package', type=str,
```
serge-sans-paille: That's a strange option name; Maybe only accept long option?

pythran/toolchain.py
```diff
@@ -451,6 +454,8 @@ def compile_pythranfile(file_path, output_file=None, module_name=None,
 
     # Add compiled module path to search for imported modules
     sys.path.append(os.path.dirname(file_path))
+    if package is not None:
+        sys.path.append(os.getcwd())
```
serge-sans-paille: I don't get this one.

serge-sans-paille:
You should add a few test case in pythran/tests/test_user_defined_import.py

pythran/run.py
diorcety: sure

pythran/toolchain.py
diorcety: This make the setup.py hook work nice :)

diorcety: This need improvements I know

serge-sans-paille: yeah :-)

serge-sans-paille:
Obsoleted by #913

PR #836, opened 2018-03-30
Fix loop optimization on python3 (xrange)

diorcety:
On python 3 xrange.__name__ = "range", so the optimisation is not done (and OMP doesn't work then)

### DIFF
--- a/pythran/backend.py
+++ b/pythran/backend.py
@@ -628,18 +628,25 @@ class CxxFunction(Backend):
             - order have to be known at compile time
         """
         assert isinstance(node.target, ast.Name)
-        pattern = ast.Call(func=ast.Attribute(
+        pattern_range = ast.Call(func=ast.Attribute(
             value=ast.Name(id='__builtin__',
                            ctx=ast.Load(),
                            annotation=None),
-            attr=xrange.__name__, ctx=ast.Load()),
+            attr='range', ctx=ast.Load()),
+            args=AST_any(), keywords=[])
+        pattern_xrange = ast.Call(func=ast.Attribute(
+            value=ast.Name(id='__builtin__',
+                           ctx=ast.Load(),
+                           annotation=None),
+            attr='xrange', ctx=ast.Load()),
             args=AST_any(), keywords=[])
         is_assigned = {node.target.id: False}
         [is_assigned.update(self.passmanager.gather(IsAssigned, stmt))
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
I tested the following snippet

```
def a(x):
    s = 0
    for i in range(x):
        s += i
    return s
```

on both python2 and python3 using master version and I don't understand why you need this commit. Can you provide me with a setup where it is needed?

diorcety:
I modified the PR. The issue is with xrange which is accepted by pythran by not correctly handled

with this example:

import numpy as np

```
#pythran export a(int)
def a(x):
    s = 0
    arr = np.empty((x,))
    # omp parallel
    for i in range(x):
        arr[i] = x *x
    # omp parallel
    for i in xrange(x):
        arr[i] = x *x
    return arr
```
Here the diff

```
-      typename pythonic::assignable<decltype(std::declval<typename std::remove_cv<typename std::remove_reference<decltype(pythonic::__builtin__::functor::xrange{})>::type>::type>()(std::declval<typename std::remove_cv<typename std::remove_reference<argument_type0>::type>::type>()))>::type __iter1 = pythonic::__builtin__::functor::xrange{}(x);
+      long  __target1 = x;
       #pragma omp parallel
-      for (typename pythonic::assignable<decltype(std::declval<typename std::remove_cv<typename std::remove_reference<decltype(pythonic::__builtin__::functor::xrange{})>::type>::type>()(std::declval<typename std::remove_cv<typename std::remove_reference<argument_type0>::type>::type>()))>::type::iterator __target1 = __iter1.begin(); __target1 < __iter1.end(); ++__target1)
+      for ( i=0L; i < __target1; i += 1L)
       {
-         i= *__target1;
-        {
-          arr.fast(i) = pythonic::numpy::functor::square{}(x);
-        }
+        arr.fast(i) = pythonic::numpy::functor::square{}(x);
       }
+      if (i == __target1)
+      i -= 1L;
     }
```

serge-sans-paille:
If I understand correctly, the problem is that xrange should *not* be accepted in pythran3, right?

diorcety:
Should be.

serge-sans-paille:
I'm planning to move to python3 soon, I think this can wait until then :-)

serge-sans-paille:
Closing as it only showcases the fact that pythran2 accepts some python2 constructs.

PR #845, opened 2018-04-05
Implement numpy.fft.rfft and numpy.fft.irfft

jeanlaroche:
This implements numpy.fft.rfft and numpy.fft irfft. It uses the original C code used by numpy.

### DIFF
--- a/pythran/pythonic/numpy/fft/irfft.hpp
+++ b/pythran/pythonic/numpy/fft/irfft.hpp
@@ -12,7 +12,7 @@
 #include "pythonic/numpy/swapaxes.hpp"
 #include <string.h>
 #include <math.h>
-
+#include <mutex>
 
 #include <stdio.h>
 #include "pythonic/numpy/fft/fftpack.c"
@@ -24,6 +24,7 @@ namespace numpy
 {
     namespace fft
     {
+        std::mutex mtx_irfft;           // mutex for critical section
         
         // Aux function
         template<class T, size_t N>
@@ -39,19 +40,20 @@ namespace numpy
             types::ndarray<double, N> out_array(out_shape,__builtin__::None);
             
             // Create the twiddle factors. These must be kept from call to call as it's very wasteful to recompute them.
-            // This is from fftpack_litemodule.c
-            static std::map<long, types::ndarray<double, 1>> all_twiddles;
-            if (all_twiddles.find(NFFT) == all_twiddles.end()) {
+            mtx_irfft.lock();
+            static std::map<long, types::ndarray<double, 1>> all_twiddles_irfft;
+            if (all_twiddles_irfft.find(NFFT) == all_twiddles_irfft.end()) {
                 // Insert a new twiddle array into our map and initialize it
-                all_twiddles.insert(std::make_pair(NFFT,types::ndarray<double, 1>({(long)(2*NFFT+15)},__builtin__::None)));
-                npy_rffti(NFFT, (double *)all_twiddles[NFFT].buffer);
+                all_twiddles_irfft.insert(std::make_pair(NFFT,types::ndarray<double, 1>({(long)(2*NFFT+15)},__builtin__::None)));
+                npy_rffti(NFFT, (double *)all_twiddles_irfft[NFFT].buffer);
             }
+            mtx_irfft.unlock();
             
             // Call fft (fftpack.py) r = work_function(in_array, wsave)
             // This is translated from https://raw.githubusercontent.com/numpy/numpy/master/numpy/fft/fftpack_litemodule.c
             
             double* rptr            = (double*) out_array.buffer;
-            double* twiddle_buffer  = (double*) all_twiddles[NFFT].buffer;
+            double* twiddle_buffer  = (double*) all_twiddles_irfft[NFFT].buffer;
             typename T::value_type* dptr =  (typename T::value_type*) in_array.buffer;
             long nrepeats           = out_array.flat_size()/out_size;
             long to_copy            = (NFFT/2+1 <= npts) ? (NFFT-1) : (2*npts-2);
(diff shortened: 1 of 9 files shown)


pythran/pythonic/numpy/fft/rfft.hpp
```diff
@@ -0,0 +1,129 @@
...
+            types::ndarray<std::complex<typename std::common_type<T, double>::type>, N> out_array(out_shape,__builtin__::None);
+            
+//            printf("Shape %d %d N = %d -- Flat size: %d NFFT = %d npts = %d out_size %d\n",
+//                   shape[0],shape[1],N,in_array.flat_size(),NFFT,npts,out_size);
+//            for (i = 0; i < NFFT; i++)
+//                printf("in: %.4f\n",dptr[i]);
+            
+            // Create the twiddle factors. These must be kept from call to call as it's very wasteful to recompute them.
+            // This is from fftpack_litemodule.c
+            static std::map<long, types::ndarray<double, 1>> all_twiddles;
```
serge-sans-paille: this is not thread safe, we shoyuld either use a lock or make it thread_private

pythran/pythonic/numpy/fft/rfft.hpp
```diff
@@ -0,0 +1,129 @@
...
+            if (all_twiddles.find(NFFT) == all_twiddles.end()) {
+//                printf("CREATING TWIDDLE FOR %d\n",NFFT);
+                types::array<long, 1> twiddle_shape = {(long)(2*NFFT+15)};
+                // Insert a new twiddle array into our map
+                all_twiddles.insert(std::make_pair(NFFT,types::ndarray<double, 1>(twiddle_shape,__builtin__::None)));
```
serge-sans-paille: could be emplace_back to avoid a copy

pythran/pythonic/numpy/fft/rfft.hpp
```diff
@@ -0,0 +1,129 @@
...
+                // Then initialize it.
+                npy_rffti(NFFT, (double *)all_twiddles[NFFT].buffer);
+            }
+            else {
+                //printf("RECALLING TWIDDLE FOR %d\n",NFFT);
```
serge-sans-paille: please remove these printf etc :-)

pythran/pythonic/numpy/fft/rfft.hpp
```diff
@@ -0,0 +1,129 @@
...
+            
+            double* rptr            = (double*) out_array.buffer;
+            double* twiddle_buffer  = (double*) all_twiddles[NFFT].buffer;
+            long nrepeats           = in_array.flat_size()/npts;
+            long to_copy            = (NFFT <= npts) ? NFFT : npts;
+            for (i = 0; i < nrepeats; i++) {
+                rptr[2*out_size-1] = 0.0; // We didn't zero the array upon allocation. Make sure the last element is 0.
+                std::copy(dptr,dptr+to_copy,rptr+1);
+                // Zero padding if the FFT size is > the number points
+                memset((char *)(rptr+1+to_copy),0,(NFFT-to_copy)*sizeof(double));
```
serge-sans-paille: could be a std::fill :-)

pythran/pythonic/numpy/fft/irfft.hpp
```diff
@@ -0,0 +1,104 @@
...
+PYTHONIC_NS_BEGIN
+
+namespace numpy
+{
+    namespace fft
+    {
+        
+        // Aux function
+        template<class T, size_t N>
+        types::ndarray<double, N> _irfft(types::ndarray<T, N> const &in_array, long NFFT, bool norm)
```
serge-sans-paille: this code is very similar to rfft, maybe you could aggregate stuff ?

pythran/tests/test_numpy_fft.py
```diff
@@ -0,0 +1,75 @@
+import unittest
+from test_env import TestEnv
+import numpy
+from pythran.typing import NDArray
```
serge-sans-paille: you must also edit the .travis.yml file to have travis run these tests, but I can take care of that if you want

pythran/pythonic/numpy/fft/irfft.hpp
jeanlaroche: Sadly not! :( It's different enough that that would be ugly. For one things, rfft goes from real -> complex. Irfft goes from complex -> real. The sizes are also quite different for the two functions (rfft goes from N -> N/2+1, irfft goes from N -> 2(N-1) The only thing that's really shared are the twiddle factors.

pythran/pythonic/numpy/fft/rfft.hpp
jeanlaroche: True. I'll look into it.

jeanlaroche:
BTW, the pythran implementation is faster than numpy by about 30% on 2048 point FFT, **except** when a transpose is needed (i.e., axis=0 instead of -1) in which case the pythran version is excruciatingly slower.
The call 
```
auto swapped_array = swapaxes(in_array, axis, N-1);
return swapaxes(_irfft(swapped_array,NFFT,norm),axis,N-1);
```
Is very slow. I'm not sure if I'm doing something I shouldn't do. This follows what happens in numpy (use of swapaxes). 

NUMPY:
```
python -m timeit -s 'import numpy;A=numpy.exp(1j*numpy.random.random((2048,2)))' 'numpy.fft.irfft(A,2048,0)'
10000 loops, best of 3: 65 usec per loop
```
PYTHRAN:
```
python -m timeit -s 'import numpy;import rfftFunc;A=numpy.exp(1j*numpy.random.random((2048,2)))' 'rfftFunc.testIFunc(A,2048,0,"no")'
1000 loops, best of 3: 314 usec per loop
```

jeanlaroche:
About swapaxis: the numpy doc indicates that: 

> For NumPy >= 1.10.0, if a is an ndarray, then a view of a is returned; otherwise a new array is created. For earlier NumPy versions a view of a is returned only if the order of the axes is changed, otherwise the input array is returned.

I believe the pythran implementation does a lot of copying instead so that's probably where the difference comes from.

serge-sans-paille:
Obsoleted by #856

PR #867, opened 2018-04-24
Fix crash when g_expr is used in lazy variable

### LINKED ISSUE #861: crash with gexpr
```
def efgh(a, y):
    _,_,_ = a # Comment it will not compile
    return np.zeros(102)[:-1]

def efgh2(s, m, x, y):
    if x == 0: # Without the if it doesn't compile
        return np.zeros(102)[:-1]
    return np.zeros(102)[:-1]

#pythran export ijkl(str, str, int, int)
def ijkl(s, m, x, y):

    ll = (s, m, x)
    if m == 'a':
        b = efgh(ll, y + 1)
    else:
        b = efgh(ll, y)

    return b

#pythran export ijkl2(str, str, int, int)
def ijkl2(s, m, x, y):

    ll = (s, m, x)
    b = efgh(ll, y)

    return b

#pythran export ijkl3(str, str, int, int)
def ijkl3(s, m, x, y):
    if m == 'a':
        b = efgh2(s, m, x, y + 1)
    else:
        b = efgh2(s, m, x, y)
    return b

#pythran export ijkl4(str, str, int, int)
def ijkl4(s, m, x, y):
    return efgh2(s, m, x, y)

```

```
import numpy as np
import aa
print(aa.ijkl("c", "b", 609, 610)) # crashs
print(aa.ijkl2("c", "b", 609, 610)) # works
print(aa.ijkl3("c", "b", 609, 610)) # crashs
print(aa.ijkl4("c", "b", 609, 610)) # works
```

diorcety:
Fix #861

### DIFF
--- a/pythran/pythonic/types/numpy_gexpr.hpp
+++ b/pythran/pythonic/types/numpy_gexpr.hpp
@@ -487,7 +487,15 @@ namespace types
   numpy_gexpr<Arg, S...> &numpy_gexpr<Arg, S...>::
   operator=(numpy_gexpr<Arg, S...> const &expr)
   {
-    return _copy(expr);
+    if (buffer == nullptr) {
+      arg = expr.arg;
+      slices = expr.slices;
+      buffer = arg.buffer;
+      _shape = expr.shape();
+      return *this;
+    } else {
+      return _copy(expr);
+    }
   }
 
   template <class Arg, class... S>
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
@diorcety there  are a few regression :-)

serge-sans-paille:
Obsoleted by #883

PR #877, opened 2018-04-25
Flake8 compliancy

serge-sans-paille:
Move checks to appveyor to remove some of travis burden, as suggested by
@cclauss.


docs/DEVGUIDE.rst
```diff
@@ -41,10 +41,12 @@ is to install Pythran once and to add
 Coding Style
 ------------
 
-All Python code must be conform to the PEP 8, and the ``pep8`` command must not
+All Python code must be conform to the PEP 8, and the ``flake8`` command must not
 yield any message when run on our database. Additionally, avoid backslashes,
 and try to make your code as concise as possible.
 
+    $> flake8 pythran/*.py pythran/*/*.py --exclude="pythran/tests/test*.py,__init__.py"
```
cclauss: Would it make sense to put __# noqa__ at the end of the offending lines in \_\_init\_\_.py instead of excluding it?

serge-sans-paille: I don't like coupling code to a given tool that way: the comment would be useless if we move away from flake8

PR #879, opened 2018-04-26
Fix OSX bug when dealing with numpy_gexpr

serge-sans-paille:
Probably due to std::tuple_cat misuse, but... who knows?


serge-sans-paille:
@jeanlaroche I did a few update to the patch, can you confirm it still works for you?

jeanlaroche:
I can confirm that the latest version of this branch still fixes bug 870. Is that what you wanted me to confirm? Or did you mean on master?

serge-sans-paille:
Thanks, that's what I wanted o/

PR #881, opened 2018-04-27
import unittest

cclauss:
Resolves undefined names in legacy tests

### DIFF
--- a/pythran/tests/openmp.legacy/omp_parallel_num_threads.py
+++ b/pythran/tests/openmp.legacy/omp_parallel_num_threads.py
@@ -1,3 +1,5 @@
+import unittest
+
 unittest.skip
 #segfault ....
 def omp_parallel_num_threads():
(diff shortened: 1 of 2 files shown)


pythran/tests/openmp.legacy/omp_parallel_num_threads.py
```diff
@@ -1,3 +1,5 @@
+import unittest
+
 unittest.skip
```
serge-sans-paille: As this is not used as a decorator, this actually has no effect... I'm worried by the fact that this never triggered a failure, need to check that :-/

cclauss:
If some other file has imported __unittest__ before this file is run then the global is set up and all is good at runtime.  However this is an implicit condition and not an explicit condition and we all know about what the Zen of Python says about that.  So it may work today but not tomorrow if the order of execution changes.

serge-sans-paille:
ok, but as the decorator is not used, maybe the good fix was just prunning ``unittest.skip``

PR #890, opened 2018-04-30
Remove -arch i386 if -arch x86_64 is present in the compilation flags

jeanlaroche:


serge-sans-paille:
@jeanlaroche can you confirm this is still necessary after the upgrade of your system?

jeanlaroche:
> @jeanlaroche can you confirm this is still necessary after the upgrade of your system?
As far as I can tell it's still necessary. I'm still getting the error if both flags are present. This is with 
Apple LLVM version 9.1.0 (clang-902.0.39.1)
which I believe is the current version of the compiler.

pythran/toolchain.py
```diff
@@ -53,6 +53,13 @@ def CCompiler_customize(self, _, need_cxx=0):
             self.compiler_so.remove('-Wstrict-prototypes')
         except (AttributeError, ValueError):
             pass
+    # Remove -arch i386 if 'x86_64' is specified.
+    try:
+        if 'x86_64' in ' '.join(self.compiler_so):
+            # Ugly shortcut to remove '-arch', 'i386' from the list of arguments:
```
serge-sans-paille: what about:

```
archs = defaultdict(list)
for i, flag in enumerate(self.compiler_so[1:]):
    if self.compiler_so[i] == '-arch':
        archs[flag].append(i)

if 'x86_64' in archs and 'i386' in archs:
    for i in archs['i386']:
        self.compiler_so[i] = 'x86_64'
```

or something similar?

jeanlaroche: OK I rewrote it according to your example.

serge-sans-paille:
Merged as b966fe6c

PR #901, opened 2018-05-11
Fix default argument offset in alias computation

serge-sans-paille:


serge-sans-paille:
Closing due to travis misconfig issue.

PR #919, opened 2018-05-24
Fix incorrect memory handling in variant functor

serge-sans-paille:
Thanks to Yann Diorcet for spotting the bug!


serge-sans-paille:
@diorcety is that ok to you?

serge-sans-paille:
@diorcety is that ok to you?

diorcety:
Yes it is!

PR #923, opened 2018-05-29
Add return_counts to np.unique

diorcety:


serge-sans-paille:
Extended by #938, that improve performances and perfom some clang-formating. thanks a lot o/

PR #934, opened 2018-06-07
Fix setup.py develop option

ashwinvis:


setup.py
```diff
@@ -81,6 +81,7 @@ def initialize_options(self):
     def copy_boost(self, src_only=False):
         "Install boost-simd and boost deps from the third_party directory"
 
+        # if hasattr(self, 'no_boost') and
```
serge-sans-paille: useless comment :-)

ashwinvis: Traces of debug. Removed and squashed.

PR #936, opened 2018-06-08
Prevent pythran from having a permanent side effect on CC and CXX

serge-sans-paille:
Solved by winning some king of monkey-patching race with numpy, cython
and setuptools...

### DIFF
--- a/pythran/config.py
+++ b/pythran/config.py
@@ -27,11 +27,6 @@ def init_cfg(sys_file, platform_file, user_file):
         cfgp.readfp(open(required))
     cfgp.read([user_config_path])
 
-    for key in ('CC', 'CXX'):
-        value = cfgp.get('compiler', key)
-        if value:
-            os.environ[key] = str(value)
-
     for obsolete_section in ('user', 'sys'):
         if cfgp.has_section(obsolete_section):
             logger.warn("Your pythranrc has an obsolete `%s' section",
@@ -41,13 +36,16 @@ def init_cfg(sys_file, platform_file, user_file):
 
 
 def make_extension(**extra):
+
     def parse_define(define):
         index = define.find('=')
         if index < 0:
             return (define, None)
         else:
             return define[:index], define[index + 1:]
+
     extension = {
+        "language": "c++",
         # forcing str conversion to handle Unicode case (the default on MS)
         "define_macros": [str(x) for x in
                           cfg.get('compiler', 'defines').split()],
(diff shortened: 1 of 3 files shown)


ashwinvis:
I don't know what to say! If we invert the order of imports will it work (see below)?

```python
from pythran.dist import PythranBuildExt
from Cython.Distutils.build_ext import build_ext
```
I hope we are not trading one side effect for another :smile:

serge-sans-paille:
> I don't know what to say! If we invert the order of imports will it work (see
> below)?

build_ext is monkey-patched by pythran, so the user should not do that
kind of things :-)

serge-sans-paille:
but yeah, i'm not happy with that either...

pythran/dist.py
```diff
@@ -4,13 +4,48 @@
...
+
+                try:
+                    return LegacyBuildExt.build_extension(self, ext)
+                finally:
+                    for attr in prev.keys():
+                        getattr(self.compiler, attr)[0] = prev[attr]
+            else:
+                return LegacyBuildExt.build_extension(self, ext)
+
+    build_ext.build_ext = PythranBuildExt
```
ashwinvis: Instead of monkey-patching, leave it to the user to inherit the custom build_ext class?

pythran/dist.py
```diff
@@ -4,13 +4,48 @@
...
+                    prev[attr] = getattr(self.compiler, attr)[0]
+                    getattr(self.compiler, attr)[0] = ext.cxx
+
+                try:
+                    return LegacyBuildExt.build_extension(self, ext)
+                finally:
+                    for attr in prev.keys():
+                        getattr(self.compiler, attr)[0] = prev[attr]
+            else:
+                return LegacyBuildExt.build_extension(self, ext)
```
ashwinvis: Maybe use `super(PythranBuildExt, self).build_extension(self, ext)`  here and above?

pythran/config.py
```diff
@@ -99,8 +105,9 @@ def parse_define(define):
         sys.stdout = old_stdout
 
     # final macro normalization
-    extension["define_macros"] = [parse_define(dm) for dm in
-                                  extension["define_macros"]]
+    extension["define_macros"] = [
+        dm if isinstance(dm, tuple) else parse_define(dm)
+        for dm in extension["define_macros"]]
     return extension
```
ashwinvis: Function `compiler()` a few lines below still relies on `os.environ.get('CXX'...`

pythran/dist.py
```diff
@@ -20,6 +55,9 @@ class PythranExtension(Extension):
     They must be processable by pythran, of course.
 
     The compilation process ends up in a native Python module.
+
+    Unlike distutils.core.Extension it also accepts a cxx keyword argument to
+    force the compiler executable.
```
ashwinvis: This is a good feature. If only it were implemented upstream, we could have avoided all these hacks :)

pythran/config.py
```diff
@@ -72,13 +70,21 @@ def parse_define(define):
...
     # using / as separator as advised in the distutils doc
     extension["include_dirs"].append(here + '/pythran')
+
+    extra.pop('language', None)  # forced to c++ anyway
+    cxx = extra.pop('cxx', None)
+    if cxx is not None:
+        extension['cxx'] = cxx
+    elif cfg.get('compiler', 'CXX'):
+        extension['cxx'] = str(cfg.get('compiler', 'CXX'))
+
```
ashwinvis: It maybe useful to add some sort of warning to the user (`DeprecationWarning` ?), to use the custom build_ext class when `CXX` is configured.

ashwinvis:
Updated this branch: https://github.com/fluiddyn/pythran/commit/bcc45e05317b6d9775a4bc207461f4f8705c8093

serge-sans-paille-qb:
Obsoleted by #940

PR #941, opened 2018-06-18
Fix/pythran extension side effect

serge-sans-paille:


ashwinvis:
The following lines should be redundant now:

https://github.com/serge-sans-paille/pythran/blob/4dd3b9b45402288bc11242ebc1fd41e4a7995df8/omp/__init__.py#L9-L14

PR #943, opened 2018-06-21
Use customize_compiler

diorcety:
Disable detect_gmp for Windows (cf new_compiler("posix"))
Allow to use the defined compiler (CC=clang for example) for detecting gmp (gcc could me not here at all)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,7 @@ from setuptools.command.develop import develop
 
 from setuptools import setup
 from distutils import ccompiler
+from distutils.sysconfig import customize_compiler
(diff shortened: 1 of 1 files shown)


ashwinvis:
Using `customize_compiler` makes the setup sensitive to environment variables. I can't prove it, but it could pose problems while packaging pythran as a dependency.

serge-sans-paille:
@diorcety in order to prepare for python3, I'm going to remove gmp (and thus long) support. This probably make your PR obsolete?

serge-sans-paille:
GMp is no longer supported, closing

PR #944, opened 2018-06-21
Add --generic flag for pythran-config

diorcety:
Allow to have pythran flags (like from .pythranrc file) without python specific flags
This is not so clever implementation, I know ...


serge-sans-paille:
concerning the API, maybe we should make the default behave as you propose with ``--generic``, and have ``--python`` add the python flags?

diorcety:
As you want on this point

serge-sans-paille:
Replaced by #952

PR #946, opened 2018-06-21
Improve pythran dist

diorcety:
The current implementation run multiple times the transpilation (for example sdist then for build and for install)
This one do that in simpliest way

### DIFF
--- a/pythran/dist.py
+++ b/pythran/dist.py
@@ -83,52 +83,38 @@ class PythranExtension(Extension):
     They must be processable by pythran, of course.
 
     The compilation process ends up in a native Python module.
-
-    Unlike distutils.core.Extension it also accepts a cxx keyword argument to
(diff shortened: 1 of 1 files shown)


diorcety:
Okay the Travis failures don't seem related to my PR

pythran/dist.py
```diff
@@ -9,13 +9,13 @@
 import os.path
 import os
 
-from distutils.command.build_ext import build_ext as LegacyBuildExt
+from setuptools.command.build_ext import build_ext as LegacyBuildExt
```
serge-sans-paille: this adds a dependency on setuptools... Is that required ?

serge-sans-paille:
Thans @diorcety for the PR!

serge-sans-paille:
Thanks @diorcety for the PR!

PR #949, opened 2018-06-30
Feature/partially static shape

serge-sans-paille:


serge-sans-paille:
@diorcety , @jeanlaroche : can you check the impact of this commit on your code base, in terms of ndarray change ?

jeanlaroche:
Sure let me check.

jeanlaroche:
mmm, one of my functions no longer compiles on the mac. It does on the latest from master but not on feature/partially-static-shape 
I will send you the file.

PR #952, opened 2018-07-07
Fix/pythran config nopython mode

serge-sans-paille:


serge-sans-paille:
Supersedes #944

serge-sans-paille:
@diorcety : please tell me if you're ok with the change!

diorcety:
:+1:

PR #957, opened 2018-07-13
numpy_texpr_2::fast use std::get<0> instead of [0]

diorcety:


serge-sans-paille:
LGTM, you need to rerun the appveyor build that failed to DL some package.

Issue #956, opened 2018-07-13
CXX from pythranrc


diorcety:
When compiling directly using pythran command line doesn't use CXX and CC from pythranrc file

serge-sans-paille:
Thanks @ashwinvis for the feedback. Looks like you're not using PythranBuildExt in  your setup.py, so the CXX override behavior is not activated. Funny enough, if I recall correctly, you were the one to advocate for that feature :-/

Am I missing something?

serge-sans-paille:
Adding ``cmdclass={"build_ext": PythranBuildExt},`` to your setup.py instead fixes the issue.

ashwinvis:
I am certainly missing something... Need some more time to track it down. My reply above was when I used PythranBuildExt.

I guess the catch i Pythran extensions arr built after C++ Cython extensions in fluidfft.

Sent from ProtonMail mobile

-------- Original Message --------

serge-sans-paille:
And anyway, we should not *ignore* CXX if ``PythranBuildExt`` is not used, I'll check that too.

ashwinvis:
No you are right.... It was my mistake. I was subclassing PythranBuildExt like:

```python
        from pythran.dist import PythranBuildExt

        class fluidfft_build_ext(build_ext, PythranBuildExt):
               pass
```

Certainly didn't work as intended. Worked when I tried `PythranBuildExt` explicitly.

serge-sans-paille:
And CXX is not ignored, I just forgot to set CC too, merging in that case

serge-sans-paille:
Fixed by #958

PR #958, opened 2018-07-13
Fix compilation setp: honor CXX from pythranrc

serge-sans-paille:
Use Pythran BuildExt as *the* standard way to build things, even
internally.


serge-sans-paille:
Fixes #956. @diorcety, @ashwinvis can you confirm it's okay with you, + @jeanlaroche on OSX

serge-sans-paille:
LGTM, waiting for some feedback before merging.

PR #964, opened 2018-07-17
Fix numpy_iexpr operator[]

diorcety:


diorcety:
I'm not sure about that. But there is an issue if the ndarray shape size is only 1

diorcety:
Yes

import numpy as np

#pythran export aaa(int32[][], int32[][])
def aaa(mat, mask):
    assert mask.shape == mat.shape
    mat[mask == False] = np.nan
    return mat

serge-sans-paille:
Thanks, I can reproduce the issue and will provide a correct (^^!) fix soon.

serge-sans-paille:
Obsoleted by #967

Issue #961, opened 2018-07-16
Performance issue: nearly twice slower than gfortran


paugier:
Serge, you wrote that you are happy with performance issues :relaxed: It is not a big issue, but still... Fortran is faster here... And nearly twice faster is a little bit too much, isn't it?

I think these two functions are approximately equivalent:
```python
%%pythran -O3 -march=native
#pythran export func_pythran(int):

def func_pythran(d):
    result = 0.
    for n in range(1, d):
        result += (1. + 1./n)**n
    return result
```

```fortran
%%fortran
double precision function func_fortran(d)
    integer :: n
    double precision :: result = 0.
    do n = 1, d
        result = result + (1. + 1./n)**n
    end do
    func_fortran = result
end function func_fortran
```
However,  it is nearly twice faster with Fortran:

```
d = 2**17

func_py(d)
356272.38805723446

func_fortran(d)
356230.0819711685

func_pythran(d)
356272.38805723446

%timeit func_py(d)
27.7 ms ± 663 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit func_fortran(d)
6.16 ms ± 5.56 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit func_pythran(d)
11.7 ms ± 2.79 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

- Note the numerical difference between (python, pythran) and fortran.

- The ratio `(time pythran)/(time fortran)` does not vary too much with the value of `d`.

serge-sans-paille:
Thanks @paugier, I should be carefull when speaking my mind out loud with you :-)
I'll have a look at this issue, and let the assembly speak!

serge-sans-paille:
After first inspection: the fortran code ends up with an unrolled loop calling https://github.com/llvm-mirror/compiler-rt/blob/master/lib/builtins/powisf2.c

The Pythran code ends up with an unrolled loop calling boost::simd::pow. So that's likely to be a pow specialization issue, I'll check that.

serge-sans-paille:
Confirmed, specializing the pow operator leads to better speedup!

serge-sans-paille:
(side note: unlike Python's range, Fortran do loop iterate until the upper bound, included, so the two codes are not exactly the same).

PR #966, opened 2018-07-17
Fix performance issue with np.power when second argument is an integer

serge-sans-paille:
Using static dispatch to exponentiation by squaring solves the issue.


serge-sans-paille:
Mostly fixes #961, fortran is still a bit faster because gfortran chooses to unroll.

PR #974, opened 2018-07-30
Better distutils tests support

diorcety:
On some system, distutils deny to install on a prefix which is not inside PYTHONPATH
Updating the PYTHONPATH environment variable according to prefix value and creating python library directories allows to make it works in all cases.

### DIFF
--- a/pythran/tests/test_distutils.py
+++ b/pythran/tests/test_distutils.py
@@ -1,5 +1,7 @@
 from subprocess import check_call
+from distutils.sysconfig import get_python_lib
 import os
+import errno
 import shutil
 import sys
 import sysconfig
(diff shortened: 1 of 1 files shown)


pythran/tests/test_distutils.py
```diff
@@ -14,37 +16,66 @@
 else:
     so_version = ""
 
+def merge(x, y):
+    z = x.copy()
+    z.update(y)
+    return z
+
+def mkdir_p(path):
```
serge-sans-paille: I'd rather have a version test that uses os.makedirs under python3 if possible

pythran/tests/test_distutils.py
```diff
@@ -14,37 +16,66 @@
...
+
+
+def my_check_call(args, **kwargs):
+    if 'prefix' in kwargs:
+        prefix = kwargs['prefix']
+        pythonpath = get_python_lib(prefix=prefix)
+        del kwargs['prefix']
+        args += ('--prefix=%s' % prefix, )
+        env = kwargs.get('env', os.environ)
+        env = merge(env, {'PYTHONPATH': pythonpath})
```
serge-sans-paille: as you're merging inpalce, why not just ``env.update``?

pythran/tests/test_distutils.py
```diff
@@ -14,37 +16,66 @@
...
+def mkdir_p(path):
+    try:
+        os.makedirs(path)
+    except OSError as exc:  # Python >2.5
+        if exc.errno == errno.EEXIST and os.path.isdir(path):
+            pass
+        else:
+            raise
+
+
```
serge-sans-paille: please add a comment that explain why you're doing that. Probably take it from the commit message

PR #979, opened 2018-08-08
Fix dist/compilation with MSVC

diorcety:


serge-sans-paille:
Obsoleted by #997 (same commit rebase + minor fix)

PR #980, opened 2018-08-08
Remote/diorcety/past xrange range

serge-sans-paille:


serge-sans-paille:
@diorcety I reworked this pone a lot, I hope it's still ok with you.

PR #981, opened 2018-08-09
Add missing rint declaration

diorcety:


serge-sans-paille:
Missing test case :-) Otherwise that's OK

PR #986, opened 2018-08-10
Improve type conversions from str

### LINKED ISSUE #985: Float conversion issue
This following example fails with pythran

hh.py:
```
import numpy

#pythran export aa(str):
def aa(x):
    return float(x)
```

ii.py:
```
import numpy
import hh

x = hh.aa("0.000172")
print(x)
assert x == 0.000172
```

diorcety:
Fix #985

### DIFF
--- a/pythran/pythonic/__builtin__/bool_.hpp
+++ b/pythran/pythonic/__builtin__/bool_.hpp
@@ -15,7 +15,7 @@ namespace __builtin__
     template <class T>
     bool bool_::operator()(T &&val) const
     {
-      return (bool)val;
+      return static_cast<bool>(val);
     }
 
     bool bool_::operator()() const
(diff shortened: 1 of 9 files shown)


serge-sans-paille:
@diorcety validation fails, I let you insypect that

diorcety:
I doesn't understand the test, because if you add this test. This test fails on master, not with my PR
```
    def test_generator_handler_name2(self):
        code = '''
            def foo(x):
                if x > 2:
                    for i in range(x):
                        if i > 1:
                            break
                        yield i
                else:
                    for i in ["1"] * x:
                        if len(i) > 1:
                            break
                        yield i

            def test_generator_handler_name2(n):
                return list(foo(n))'''
        self.run_test(code,
                      2,
                      test_generator_handler_name2=[int])
```

serge-sans-paille:
@diorcety I'm ok with that PR, just missing the test update we discussed about.

diorcety:
https://github.com/serge-sans-paille/pythran/pull/986/commits/7d65789312a9cb5abb0c5514cb675d077e4f876d#diff-d3c9d2d871f95edff366faf729fb7f3eR255

PR #988, opened 2018-08-10
Fix is_none with types::none<T, True>

### LINKED ISSUE #987: Is None issue
The following code fails with pythran
oo.py
```
#pythran export aa(str):
def aa(x):
    h = None
    if x == "ABCD":
        h = 23
    if h is not None:
        return 2
    return 1
~                     
```
jj.py
```
import oo

x = oo.aa("EEEE")
print(x)
assert x == 1
```

diorcety:
Fix #987


diorcety:
Static if x is None is broken

diorcety:
This is quiet ugly PR, but It shows a solution for all added tests

diorcety:
We need to propagate "staticness" of variables in order to check if a call is static

PR #989, opened 2018-08-15
T cannot convert to bool implicitly

luyi0619:
If `T` is `std::tuple<long, long>` or other class, `T` is not able to convert to `bool`.

### DIFF
--- a/pythran/pythonic/types/NoneType.hpp
+++ b/pythran/pythonic/types/NoneType.hpp
@@ -55,7 +55,7 @@ namespace types
   template <class T>
   none<T, false>::operator bool() const
   {
-    return !is_none && static_cast<const T &>(*this);
+    return !is_none;
   }
 
   template <class T>


serge-sans-paille:
Maybe the good patch is to use the ``__builtin__.bool`` operator then, and fix it if needed.

luyi0619:
I don't think I fully understand your point. Do we need to test if T can convert to bool before using that operator? Are you suggesting we will use std::is_convertible to do that check?

serge-sans-paille:
The original intent of the implicit bool cast was to say « this type is false if it's none or if the value if false »
Unfortunately, there is no bool conversion operator for tuple, but this should already part of https://github.com/serge-sans-paille/pythran/blob/master/pythran/pythonic/include/__builtin__/bool_.hpp and if not, it's a bug easily fixed. So my take on this is: explicitly call ``__builtin__::functor::bool_{}`` instead of the implicit conversion. I'm not sure if I'm clear?

serge-sans-paille:
#993 implement the correct fix, @luyi0619 does that look okay to you?

Issue #994, opened 2018-08-26
Scipy.special backend for Bessel functions


Chronum94:
How difficult would it be to implement this? They're very commonly used in a lot of physics problems in spherical coordinates, and Boost has them, so they can be used from there. It would be really nice to have these functions supported.

On a related note, fantastic package. Much appreciated.

serge-sans-paille:
That sounds feasible. Can you list the exact name of the functions you need?

Chronum94:
From https://docs.scipy.org/doc/scipy-0.19.1/reference/special.html

`jv`, `yv`, `kv`, `iv`, and their derivatives, which have are `jvp`, `yvp`, etc. Optionally, `hankel1`, `hankel2` would also be great to have, and also exist in Boost special functions.

serge-sans-paille:
ok! Will have a look at it :-)

serge-sans-paille:
Fixed by #1005 !

PR #1005, opened 2018-09-02
Add support for several scipy.special functions

serge-sans-paille:
Namely:

- hankel1
- hankel2
- iv
- ivp
- jv
- jvp
- kv
- kvp
- yv
- yvp

For the record, this relies on boost.math (already vendored as part of
pythran) but I considered using the cephes library, to match scipy's
choice. It's unfortunately written in goold old'C, with reference to
Python/Numpy stuff, which does not easily matches pythran design.

C++17 also has some implementation of the bessel functions, but that's
too far ahead for Pythran :-)


serge-sans-paille:
Fixes #994

Chronum94:
This looks absolutely fantastic. I'll try this out soon. Thanks so much.

PR #1008, opened 2018-09-06
Feature/xsimd

serge-sans-paille:


wolfv:
hmm there seems to be a problem with the int16 `min` thing! good that these tests found it + sorry :)

serge-sans-paille:
ok, there seems to be an issue with copysign under windows :-) cc @wolfv

serge-sans-paille:
yeah I did

wolfv:
There seems to be a problem with `_mm256_xor_si256` which doesn't exist for `AVX`, only for `AVX2`. I'll take care of that.

serge-sans-paille:
@wolfv I'll merge that one if it validates. Thanks a lot for the xsimd support!

wolfv:
awesome, really happy to see this land in pythran!

PR #1018, opened 2018-09-18
First try for shapes

aguinet:
(WIP!!)
I am not happy with multiple things:

* ``IntegralConstant`` is, as in C++, a type which contains the underlying python type + the constant. I did that with the existing metaclass system, and I can't manage to have ``__repr`` being implemented (so that the Python backend prints it nicely without any modification to it).
* I'm not sure the introduction of this ``IntegralConstant`` node is the best thing to do. It looks like a big hack to me!

Anyway, it works like that and generate the good ``pshape`` object!


serge-sans-paille:
Quick feedback: This is relatively intrusive. Especially the fact that you extend the ast, which is generally *not a good thing (how would the other analyse/passes behave?) I think that in List to tuple, you should *just* transform the tuples in a call to a new builtin, say __builtin__.pythran.pshape. And provide an analyse that marks Num inside such builtin as « Immediate ». then in the backend, you only need to pretty print ast.Num differently depending on the result of the Immediate Analysis. 

Still, you approch makes sense, as tehre are several points in Pythran where we should only accept immediate, but It's too intrusive as of now. The analyse seems easier to setup.

aguinet:
Extending ListToTuple does not feel like a good idea, as it is meant to only process lists, whereas we want to process tuples. Putting a pass after this one that processes tuple seems cleaner to me (as it separate both problems that are orthogonal).

About the analysis, I can do it like that. Like you say, I think it can be nice to represent immediate within the AST, but let's try this after!

aguinet:
I pushed a new version :)

aguinet:
This is a first stable version. This validates for python2 on my computer, waiting for Travis for all the other configurations!

serge-sans-paille:
The only failures are related to clang format (needs to pass clang-format-3.6 on the source code) and a remaining xrange that should be a range (cannot see where it comes from your diff though).

pythran/tables.py
```diff
@@ -2577,6 +2577,7 @@ def partialsum(seq):
             "static_if_": ConstFunctionIntr(),
             "StaticIfReturn": ConstFunctionIntr(),
             "StaticIfNoReturn": ConstFunctionIntr(),
+            "make_shape": ConstFunctionIntr()
```
serge-sans-paille: respect alphabetical order :-)

pythran/utils.py
```diff
@@ -78,6 +78,16 @@ def get_variable(assignable):
...
+def pythran_builtin(name):
+    return MODULES['__builtin__']['pythran'][name]
+
+def pythran_builtin_path(name):
+    assert name in MODULES['__builtin__']['pythran']
+    return ('__builtin__', 'pythran', name)
+
+def pythran_builtin_attr(name):
+    return path_to_attr(pythran_builtin_path(name))
+
```
serge-sans-paille: I like that.

pythran/analyses/immediates.py
```diff
@@ -0,0 +1,21 @@
...
+from pythran.passmanager import NodeAnalysis
+from pythran.utils import pythran_builtin, attr_to_path
+from pythran.tables import MODULES
+
+import gast as ast
+
+class Immediates(NodeAnalysis):
+    def __init__(self):
+        self.result = set()
+        self.make_shape = pythran_builtin('make_shape')
```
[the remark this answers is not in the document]
aguinet: Done!

pythran/pythonic/include/types/tuple.hpp
```diff
@@ -152,7 +152,7 @@ namespace types
   }
 
   template <class T>
-  struct is_pshape_element : std::false_type {
+  struct is_pshape_element : std::is_convertible<T, long> {
```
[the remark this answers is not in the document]
aguinet: Okay for is_integral!

pythran/tables.py
aguinet: Done!

pythran/pythonic/include/__builtin__/pythran/make_shape.hpp
```diff
@@ -0,0 +1,20 @@
+#ifndef PYTHONIC_INCLUDE_BUILTIN_MAKE_SHAPE_HPP
+#define PYTHONIC_INCLUDE_BUILTIN_MAKE_SHAPE_HPP
+
+PYTHONIC_NS_BEGIN
+
+namespace __builtin__ {
+namespace pythran {
+  template <class... Args>
+  auto make_shape(Args ... args) -> pythonic::types::pshape<Args...>
+  {
```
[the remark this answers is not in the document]
aguinet: Okay!

pythran/backend.py
```diff
@@ -901,16 +902,20 @@ def visit_Call(self, node):
...
+            ret = 'pythonic::numpy::nan'
         elif isinf(node.n):
-            return ('+' if node.n >= 0 else '-') + 'pythonic::numpy::inf'
+            ret = ('+' if node.n >= 0 else '-') + 'pythonic::numpy::inf'
         else:
-            return repr(node.n) + TYPE_TO_SUFFIX.get(type(node.n), "")
+            ret = repr(node.n) + TYPE_TO_SUFFIX.get(type(node.n), "")
+        if node in self.immediates:
+            assert isinstance(node.n, int)
+            return "std::integral_constant<%s, %s>{}" % (pytype_to_ctype(type(node.n)), ret)
```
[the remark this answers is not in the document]
aguinet: I tried something here https://github.com/serge-sans-paille/pythran/pull/1018/commits/9d3397669c01e90b8f78d422c9147e9494b01d11 . I'm really not sure that's what you were thinking.

Two questions here:

* wouldn't it be better to put the whole "``Num``" literal in ``cxxgen``?
* as far as I understand, ``visit_Num`` is expected to return a string (for instance if a call uses a ``Num`` node). If we return a type for ``cxxgen``, shouldn't we change a lot of other things in the backend?

Thanks!

pythran/analyses/immediates.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Immediates(NodeAnalysis):
+    def __init__(self):
+        self.result = set()
+        super(Immediates, self).__init__(Aliases)
+
+    def visit_Call(self, node):
+        func_aliases = self.aliases[node.func]
+        for fa in func_aliases:
+            if fa is _make_shape:
```
serge-sans-paille: this is unfortunately incorrect: what if (very unlikely) this func alias to make_shape *and* something else? the correct check would be ``if len(func_aliases) == 1 and next(iter(func_aliases)) is _make_shape``

pythran/cxxgen.py
```diff
@@ -174,6 +173,17 @@ def get_tp():
...
 
+# literal values
+
+class IntegralConstant(object):
+    """Declare an integral constant known at compile time. Type and value are strings"""
+    def __init__(self, typename, value):
+        self.typename = typename
+        self.value = value
+
+    def __str__(self, with_semicolon=False):
```
serge-sans-paille: It's an expression, so it's probably useless to add this ``with_semicolon`` arg

pythran/optimizations/tuple_to_shape.py
```diff
@@ -0,0 +1,48 @@
...
+        import numpy
+        return numpy.ones(__builtin__.pythran.make_shape(n, 4))
+    """
+    def __init__(self):
+        self.update = False
+        super(TupleToShape, self).__init__(Aliases)
+
+    def visit_Call(self, node):
+        func_aliases = self.aliases.get(node.func, None)
+        if not func_aliases is None and func_aliases.issubset(patterns):
```
serge-sans-paille: same here: you need to ensure that *all* aliases match one of the pattern:

```
if init == 1:
    f = ones
else:
    f = empty
g = f((10, 3))
```

pythran/pythonic/__builtin__/pythran/make_shape.hpp
```diff
@@ -0,0 +1,19 @@
...
+PYTHONIC_NS_BEGIN
+
+namespace __builtin__
+{
+  namespace pythran
+  {
+    template <class... Args>
+    auto make_shape(Args... args) -> pythonic::types::pshape<Args...>
+    {
+      return pythonic::types::pshape<Args...>{args...};
```
serge-sans-paille: could be: `` return {args...};``

pythran/pythonic/include/__builtin__/pythran/make_shape.hpp
```diff
@@ -0,0 +1,18 @@
...
+#define PYTHONIC_INCLUDE_BUILTIN_MAKE_SHAPE_HPP
+
+PYTHONIC_NS_BEGIN
+
+namespace __builtin__
+{
+  namespace pythran
+  {
+    template <class... Args>
+    auto make_shape(Args... args) -> pythonic::types::pshape<Args...>;
```
serge-sans-paille: useless use of auto btw :-)

pythran/pythonic/include/types/tuple.hpp
```diff
@@ -152,10 +152,7 @@ namespace types
   }
 
   template <class T>
-  struct is_pshape_element : std::false_type {
-  };
-  template <>
-  struct is_pshape_element<long> : std::true_type {
+  struct is_pshape_element : std::is_integral<T> {
```
serge-sans-paille: so cool

pythran/utils.py
```diff
@@ -78,6 +78,16 @@ def get_variable(assignable):
...
 
+def pythran_builtin(name):
+    return MODULES['__builtin__']['pythran'][name]
+
+def pythran_builtin_path(name):
+    assert name in MODULES['__builtin__']['pythran']
+    return ('__builtin__', 'pythran', name)
+
+def pythran_builtin_attr(name):
+    return path_to_attr(pythran_builtin_path(name))
```
serge-sans-paille: (for another PR) this would be cool to use these everywhere

pythran/cxxgen.py
aguinet: Yep, and it's also very useless because it's an argument to ``__str__``...

pythran/analyses/immediates.py
aguinet: Okay gotit, fixed!

pythran/optimizations/tuple_to_shape.py
aguinet: Yep, fixing it!

pythran/utils.py
aguinet: Yep :)

pythran/pythonic/__builtin__/pythran/make_shape.hpp
aguinet: Fixed!

pythran/pythonic/include/__builtin__/pythran/make_shape.hpp
aguinet: Fixed :)

PR #1019, opened 2018-09-18
Fix slicing of array

serge-sans-paille:
missing gexpr definition, reorder include mess to avoid including
ndarray as a side effect.


serge-sans-paille:
@jeanlaroche can you check this still fixes your issue? I made a few changes.

PR #1022, opened 2018-09-20
Fix OMP on Windows

diorcety:
Without this modification, the following issue apprears

`pythran\pythonic/utils/broadcast_copy.hpp(71): error C3016: 'i': index variable in OpenMP 'for' statement must have signed integral type`


serge-sans-paille:
This breaks the validation :-)

diorcety:
GCC issue?

serge-sans-paille:
I can reproduce with gcc higher version, let me check that first.

serge-sans-paille:
I can reproduce with gcc higher version, let me check that first.oh, and I cna even reproduce on master without your patch :-/

serge-sans-paille:
Yipee!

PR #1034, opened 2018-10-05
Fix/gitignore pythran xsimd

paugier:
I just added pythran/xsimd to gitignore.

I don't understand why there is this change to pythran/pythonic/include/types/vectorizable_type.hpp ...


serge-sans-paille:
You're not based on the correct branch, you should be based on master instead. Can you change that? Or just git rebase -i and remove 6d9a36 from your branch.

PR #1041, opened 2018-10-07
Fix combined of texpr and expr

### LINKED ISSUE #1039: Compilation error when conditionally returning an input or a function of it (Linux, MacOS)
I'm getting a compilation error on this:
The error seems to indicate that the input variable W is understood to be a list, but I'm not sure why that would be... 
```
import numpy as np

#pythran export test(float[][])
def test(W):
    S = func1()
    if S == 0:
        return W            # <- Fails to compile.
        #return np.array(W)  # <- Compiles
    return 2*W

def func1():
    return 1

```

serge-sans-paille:
This fixes #1039


serge-sans-paille:
@jeanlaroche is that ok with you?

PR #1067, opened 2018-10-29
Remove yaourt and unmaintained python2-pythran-git recommendation

ashwinvis:
`yaourt` is not recommended and frowned upon among Arch Linux people, as it fail to [satisfy many criterias](https://wiki.archlinux.org/index.php/AUR_helpers#Pacman_wrappers) :) . I will maintain the `python-pythran` and `python2-pythran` packages, for which the PKGBUILD is derived from the old PKGBUILD by @xantares. You can see the what the package does below:

https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=python-pythran

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,9 +70,9 @@ Depending on your setup, you may need to add the following to your \\~/.pythranr
 ArchLinux
 =========
 
-Using `yaourt`::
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
Great! Thanks

PR #1073, opened 2018-11-01
Use faster scalar version for trigo function

serge-sans-paille:
Second try.


JohanMabille:
If it turns out to be faster than the scalar (std) version, we can implement the same algorithms for scalars as the ones operating on batches (so we save the transfer to and from simd registers).

serge-sans-paille:
the speedup may come from the usage of sse instruction though. This needs more investigation. But yes, this is faster for the flagged functions, at least on my laptop, but this is probably hardware-dependent.

PR #1076, opened 2018-11-03
Missing xor-update operator for array-like

serge-sans-paille:


serge-sans-paille:
@nbecker this is needed to compile https://gist.github.com/nbecker/e60920fc54ceb443437733631cc402f6 

Can't do anything for GCC right now.

Issue #1085, opened 2018-11-12
Can't init global variable using imported function


diorcety:
If you use cc() instead of aa() inside XX variable, the compilation works

aa.py
```
from .bb import aa

def cc():
    return aa()


XX = [aa(), 3]

#pythran export bb()
def bb():
    return XX
```

bb.py
```
def aa():
    return 3.14
```

serge-sans-paille:
I can reproduce that one... that's a bug :-)

Issue #1084, opened 2018-11-12
Can't use imported variable inside relative imported module


diorcety:
If you comment the first implementation of aa inside bb.py and use the second, the compilation works

aa.py
```
from .bb import aa

def cc():
    return aa()


XX = [cc(), 3]

#pythran export bb()
def bb():
    return XX
```

bb.py
```

from numpy import pi

def aa():
    return pi
"""

def aa():
    return 3.14
"""

```

Issue #1082, opened 2018-11-10
Relative import and global constants issue


diorcety:
Pythran doesn't support relative constant (or variable) import

c.py:
`A = 2`

d.py:
```
from .cc import A

#pythran export aa(int)
def aa(a):
    return a + A
```

PR #1090, opened 2018-11-13
Feature/rework handle import

serge-sans-paille:


serge-sans-paille:
@diorcety: this should solve your recent issues, can you please try this branch?

diorcety:
Indeed it works. But still one issue if you replace the definition in #1084 by

```
def aa():
    raise ValueError("ee")
```

diorcety:
```
CRITICAL: I am in trouble. Your input file does not seem to match Pythran's constraints...
E: identifier __pythran_imported__bb$ValueError unknown, either because it is an unsupported intrinsic, the input code is faulty, or... pythran is buggy. (bb.py, line 4)
```

pythran/transformations/handle_import.py
```diff
@@ -29,419 +34,183 @@ def is_builtin_function(func_name):
...
-
-        # Function is not defined locally, maybe it is an alias, like
-        # c = math.cos
-        # c()
-        # Just give up here and hope for the best!
-        if func_name not in self.functions:
-            return None
-
-        func = self.functions[func_name]
+                node.id = self.prefixes[-1] + node.id
```
diorcety: Should be not is_builtin_function(node.id): instead of else?

pythran/transformations/handle_import.py
```diff
@@ -29,419 +34,183 @@ def is_builtin_function(func_name):
...
 
-        return func.name
+        renaming = self.lookup(root.id)
 
+        if not renaming:  # happens for builtins
+            return node
 
-class BuiltinModule(object):
+        if not renaming.endswith('$'):
+            return node
```
diorcety: node.id = renaming before return node?

diorcety:
Here the patch:
https://gist.github.com/diorcety/8f2d94fcab93aee5d3d72048774ec51c

serge-sans-paille:
I need test cases!

diorcety:
Replace aa by
```
def aa(a):
    ret = []
    while a > 2:
        ret.append(0)
        a -=1 
    return ret
```

diorcety:
https://gist.github.com/diorcety/970f5d1502258c08919f6407b9f99748

serge-sans-paille:
@diorcety better now?

diorcety:
Better, but not perfect :)
https://github.com/diorcety/pythran/tree/rework-handle-import

serge-sans-paille:
@diorcety go?

serge-sans-paille:
(this does not fix your performance issue though)

serge-sans-paille:
This fixes #1085, #1084 and #1082!

diorcety:
Seems good 👍

PR #1106, opened 2018-11-21
Fix remainder implementation

serge-sans-paille:
numpy.remainder is different from std::remainder


pythran/tests/test_numpy_func1.py
```diff
@@ -295,6 +295,12 @@ def test_isclose7(self):
                       numpy.array([[-1.01, 1e-10+1e-11, -0, 7.],[9., 1e-10, 0., float('NaN')]]),
                       np_isclose7=[NDArray[float,:], NDArray[float,:,:]])
 
+    def test_remainder0(self):
```
jeanlaroche: You can't test with integer numbers, this will always work!
You have to test with floats that have a fractional part.

jeanlaroche: Never mind, I'm an idiot!

Issue #1111, opened 2018-11-26
Pythran much slower than Numpy to generate random numbers


paugier:
### Pythran slower than Numpy...

We have a big performance issue with random number generation:

```
# tmp_rand.py
import numpy as np
# pythran export rand2d(int)
def rand2d(n):
    return np.random.rand(n, n)
```
compiled with `pythran -march=native -DUSE_XSIMD tmp_rand.py -o tmp_rand_pythran.so`

With Numpy:
```
In [1]: from tmp_rand import rand2d as rand2d_py                                  
In [2]: for n in [100, 1000, 2000, 4000, 10000]: 
   ...:     %timeit rand2d_py(n) 
   ...:                                                                           
107 µs ± 32.5 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
10.5 ms ± 3.31 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
42.1 ms ± 6.91 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
217 ms ± 204 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.35 s ± 1.86 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

With Pythran:
```
In [1]: from tmp_rand_pythran import rand2d                                       
In [2]: for n in [100, 1000, 2000, 4000, 10000]: 
   ...:     %timeit rand2d(n) 
   ...:                                                                           
942 µs ± 322 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
94.2 ms ± 68.4 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
378 ms ± 2.75 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.56 s ± 422 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
9.72 s ± 875 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Oups, Pythran is approximately 10 times slower than Numpy to generate random numbers...

Since many benchmarks involve random number generation, Pythran gets really bad results for these benchmarks because of this random generation issue!  I know that random number generation performance is not be so important in most codes, but communication matters...So silly micro-benchmarks matter :slightly_smiling_face: 

Note that it seems to be a regression since Numpy and Pythran performance were similar some months ago (#759 ).

### Numpy much slower than other solutions for random number generation

[...]

PR #1118, opened 2018-11-28
Feature/faster random engine

serge-sans-paille:


serge-sans-paille:
This should fix #1111

PR #1132, opened 2018-12-07
Make arange a lazy expression

serge-sans-paille:
This speedsup expression like ``np.arange(x) * y``.


serge-sans-paille:
Fixes #1114, and that should be a performance improvment for @jeanlaroche !

jeanlaroche:
Awesome, let me grab, and retest!

jeanlaroche:
I get a compilation error using the new branch but not with master.
Source code: https://pastebin.com/m89JeBzb

Error:
https://pastebin.com/LG0Dk2jD

I didn't do a python setup.py -install. Is that necessary here?

serge-sans-paille:
@jeanlaroche better now?

jeanlaroche:
WOn't be able to check until monday, but I trust you! :)

PR #1151, opened 2018-12-21
Correct support of assignment of a transposed view through a slice

### LINKED ISSUE #1148: Transposed assignement
There is an issue with the transposed assignement. It can easily seen in this example:

Results: https://imgur.com/a/pp1Goah

test.py
```
from matplotlib import pyplot as plt
import fct2
import numpy as np
import matplotlib as mpl

mpl.use("TkAgg")
data = fct2.draw_circle((500, 500), (130.0, 180.0))
print(data.shape)
print(data.dtype)
plt.imshow(data.T, interpolation='nearest')
plt.show()
```

fct2.py
```
import numpy as np

def get_mgrid(shape):
    #return np.mgrid[:shape[0], :shape[1]]
    xx = np.empty(shape, dtype=int)
    xx.T[:] = np.arange(0, shape[0], 1, dtype=int)
    yy = np.empty(shape, dtype=int)
    yy[:] = np.arange(0, shape[1], 1, dtype=int)
    return xx, yy


# pythran export draw_circle((int, int), (float, float))
def draw_circle(shape, center):
    xx, yy = get_mgrid(shape)
    circle = (xx - center[0]) ** 2 + (yy - center[1]) ** 2
    return circle

```

serge-sans-paille:
Fix #1148

### DIFF
--- a/pythran/pythonic/include/numpy/transpose.hpp
+++ b/pythran/pythonic/include/numpy/transpose.hpp
@@ -11,6 +11,10 @@ PYTHONIC_NS_BEGIN
 namespace numpy
 {
 
+  template <class T>
+  types::numpy_texpr<types::ndarray<T, types::pshape<long, long>>>
+  transpose(types::ndarray<T, types::array<long, 2>> const &arr);
+
   template <class T, class pS0, class pS1>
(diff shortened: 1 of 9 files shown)


serge-sans-paille:
@diorcety this is my end-of-year present ;-) Thanks a lot for all your issues and bugfixes!

PR #1161, opened 2019-01-08
Implement stack

jeanlaroche:
Added stack and test cases. Currently this only works for a list of ndarray input. Two of the added tests do not pass. 
Also I don't know/think that the code is optimal in terms of avoiding copy operations...


serge-sans-paille:
Obsoleted by #1164

pythran/pythonic/numpy/stack.hpp
```diff
@@ -29,16 +26,32 @@ namespace numpy
...
+    }
+
+    
+    template <class E, size_t M>
+    types::ndarray<typename E::dtype, types::array<long, E::value+1>>
+    stack(types::array<E, M> const &args, long axis)
+    {
+        return _stack<types::array<E, M>, typename E::dtype, E::value>(args,axis);
+    }
+
```
serge-sans-paille: I think I handled that case in my PR too :-)

jeanlaroche: ah ah ah of course of course. Took you 1/2 a second I'm sure! :)

PR #1165, opened 2019-01-10
Rework and improve expression broadcasting support

### LINKED ISSUE #1163: Compilation issue depending on instruction decomposition
Maybe there is two issue with this test:
- Following the decomposition the generated code is not the same (after optimizations, it should be?)
- That the code doesn't work in one case

```
import numpy as np
import numpy.linalg as npl

def _get_cross(a, b):
    #return np.cross(far - center, coords - center)
    #return a[..., 0] * b[..., 1] - a[..., 1] * b[..., 0]
    return a[0] * b[:, :, 1] - a[1] * b[:, :, 0]


def _get_mgrid(shape):
    #return np.mgrid[:shape[0], :shape[1]]
    xx = np.empty(shape, dtype=int)
    xx.T[:] = np.arange(0, shape[0], 1, dtype=int)
    yy = np.empty(shape, dtype=int)
    yy[:] = np.arange(0, shape[1], 1, dtype=int)
    return xx, yy

#pythran export bb((int, int), (float, float), float)
def bb(shape, center, angle):
    xx, yy = _get_mgrid(shape)
    coords = np.rollaxis(np.array([xx, yy]), 0, 3)
    far = center + np.array([np.cos(angle), np.sin(angle)]) * 100

    # Don't work
    #z = np.absolute(_get_cross(far - center, coords - center))
    #return z

    # Work
    y = _get_cross(far - center, coords - center)
    return np.absolute(y)

```

serge-sans-paille:
It's now possible to slice, index, etc expression involving broadcasting.

Fix #1163

### DIFF
--- a/pythran/pythonic/include/types/ndarray.hpp
+++ b/pythran/pythonic/include/types/ndarray.hpp
@@ -421,6 +421,14 @@ namespace types
     long size() const;
 
     /* extended slice indexing */
+    template <class Ty>
+    auto operator()(Ty s) const ->
+        typename std::enable_if<std::is_integral<Ty>::value,
+                                decltype((*this)[s])>::type
+    {
+      return (*this)[s];
+    }
+
     template <class S0, class... S>
     auto operator()(S0 const &s0, S const &... s) const & -> decltype(
         extended_slice<count_new_axis<S0, S...>::value>{}((*this), s0, s...));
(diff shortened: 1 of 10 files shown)


serge-sans-paille:
@diorcety can you confirm this one is okay for you?

diorcety:
@serge-sans-paille Almost.
Is it normal that doesn't work if you force center to be an array?
`center = np.asarray(center)`

serge-sans-paille:
@diorcety should be fixed by now!

diorcety:
👍

serge-sans-paille:
Hell yeah!

PR #1169, opened 2019-01-10
Allow reshaping of an extended slice

### LINKED ISSUE #1166: Issue with stack[::][::] Mac and Linux
I'm getting a compilation error on this:
```
import numpy as np

#pythran export test(float[::][::], float[::][::], int)
def test(a,b,axis):
    return np.stack((a,b), axis)
```
```
/home/UAUDIO/jlaroche/temp/pythran/pythran/pythonic/include/types/numpy_gexpr.hpp:611:7: error: static assertion failed: should never be instanciated
       static_assert(M == 0, "should never be instanciated");
```
Is this a bug?

serge-sans-paille:
This fixes #1166

### DIFF
--- a/pythran/pythonic/include/types/numpy_gexpr.hpp
+++ b/pythran/pythonic/include/types/numpy_gexpr.hpp
@@ -605,12 +605,6 @@ namespace types
       return _shape;
     }
 
-    template <size_t M>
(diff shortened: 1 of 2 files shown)


jeanlaroche:
I confirm this fixes it for me! A million thanks!

PR #1172, opened 2019-01-15
Implement interp

jeanlaroche:
Interp is implemented, along with tests.
I must be doing something wrong though, because it's far slower than numpy (on my mac).
For example:
```
from interp import *
import numpy as np
x = np.linspace(-1,2,200)
xp = np.sort(np.random.randn(1000))
fp = np.random.randn(1000)

% timeit np.interp(x,xp,fp)
% timeit interp(x,xp,fp)
```
returns this:
```
In [6]: % timeit np.interp(x,xp,fp)
100000 loops, best of 3: 3.71 µs per loop

In [7]: % timeit interp(x,xp,fp)
100000 loops, best of 3: 12.2 µs per loop
```
Could you verify that it's the same on your machine? I'm a bit puzzled as to why that's happening.


serge-sans-paille:
@jeanlaroche please rebase against master! Also can this wait after the release?

serge-sans-paille:
@jeanlaroche can you rebase against master?

jeanlaroche:
Should be done now...

jeanlaroche:
Serge, I have to clean this up as well, there should not be all the commits!

serge-sans-paille:
On Sat, Jan 26, 2019 at 09:48:21AM -0800, jean laroche wrote:
> Serge, I have to clean this up as well, there should not be all the commits!
sure, you can just push --force in your branch instead of deleting/ recreating PR

jeanlaroche:
> 
> On Sat, Jan 26, 2019 at 09:48:21AM -0800, jean laroche wrote: Serge, I have to clean this up as well, there should not be all the commits!
> sure, you can just push --force in your branch instead of deleting/ recreating PR

Yes, I found out about that. I think I closed this PR by mistake. But I think the new PR is good now. I'm not sure how my master got so messed up but it's now hard reset to you, so I'm all good.

PR #1179, opened 2019-01-19
No such module pythran.test

hroncok:
Not sure if this is even used, but `python setup.py test` failed with ModuleNotFoundError.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -181,6 +181,6 @@ setup(name='pythran',
       setup_requires=["pytest-runner"],
       tests_require=['pytest', 'flake8'],
       extras_require={'deps': ['numpy']},
-      test_suite="pythran/test",
+      test_suite="pythran/tests",
       cmdclass={'build_py': BuildWithThirdParty,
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
mmmh, I tried your patch and it does not seem to work. What is the complete expected sequence to run tests that way (yep, I'm obviously not using it but I'm willing to enable that if it makes sense to some people)?

hroncok:
it was just: `python setup.py test`

but I guess if you want to support that with pytest, you already support `python setup.py pytest`, so an alias in setup.cfg would do.

https://pypi.org/project/pytest-runner/

> Optionally, set `test=pytest` in the `[aliases]` section of `setup.cfg` to cause `python setup.py test` to invoke pytest.

serge-sans-paille:
Obsoleted by #1180

serge-sans-paille:
I took the occasion of cleaning the test suite a bit, thanks a lot @hroncok for motivating that!

PR #1188, opened 2019-01-24
Implement default params

jeanlaroche:
This isn't ready yet. I noticed that np.arange() breaks with this change because the C++ code expects a dtype as the last param instead of a None (the default).
I could go ahead and try to fix that but that might be a can of worms. Instead, I was wondering if you had an idea to disable the new code for functions that have complicated signatures like np.arange()


serge-sans-paille:
@jeanlaroche please open an issue for arange default, I should be able to handle that. There is also a problem in how you handle branch. there should be only a few commits in the PR, not +100

jeanlaroche:
Ah shit. I must be doing something wrong in the way I rebase against master.

PR #1192, opened 2019-01-26
Implement interp

jeanlaroche:
Re-opening this after cleanup of merge and so on.

### DIFF
--- a/pythran/pythonic/include/numpy/interp.hpp
+++ b/pythran/pythonic/include/numpy/interp.hpp
@@ -1,0 +1,50 @@
+#ifndef PYTHONIC_INCLUDE_NUMPY_INTERP_HPP
+#define PYTHONIC_INCLUDE_NUMPY_INTERP_HPP
+
+#include "pythonic/include/utils/functor.hpp"
+#include "pythonic/include/utils/numpy_conversion.hpp"
+#include "pythonic/include/types/ndarray.hpp"
+#include "pythonic/include/__builtin__/None.hpp"
+
+PYTHONIC_NS_BEGIN
+
+namespace numpy
+{
+
+  // None,None,None
+  template <class T1, class T2, class T3>
+  types::ndarray<double, types::pshape<long>>
+  interp(T1 x, T2 xp, T3 fp, types::none_type left = types::none_type{},
+         types::none_type right = types::none_type{},
+         types::none_type period = types::none_type{});
+
+  // left None None
+  template <class T1, class T2, class T3, typename t1>
+  types::ndarray<double, types::pshape<long>>
+  interp(T1 x, T2 xp, T3 fp, t1 left,
+         types::none_type right = types::none_type{},
+         types::none_type period = types::none_type{});
+
+  // None right None
+  template <class T1, class T2, class T3, typename t1>
+  types::ndarray<double, types::pshape<long>>
+  interp(T1 x, T2 xp, T3 fp, types::none_type left, t1 right,
+         types::none_type period = types::none_type{});
+  // None None period
+  template <class T1, class T2, class T3, typename t1>
+  types::ndarray<double, types::pshape<long>>
+  interp(T1 x, T2 xp, T3 fp, types::none_type left, types::none_type right,
+         t1 period);
+
+  // left right None
+  template <class T1, class T2, class T3, typename t1, typename t2>
+  types::ndarray<double, types::pshape<long>>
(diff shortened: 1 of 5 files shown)


pythran/tables.py
```diff
@@ -19,7 +19,7 @@
 from pythran import interval
 from functools import reduce
 
-if sys.version_info.major == 3:
+if sys.version_info.major == 3:
```
serge-sans-paille: useless diff

pythran/tests/test_numpy_func3.py
```diff
@@ -9,7 +9,7 @@
 
 
 @TestEnv.module
-class TestNumpyFunc3(TestEnv):
+class TestNumpyFunc3(TestEnv):
```
serge-sans-paille: same here

pythran/pythonic/include/numpy/interp.hpp
```diff
@@ -0,0 +1,50 @@
+#ifndef PYTHONIC_INCLUDE_NUMPY_INTERP_HPP
```
serge-sans-paille: as this code is inspired from the numpy code base, you should state the original license and source used to derive this code.

pythran/pythonic/numpy/interp.hpp
```diff
@@ -0,0 +1,126 @@
+#ifndef PYTHONIC_NUMPY_INTERP_HPP
+#define PYTHONIC_NUMPY_INTERP_HPP
```
serge-sans-paille: The header should actually be here.

pythran/pythonic/numpy/interp_core.cpp
```diff
@@ -0,0 +1,256 @@
+//
+//  interp_core.cpp
+//  From /Users/jlaroche/temp/NumpySrc/numpy/core/src/multiarray/compiled_base.c
```
serge-sans-paille: Well, the licencing should be reproduced here.

pythran/pythonic/numpy/interp_core.cpp
```diff
@@ -0,0 +1,256 @@
...
+//    else {
+//        rval = PyFloat_AsDouble(right);
+//        if (error_converting(rval)) {
+//            goto fail;
+//        }
+//    }
+
+// xp->dx   fp->dy  x -> dz
+
+#define PyArray_malloc malloc
```
serge-sans-paille: avoid that and use a std::vector that you resize appropriately, that makes memory management safer

pythran/pythonic/numpy/interp_core.cpp
```diff
@@ -0,0 +1,256 @@
...
+    }
+    //        NPY_END_THREADS;
+  } else {
+    npy_intp j = 0;
+
+    /* only pre-calculate slopes if there are relatively few of them. */
+    if (lenxp <= lenx) {
+      slopes = (npy_double *)PyArray_malloc((lenxp - 1) * sizeof(npy_double));
+      if (slopes == NULL) {
+        // goto fail;
```
serge-sans-paille: just remove the test, the resie will throw an exception if it fails.

pythran/pythonic/numpy/interp_core.cpp
jeanlaroche: There isn't any licensing in the original file.

serge-sans-paille:
> There isn't any licensing in the original file.

Then just copy the header of the numpy license, the one at the top of the project?

jeanlaroche:
Performance does not seem to be very good on this for a reason that eludes me.
```
import numpy as np
np.random.seed(0)
xi=np.random.randn(10000)
xi=np.sort(xi)
yi=np.random.randn(10000)
x=np.random.randn(100000)

%timeit np.interp(x,xi,yi)
import jean as tt
%timeit tt.test(x,xi,yi)
```
```
import numpy as np

#pythran export test(float[],float[],float[])
def test(x,xi,yi):
    return np.interp(x,xi,yi)
```

Results in linux:
```
100 loops, best of 3: 9.87 ms per loop <- numpy
100 loops, best of 3: 9.92 ms per loop <- pythran
```

Results on macos:
```
8.76 ms ± 122 µs per loop (mean ± std. dev. of 7 runs, 100 loops each) <- numpy
9.83 ms ± 126 µs per loop (mean ± std. dev. of 7 runs, 100 loops each) <- pythran
```
Adding -march=native or -fopenmp does not change anything.

Issue #1189, opened 2019-01-25
Transpose ignored in recent Pythran


Dapid:
```
import random
import numpy as np

#pythran export call()
def call():
    return np.random.random((3, 5)).T
```
```
import test

out = test.call()
print(out.shape)
```

On Pythran master and released, I get (3, 5), instead of (5, 3). Pythran 0.9.0 works fine.

Dapid:
Git bisect points to the culprit:

```
8b4f22aa0ef0a33308d87b2dcb7d9b4cef3a6187 is the first bad commit
commit 8b4f22aa0ef0a33308d87b2dcb7d9b4cef3a6187
Author: serge-sans-paille <serge.guelton@telecom-bretagne.eu>
Date:   Fri Dec 21 18:47:26 2018 +0100

    Correct support of assignment of a transposed view through a slice
    
    Fix #1148
```

https://github.com/serge-sans-paille/pythran/commit/8b4f22aa0ef0a33308d87b2dcb7d9b4cef3a6187

serge-sans-paille:
Probably related to #1183. This is a critical bug: I'll fix that and release a patch version of pythran.

Dapid:
Thanks!

I think it would be nice to have some nightly integration test testing applications using Pythran. For example, a Travis nightly run. Hopefully that will catch more bugs.

Dapid:
I am now running a weekly test with Pythran master on my code:

https://travis-ci.org/ElofssonLab/pyGaussDCA/

Issue #1183, opened 2019-01-22
Invalid double transpose behavior


serge-sans-paille:
reproducer:
```python
#pythran export next_pythagorean_triples(int64[:,:])
import numpy as np
def next_pythagorean_triples(previous):
    matrices = np.array(
        [[-1, 2, 2],
         [-2, 1, 2],
         [-2, 2, 3],
         [1, 2, 2],
         [2, 1, 2],
         [2, 2, 3],
         [1, -2, 2],
         [2, -1, 2],
         [2, -2, 3]])

    print(previous.shape)
    next_triples = np.transpose(matrices @ np.transpose(previous))
    print(next_triples.shape)
    next_triples = next_triples.reshape((3 * previous.shape[0], previous.shape[1]))
    print(next_triples.shape)
    return next_triples
```

```python
import numpy as np
from pt import next_pythagorean_triples
current = np.array([[3, 4, 5]])                     # Initial seed
for i in range(2):
    current = next_pythagorean_triples(np.array(current))     # Next iteration
print(current)
```

test case from https://deque.blog/2019/01/11/pythagorean-triples-in-modern-c/

PR #1194, opened 2019-01-28
Fix "Correct support of assignment of a transposed view through a slice"

serge-sans-paille:
This fixes commit 8b4f22aa0ef0a33308d87b2dcb7d9b4cef3a6187.


serge-sans-paille:
Fixes #1189 
Fixes #1183

PR #1196, opened 2019-01-29
numpy fft: Copy twiddle factors into local array to make algo thread safe

jeanlaroche:
The core C++ routines in numpy.fft are not thread safe because they modify the twiddle factors during the function call (the get restored at the end).
This is to avoid the threading issue: I create a temp copy of the twiddle factors before calling the fft routine. It's not great but it does fix the issue.

### DIFF
--- a/pythran/pythonic/numpy/fft/irfft.hpp
+++ b/pythran/pythonic/numpy/fft/irfft.hpp
@@ -43,16 +43,35 @@ namespace numpy
 
       // Create the twiddle factors. These must be kept from call to call as
       // it's very wasteful to recompute them.
+      // The call to npy_irfftf makes changes to the twiddle factor buffer (then
+      // restores it) so it's
+      // not thread safe. To avoid this:
+      // I keep one copy of the twiddle factors that no thread is allowed to
+      // modify. Then each thread
+      // has its own copy of it that it can modify at will.
+      // This is from fftpack_litemodule.c
+      static std::map<long, std::vector<double>> all_twiddles_irfft_global;
+      static thread_local std::map<long, std::vector<double>>
+          all_twiddles_irfft_local;
       mtx_irfft.lock();
-      static std::map<long, types::ndarray<double, types::pshape<long>>>
-          all_twiddles_irfft;
-      if (all_twiddles_irfft.find(NFFT) == all_twiddles_irfft.end()) {
+      if (all_twiddles_irfft_global.find(NFFT) ==
+          all_twiddles_irfft_global.end()) {
+        // This happens only once across all threads for a give NFFT
         // Insert a new twiddle array into our map and initialize it
-        all_twiddles_irfft.insert(std::make_pair(
-            NFFT, types::ndarray<double, types::pshape<long>>(
-                      {(long)(2 * NFFT + 15)}, __builtin__::None)));
-        npy_rffti(NFFT, (double *)all_twiddles_irfft[NFFT].buffer);
+        all_twiddles_irfft_global.emplace(NFFT,
+                                          std::vector<double>(2 * NFFT + 15));
+        // std::cout << "INIT GLOBAL " << NFFT << " \n";
+        // This is the (costly) initialization of the array.
+        npy_rffti(NFFT, all_twiddles_irfft_global[NFFT].data());
       }
+
+      if (all_twiddles_irfft_local.find(NFFT) ==
+          all_twiddles_irfft_local.end()) {
+        // This happens once per thread.
+        all_twiddles_irfft_local.emplace(NFFT, all_twiddles_irfft_global[NFFT]);
+      }
+
+      double *twiddle_buffer = all_twiddles_irfft_local[NFFT].data();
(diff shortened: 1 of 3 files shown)


pythran/pythonic/numpy/fft/irfft.hpp
```diff
@@ -43,24 +43,44 @@ namespace numpy
...
       // Create the twiddle factors. These must be kept from call to call as
       // it's very wasteful to recompute them.
+      // The call to npy_irfftf makes changes to the twiddle factor buffer (then
+      // restores it) so it's
+      // not thread safe. To avoid this:
+      // I keep one copy of the twiddle factors that no thread is allowed to
+      // modify. Then each thread
+      // has its own copy of it that it can modify at will.
+      // This is from fftpack_litemodule.c
+      static std::map<long, double *> all_twiddles_irfft_global;
```
serge-sans-paille: Can you do that with vector (and without vector copy of course)?

jeanlaroche: Done. Let me know if that's what you had in mind. Indeed, it's a lot simpler that way!

pythran/pythonic/numpy/fft/irfft.hpp
```diff
@@ -43,24 +43,39 @@ namespace numpy
...
-        npy_rffti(NFFT, (double *)all_twiddles_irfft[NFFT].buffer);
+        all_twiddles_irfft_global.emplace(NFFT, std::vector<double>(2 * NFFT + 15));
+        // std::cout << "INIT GLOBAL " << NFFT << " \n";
+        // This is the (costly) initialization of the array.
+        npy_rffti(NFFT, all_twiddles_irfft_global[NFFT].data());
       }
+
+      if (all_twiddles_irfft_local.find(NFFT) == all_twiddles_irfft_local.end()) {
+        // This happens once per thread.
+        all_twiddles_irfft_local.emplace(NFFT, std::vector<double>(all_twiddles_irfft_global[NFFT]));
```
serge-sans-paille: you don't need to call  `std::vector<double>(all_twiddles_irfft_global[NFFT])` here, `all_twiddles_irfft_global[NFFT]` should be enough. Note that if the global value doesn't change, you could store a pointer to it, avoiding an extra copy (like the global has ownership, and the local just stores aview (you could use a unique_pointer to reflect that, but dont worry with that). If the local needs a copy, then just removin the extra ``std::vector`` is ok.

pythran/pythonic/numpy/fft/rfft.hpp
```diff
@@ -43,25 +43,39 @@ namespace numpy
...
+      if (all_twiddles_rfft_global.find(NFFT) ==
+          all_twiddles_rfft_global.end()) {
+        // This happens only once across all threads for a give NFFT
         // Insert a new twiddle array into our map and initialize it
-        all_twiddles_rfft.insert(std::make_pair(
-            NFFT, types::ndarray<double, types::pshape<long>>(
-                      {(long)(2 * NFFT + 15)}, __builtin__::None)));
-        npy_rffti(NFFT, (double *)all_twiddles_rfft[NFFT].buffer);
+        all_twiddles_rfft_global.emplace(NFFT, std::vector<double>(2 * NFFT + 15));
+        // std::cout << "INIT GLOBAL " << NFFT << " \n";
```
serge-sans-paille: please remove this comment from the final PR.

pythran/pythonic/numpy/fft/irfft.hpp
jeanlaroche: The local really needs a copy because otherwise it would modify the global and screw up everybody else! So when I call emplace(NFFT,existing_array) what's put in the map is a *copy* of existing_array? I'm not very clear on which situtations emplace avoids a copy. But I'll do what you recommend.

serge-sans-paille:
@jeanlaroche could you add a test case where OpenMP is involved (to test multi threading) and pass clang-format on the modified C++ header, and finally squash all commits?

jeanlaroche:
I don't know anything about openMP, will try to educate myself and ping you if I get stuck

pythran/tests/test_numpy_fft.py
```diff
@@ -40,6 +40,18 @@ def test_rfft_10(self):
...
 
+    # Test parallel:
+    def test_rfft_12(self):
+        self.run_test('''
+import numpy as np
+def test_rfft_12(x):
+    out = []
+    #omp parallel for
+    for ii in range(20):
+        out.append(np.fft.rfft(x))
```
serge-sans-paille: unfortunately, this is not thread safe because the list container is not thread safe.
You'd better initialize `out` as `out = [np.empty_like(x, dtype=complex) for i in range(20)]` and then access each element of `out`

jeanlaroche: Ooops, missed that! Should be corrected now.

serge-sans-paille:
Nice job! You leveled up in OpenMP :-)

PR #1200, opened 2019-02-03
Fix/none loop break continue

serge-sans-paille:


serge-sans-paille:
@diorcety this commit changes various aspect of NoneType and is None handling. Can you check it's still ok with your codebase? Thanks!

Issue #1206, opened 2019-02-07
NoneType != support


ge-smile:
Hi, 

We are missing "!=" operator overloaded in pythran/pythran/pythonic/types/NoneType.hpp. 

Also for "==" operator for NoneType, we seem not consider the case when O is None type. The function I mean is:

  template<class T>
  template<class O>
  bool none<T, false>::operator==(O const &t) const
  {
    return !is_none && static_cast<const T &>(*this) == t;
  }

Thanks,
Alice

serge-sans-paille:
Thanks for reporting! I'll have a look.

serge-sans-paille:
Concerning case when O is none, there is an overload

```
  template <class T>
  bool none<T, false>::operator==(none_type const &) const
  {
    return is_none;
  }
```

that should handle it.

I confirm that `operator!=` is missing, i'm adding it.

serge-sans-paille:
Fixed by #1207

ge-smile:
Thanks for the fix!

PR #1207, opened 2019-02-07
Support != for none type

serge-sans-paille:


serge-sans-paille:
This should fix #1206

Issue #1181, opened 2019-01-22
Values in tuple are not modifiables


diorcety:
The following example doesn't work well when compiled with pythran

```
def xx(t, v):
    b, _ = t
    b[0] += v

#pythran export fct(int)
def fct(a):
    t = ([0], 12)
    xx(t, a)
    b, _ = t
    return b[0]
```

diorcety:
Disabling extra optimizations solve this issue

diorcety:
This is dead_code_elimination optimization

diorcety:
The issue is: `b, _ = t`
Dead code elimination assume that `xx(t, a)` is pure because t has no argument_effect due to the tuple unpacking

serge-sans-paille:
ok, that's an aliasing issue :-)

diorcety:
I just found a fix if you want :)

serge-sans-paille:
of course :-)

serge-sans-paille:
@diorcety the code now compiles fine, can you confirm and close the issue?

diorcety:
Yes it's good!

PR #1210, opened 2019-02-11
Fix bug #1181 aliasing issue

diorcety:


serge-sans-paille:
Thanks @diorcety It's a pleasure to see such contributions!

PR #1212, opened 2019-02-11
Fix str assign behaviour from a str slice

### LINKED ISSUE #1211: String variable assign modify the original string content
str operator= behaviour is not correct

```
def sample_datatype(value):
    definitions = [ 
        ('LEFT-', 1), 
        ('RIGHT-', 2), 
        ('', 3)
    ]   

    plate_number = None
    for definition in definitions:
        s, n = definition
        if value.startswith(s):
            plate_number = n 
            value = value[len(s):]
            break
    if plate_number is None:   # Comment for make it works
       raise ValueError("Invalid value")
    return None

#pythran export aa(str)
def aa(s1):
    sample_datatype(s1)
    sample_datatype(s1)
    return s1

```

diorcety:
Fix #1211

### DIFF
--- a/pythran/pythonic/types/str.hpp
+++ b/pythran/pythonic/types/str.hpp
@@ -337,6 +337,7 @@ namespace types
     auto &other_data = other.get_data();
     auto &other_slice = other.get_slice();
     auto other_size = other.size();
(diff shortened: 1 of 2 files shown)


diorcety:
Still some Travis CI more than the PR itself

PR #1215, opened 2019-02-13
Add id defination to types::str

ge-smile:
So we could tell if types::str is None or not. Currently it will fail if we do the None check.

### DIFF
--- a/pythran/pythonic/types/str.hpp
+++ b/pythran/pythonic/types/str.hpp
@@ -286,6 +286,11 @@ namespace types
     *data = oss.str();
   }
 
+  intptr_t str::id() const
+  {
+    return reinterpret_cast<intptr_t>(&(*data));
+  }
+  
   str::operator char() const
   {
     assert(size() == 1);


serge-sans-paille:
Thanks for the PR! Some things you need to do

1. add the header from #1214 to that PR
2. add a test case under test/test_str.py
3. squash everything to have a clean history with a single commit

Are you ok with that?

serge-sans-paille:
@ge-smile up!

serge-sans-paille:
@ge-smile: can you provide me with a failing test case?

ge-smile:
Hi Sorry for the late reply. 

When I translated the following python function to c++, it would fail. 
def test_str(a:str):
   if a is None:
          return None
  else:
         return True

serge-sans-paille:
@ge-smile: this example works perfectly fine with master version; Which version of Python are you using?

ge-smile:
The version I'm using is 0.8.4. Probably it is too old?

serge-sans-paille:
It is :-) Any reason you're stuck on that version?

serge-sans-paille:
Already implemented, see https://github.com/serge-sans-paille/pythran/blob/master/pythran/pythonic/include/types/str.hpp#L188

PR #1216, opened 2019-02-14
Fix std::get with NoneType on Windows

diorcety:


serge-sans-paille:
Failed on clang-format, can you fix that? clang-format -3.6 or 3.7 should be ok.

diorcety:
Seems good now. Sorry

PR #1217, opened 2019-02-14
fix none-comparison

wizardxz:
`data` is a fundamental type, cannot equal with `t` which may be `none<>` type
on the contrary, `t` will be able to equal with `data`


serge-sans-paille:
@wizardxz can you provide a test example under `pythran/tests/test_none.py`? Something that was not working before your PR.

serge-sans-paille:
up!

PR #1221, opened 2019-02-19
Improve Windows compilation

diorcety:
Few windows improvements:
cctype missing for std::digit (maybe other headers have this issue)
struct/class warning
Support clang-cl (_MSC_VER + __clang__)


serge-sans-paille:
I'm so happy you're doing this and I'm not :-)

PR #1238, opened 2019-03-06
Document some of Pythran limitations

serge-sans-paille:


serge-sans-paille:
@nbecker this is a follow up to the thread you started on https://www.freelists.org/post/pythran/Bizarre-problem

Issue #1235, opened 2019-03-04
append compile failure


nbecker:
This code fails to compile, seems append is the issue. 
pythran version is master and already has fix for append scalar.

https://gist.github.com/nbecker/50b30eebdade692c2986762e1746ca1b

```
pythran _grow.py
/tmp/tmprh9h31tc.cpp:1026:24: error: no matching function for call to object of type '(anonymous namespace)::pythonic::numpy::functor::append'
      typedef decltype(std::declval<__type21>()(std::declval<__type56>(), std::declval<__type5>())) __type57;
                       ^~~~~~~~~~~~~~~~~~~~~~~~
/tmp/tmprh9h31tc.cpp:1796:33: note: in instantiation of template class '__pythran__grow::grow::type<long, (anonymous namespace)::pythonic::types::ndarray<std::complex<double>, (anonymous namespace)::pythonic::types::pshape<long> >, double, double, bool>' requested here
typename __pythran__grow::grow::type<long, pythonic::types::ndarray<std::complex<double>,pythonic::types::pshape<long>>, double, double, bool>::result_type grow0(long&& P, pythonic::types::ndarray<std::complex<double>,pythonic::types::pshape<long>>&& pts, double&& r, double&& alpha, bool&& randomize) 
                                ^
/home/nbecker/.local/lib/python3.7/site-packages/pythran/pythonic/include/numpy/append.hpp:53:3: note: candidate template ignored: substitution failure [with Types = <(anonymous namespace)::pythonic::types::numpy_vexpr<(anonymous namespace)::pythonic::types::ndarray<std::complex<double>, (anonymous namespace)::pythonic::types::array<long, 1> >, (anonymous namespace)::pythonic::types::ndarray<long, (anonymous namespace)::pythonic::types::pshape<long> > >, std::complex<double>>]: no matching function for call to 'append'
  DEFINE_FUNCTOR(pythonic::numpy, append);
  ^              ~~~~~~~~
...
```

serge-sans-paille:
Thanks for reporting o/

PR #1239, opened 2019-03-06
Extend numpy.append support to all numpy expr

serge-sans-paille:
Not the most efficient implementation though.


serge-sans-paille:
Should fix #1235

PR #1244, opened 2019-03-14
Fix issue with angle

jeanlaroche:
This fixes the discrepancy between numpy and pythran for the angle() function.
Numpy uses a 4 quadrant arc-tangent, but pythran was using a regular 2 quadrant arc-tangent.

### DIFF
--- a/pythran/pythonic/include/numpy/angle_in_rad.hpp
+++ b/pythran/pythonic/include/numpy/angle_in_rad.hpp
@@ -18,7 +18,7 @@ namespace numpy
   {
     template <class T>
     auto angle_in_rad(T const &t)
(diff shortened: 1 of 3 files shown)


serge-sans-paille:
The PR looks super clear to me, thanks for the time spent in searching for the origin of the issue. o/

serge-sans-paille:
@jeanlaroche you forgot to pass clang-format on the code :-)

jeanlaroche:
Are you serious? I need to pass clang-format when I changed atan->atan2 and a / to a , ?
I guess I removed a couple lines as well. OK I'll do it.

jeanlaroche:
Done

PR #1247, opened 2019-03-16
Also set python version when generating pure c++

### LINKED ISSUE #1246: int/float divide issue when using C++ code generated by pythran.
In python3, x=3/2 returns 1.5, but this isn't the case when doing the following.
Use this python code
```
def test(a,b):
    return a/b
```
put it in newBug.py and create C++ code using:
```
pythran -e newBug.py
```
Then compile the following C++ code:
```
void test(int a, int b) {
    auto ret = __pythran_newBug::test()(a,b);
    std::cout << ret << "\n";
}
int main() {
    test(3,2);
    return 0;
}
```
Using this command:
```
c++ -std=c++11 -w -I/Users/jlaroche/Dev/BuildVEnv/lib/python3.6/site-packages/pythran -I/Users/jlaroche/Dev/BuildVEnv/lib/python3.6/site-packages/numpy/core/include -I/Users/jlaroche/Dev/BuildVEnv/include/python3.6m -F. test.cpp -o test.exe
```

Then run test.exe.
You'll get an answer of 1, instead of 1.5 as you'd expect from python.
Of course, this makes sense, you're passing integers to __pythran_newBug::test() and the function uses an integer divide, but that's inconsistent with what python3 does.

serge-sans-paille:
Fix #1246

### DIFF
--- a/pythran/toolchain.py
+++ b/pythran/toolchain.py
@@ -34,6 +34,7 @@ import os.path
 import shutil
 import glob
 import hashlib
+import sys
 from functools import reduce
 
 logger = logging.getLogger('pythran')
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
@jeanlaroche can you confirm this fixes your issue?

jeanlaroche:
I will on Monday! Thanks for this quick fix!!!

jeanlaroche:
Confirmed fixed! Thanks!

Issue #1252, opened 2019-03-27
Severe indexing bug x[[0,1,2,3],1]


jeanlaroche:
```
import numpy as np

# pythran export test(float[][])
def test(x):
    a = x[[1,2,3,4],1]
    print(a)
    return a
```
I'm not sure how long this has been broken, but the above code does not return the right values in pythran. if x = np.ones((10,2)) it should return array([1., 1., 1., 1.]) but the pythran version returns array([1., 1.])

jeanlaroche:
I'm finding that the code above does not compile on fix/numpy-list-slicing
(in MacOS)

jeanlaroche:
The unit test you added works because this:
```
# pythran export test()
def test():
    x = np.ones((10,2))
    a = x[(1,2,3,4),1]
    print(a)
    return a
```
does compile and work.
But not if the input array is passed as an argument apparently.

PR #1254, opened 2019-03-28
Fix slicing of an array through an array

serge-sans-paille:


serge-sans-paille:
This should fix #1252

pythran/tests/test_ndarray.py
```diff
@@ -1063,6 +1063,36 @@ def test_transposed_slice_assign2(self):
...
+                              slice_through_list1=[Tuple[int, int]])
+
+    def test_slice_through_list2(self):
+                self.run_test("""def slice_through_list2(arr):
+                                  import numpy as np;
+                                  return arr[np.array([1,3,5]),1:]""",
+                              numpy.ones((10, 20)),
+                              slice_through_list2=[NDArray[float, :,:]])
+
+    def test_slice_through_list3(self):
```
jeanlaroche: You should also test return arr[np.array((1,3,5)),1:]""",
tuple instead of list. This currently does not compile.

Issue #1185, opened 2019-01-23
None issues


diorcety:
Here the pythran file

All cases should return the same results. But few compile and none have the good behaviour

```


# pythran export aaa(int):
def aaa(a):
    x = None if a < 0 else 1
    y = None if a % 2 else 2
    z = -1

    # Doesn't compile
    #if x is not None and y is not None:  # Without test on other than none it doesn't work
    #    z = 0

    # Doesn't compile
    #if x is not None:
    #    if y is not None:
    #        return 0

    # Doesn't compile
    #if x is not None:
    #    if y is not None and a != -666:
    #        z = 0

    # Compile but wrong results
    #if x is not None:
    #    if y is not None:
    #        z = 0

    # Compile but wrong results (not the same that previous one)
    if x is not None and a != -666 and y is not None:
        z = 0
    return z
```

The test code

```
import aa

print(aa.aaa(3))
print(aa.aaa(2))
print(aa.aaa(-2))
print(aa.aaa(-3))

```

PR #1256, opened 2019-03-31
Fix/none handler

serge-sans-paille:


serge-sans-paille:
@diorcety can you confirm it fixes #1185 ?

diorcety:
Can't tell until #1257

PR #1267, opened 2019-04-09
Feature/static list

serge-sans-paille:


serge-sans-paille:
@jeanlaroche this one is for you, can you try it out?

jeanlaroche:
It almost works. But here's where it does not:
```
import numpy as np

#pythran export export StridedSlice(float[][][], int[], int[], int[])
def StridedSlice(x,begins, ends, strides):
    slices = tuple([slice(b, e if e else None, s) for b, e, s in zip(begins,ends, strides)])
    return x[slices]

# pythran export test()
def test():
    x = np.zeros((10,10,10))
    return StridedSlice(x,[0,2,3], [5,0,7], [1,1,1])
```
This does not compile. But if you remove the first #pythran export, then it compiles (and runs when used through the test() call...

jeanlaroche:
Should this work?
```
import numpy as np

def StridedSlice(x,begins, ends, strides):
    slices = tuple([slice(b, e if e else None, s) for b, e, s in zip(begins,ends, strides)])
    return x[slices]

# pythran export test(float [][][])
def test(x):
    return StridedSlice(x,[0,2,3], [5,0,7], [1,1,1])
```
StridedSlice is called with arguments of known size, so I think this should work. But currently it does not compile...

serge-sans-paille:
@jeanlaroche this PR is getting fat, but it should be ok now; Can you try it?

serge-sans-paille:
@jeanlaroche I've made a few update, is that still okay on your side?

@diorcety, you're issue with this PR is related to handling of type inference withing is none condition :-/

diorcety:
Maybe linked to the previous none condition inference ?

serge-sans-paille:
@diorcety yes :-/ basically I need to find a way to perform conditional type inference when in a static if...

jeanlaroche:
Serge, I'm getting a compilation error on this branch, while everything's fine in master. 
You can see the error here: https://pastebin.com/AY2tyC3h 
I can send you the offending code, it's a bit large, and I can't post it publicly...
Let me know what you want me to do.

serge-sans-paille:
On Tue, Apr 23, 2019 at 04:03:23PM +0000, jean laroche wrote:
> Serge, I'm getting a compilation error on this branch, while everything's fine
> in master.
> You can see the error here: https://pastebin.com/AY2tyC3h
> I can send you the offending code, it's a bit large, and I can't post it
> publicly...
> Let me know what you want me to do.

Yeah, please send me the offending code, or (better) a decent minimal reproducer

jeanlaroche:
Here's a minimal reproducer
```
import numpy as np
#pythran export hanning(int)
def hanning(M):
    if M < 1:
        return np.array([])
    if M == 1:
        return np.ones(1, float)
    n = np.arange(0,float(M))
    return 0.5 - 0.5*np.cos(2.0*np.pi*n/(M-1))

```

PR #1268, opened 2019-04-12
Implement expand_dims

jeanlaroche:
np.expand_dims was missing...

### DIFF
--- a/pythran/pythonic/numpy/expand_dims.hpp
+++ b/pythran/pythonic/numpy/expand_dims.hpp
@@ -1,0 +1,36 @@
+#ifndef PYTHONIC_NUMPY_EXPAND_DIMS_HPP
+#define PYTHONIC_NUMPY_EXPAND_DIMS_HPP
+
+#include "pythonic/include/numpy/expand_dims.hpp"
+#include "pythonic/utils/functor.hpp"
+#include "pythonic/include/utils/array_helper.hpp"
+#include "pythonic/numpy/asarray.hpp"
+
+PYTHONIC_NS_BEGIN
+
+namespace numpy
+{
+  template <typename T>
+  types::ndarray<typename T::dtype, types::array<long, T::value + 1>>
+  expand_dims(T const &input, int axis)
+  {
+    const long N = T::value;
+    if (axis == -1)
+      axis += N + 1;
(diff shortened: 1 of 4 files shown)


pythran/pythonic/numpy/expand_dims.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+PYTHONIC_NS_BEGIN
+
+namespace numpy
+{
+  template <typename T>
+  types::ndarray<typename T::dtype, types::array<long, T::value + 1>>
+  expand_dims(T const &input, int axis)
+  {
+    const int N = std::tuple_size<decltype(input._shape)>::value;
```
serge-sans-paille: could be `const long N = T::value`

pythran/pythonic/numpy/expand_dims.hpp
```diff
@@ -0,0 +1,40 @@
...
+    auto in_shape = sutils::array(input._shape);
+    long ii, jj;
+    for (ii = jj = 0; ii < N + 1; ii++) {
+      if (ii == axis) {
+        dim_array[ii] = 1;
+      } else {
+        dim_array[ii] = in_shape[jj++];
+      }
+    }
+    sutils::push_front_t<decltype(input._shape), long> new_shape;
```
serge-sans-paille: this is incorrect, as your return type specifies ` types::array<long, T::value + 1>` as return type. why not reshaping directly using `dim_array` ?

pythran/pythonic/numpy/expand_dims.hpp
jeanlaroche: ah ah :) Thanks! There had to be a simpler way. You can see I'm still not familiar with the classes used in pythran!

pythran/pythonic/numpy/expand_dims.hpp
jeanlaroche: Yes, much simpler, I didn't realize I could pass an array, instead of a pshape.

jeanlaroche: Serge, if I try reshaping with
```
return ndarray::reshape(input, dim_array);
```
compilation fails. I'm having a hard time figuring out whether reshape can accept an array as input....

PR #1271, opened 2019-04-19
Implement convolve

jeanlaroche:


jeanlaroche:
Hold on, the implementation is correct, but performance is poor, I have to look into that.

jeanlaroche:
OK It's in a good shape now. The only issue I still see is that performance is worse than numpy when 1 of the arrays is very small (ie., 10?). That's because numpy uses a special implementation small_correlate with partial unrolling etc.
Also: I can't handle views inside correlate, I assume that's normal and expected, so I convert the inputs using as_array... Let me know if that's not what I should do...

pythran/pythonic/numpy/correlate.hpp
```diff
@@ -0,0 +1,182 @@
...
+  {
+    double val = 0;
+    for (int i = 0; i < N; i++)
+      val += *a++ * *b++;
+    return val;
+  }
+
+  // versions for which we don't have Blas support
+  double dot(int N, double *a, float *b, float no_conj)
+  {
```
serge-sans-paille: could be a call to numpy::dot, right?

pythran/pythonic/numpy/correlate.hpp
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  std::complex<double> dot(int N, std::complex<double> *a,
+                           std::complex<double> *b, float no_conj)
+  {
+    double temp[2];
+    cblas_zdotc_sub(N, b, 1, a, 1, &temp);
+    return std::complex<double>(temp[0], no_conj * temp[1]);
+  }
+
```
serge-sans-paille: same here.

pythran/pythonic/numpy/correlate.hpp
```diff
@@ -0,0 +1,182 @@
...
+    typedef typename decltype(inA + inB)::dtype out_type;
+    // At this point, handling views would slow things down tremendously
+    auto inA_ = functor::asarray{}(inA);
+    auto inB_ = functor::asarray{}(inB);
+    auto *ina_pt = (typename A::dtype *)inA_.buffer;
+    auto *inb_pt = (typename B::dtype *)inB_.buffer;
+
+    auto outN = 0;
+    if (type == "full")
```
serge-sans-paille: it's probably better to avoid repeated string comparison, storing the result in a bool.

pythran/pythonic/numpy/correlate.hpp
```diff
@@ -0,0 +1,182 @@
...
+      outN = NA + NB - 1;
+    else if (type == "valid")
+      outN = NA - NB + 1;
+    else if (type == "same")
+      outN = NA;
+    types::ndarray<out_type, types::pshape<long>> out = {outN, out_type()};
+    out_type *out_ptr = (out_type *)out.buffer;
+
+    // For small correlations, numpy uses small_correlate, far more efficient.
+    // /Users/jlaroche/temp/NumpySrc/numpy/core/src/multiarray/arraytypes.c.src
```
serge-sans-paille: why don't you use the same strategy here? Do you prefer implementing that in a further PR?

serge-sans-paille:
@jeanlaroche the only major change would be to use numpy::dot instead of reimplemeting it in correlate, otherwise LGTM.

Concering the temporary cast to ndarray, it's not trivially avoidable, but one could use a scratch buffer of fixed size to avoid the large heap allocation... you could probably do it for small size but that's not critical if current performance are on par with numpy.

serge-sans-paille:
@jeanlaroche the only major change would be to use numpy::dot instead of reimplemeting it in correlate, otherwise LGTM.

Concerning the temporary cast to ndarray, it's not trivially avoidable, but one could use a scratch buffer of fixed size to avoid the large heap allocation... you could probably do it for small size but that's not critical if current performance are on par with numpy.

jeanlaroche:
Ah yes, I didn't think of using numpy::dot, of course, you've already done the job, plus you probably already support type mix etc.
One question: if mixing float and double, it's a lot more advantageous for correlate to convert the float array -> double then call correlate. The extra copy is nothing compared to the savings offered by the faster dot (because we do so many dots). In fact, it's also true of complex/float, int float etc etc. 
So I need code to promote the two arrays to the output type before I call the dot function (only one is promoted of course). I supposed I'll use as_type(), is that that you'd do?

pythran/pythonic/numpy/correlate.hpp
jeanlaroche: I looked at their implementation, it's pretty gnarly. It would probably be best to copy it straight from numpy but that also requires a bit of editing because of types etc...

jeanlaroche:
I changed the code to use numpy::dot, but I find my solution rather inelegant (to use numpy::dot I had to create ndarrays from the data pointers, which would be too long to do over and over again, so I create them once and adjust pointers and pshape as needed). Let me know if there's a more elegant solution

serge-sans-paille:
@jeanlaroche let me check that.

jeanlaroche:
Your patch works, after I fixed a small index error. Now how do I push this up to my fork, since I had to merge upstream/feature/broader-dot-support to test it?

serge-sans-paille:
On Tue, Apr 23, 2019 at 01:58:51PM -0700, jean laroche wrote:
> Your patch works, after I fixed a small index error. Now how do I push this up
> to my fork, since I had to merge upstream/feature/broader-dot-support to test
> it?

I've updated the patch. If it still works for you as is, let's merge the patch first, then you rebase your branch on it.

jeanlaroche:
> On Tue, Apr 23, 2019 at 01:58:51PM -0700, jean laroche wrote: Your patch works, after I fixed a small index error. Now how do I push this up to my fork, since I had to merge upstream/feature/broader-dot-support to test it?
> I've updated the patch. If it still works for you as is, let's merge the patch first, then you rebase your branch on it.

Yes, it does with your latest changes... I rebased my branch on upstream/feature/broader-dot-support. I'll go ahead and force push that.

PR #1273, opened 2019-04-24
Handle rvalue reference in __combined

### LINKED ISSUE #1266: NoneType with rvalue reference
When compiling this code
```
import numpy as np

#pythran export to_tuple(float list)
def to_tuple(l):
    def aaa(l):
        return tuple([None if np.isnan(a) else a for a in l])
    return aaa(np.subtract(np.zeros((len(l),)), 2))
```

I have the following issue:
`error: base specifier must name a class`

serge-sans-paille:
This should fix #1266 cc @diorcety


diorcety:
👍

PR #1274, opened 2019-04-25
Fix errors in computing accuracy of means in several tests

jeanlaroche:
Several tests were using tests like abs(mean(a)) < 3.01 to test that mean(a) was close to 3.
This should have been abs(mean(a)-3) < 0.01 .


serge-sans-paille:
Correct! Thanks o/

PR #1288, opened 2019-05-11
Fix packaging of numpy/interp_core.cpp

serge-sans-paille:
Problem due to file extension, should be a .hpp t be correclty bundled.

Thanks to Jean Laroche for spotting the issue!


serge-sans-paille:
Thanks @jeanlaroche for spotting that.

PR #1295, opened 2019-05-16
Split doc strings in 16380 characters string parts (maximum MSVC capa…

diorcety:
Allow to compile with msvc when string length is more than 16380 characters ([MSVC limit](https://docs.microsoft.com/fr-fr/cpp/error-messages/compiler-errors-1/compiler-error-c2026?view=vs-2019))

### DIFF
--- a/pythran/cxxgen.py
+++ b/pythran/cxxgen.py
@@ -461,9 +461,12 @@ class PythonModule(object):
         self.metadata['moduledoc'] = moduledoc
 
     def docstring(self, doc):
-        return '"%s"' % (dedent(doc).replace('"', '\\"')
-                                    .replace('\n', '\\n')
-                                    .replace('\r', '\\r'))
+        return self.splitstring(dedent(doc).replace('"', '\\"')
+                                           .replace('\n', '\\n')
+                                           .replace('\r', '\\r'))
+
+    def splitstring(self, doc):
+        return '"{}"'.format('\\n""'.join(doc.split('\\n')))
 
     def add_to_preamble(self, *pa):
         self.preamble.extend(pa)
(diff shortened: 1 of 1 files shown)


pythran/cxxgen.py
```diff
@@ -461,9 +461,14 @@ def __init__(self, name, docstrings, metadata):
...
-                                    .replace('\n', '\\n')
-                                    .replace('\r', '\\r'))
+        return self.splitstring(dedent(doc).replace('"', '\\"')
+                                           .replace('\n', '\\n')
+                                           .replace('\r', '\\r'))
+
+    def splitstring(self, doc):
+        if len(doc) == 0:
+            return '""'
+        return "".join(['"%s"' % s for s in wrap(doc, 16380)])
```
serge-sans-paille: An interesting choice. I wonder if we could take an approach that would prevent this magic (uncommented) number, something along

```
return "".join('"{}\\n"'.format(s) for s in doc.split('\\n'))
```

pythran/cxxgen.py
```diff
@@ -29,7 +29,7 @@
 
 from __future__ import division
 
-from textwrap import dedent
+from textwrap import dedent, wrap
```
serge-sans-paille: sorry to bother you with that, but it seems wrap is no longer used :-/

pythran/cxxgen.py
```diff
@@ -461,9 +461,14 @@ def __init__(self, name, docstrings, metadata):
...
     def docstring(self, doc):
-        return '"%s"' % (dedent(doc).replace('"', '\\"')
-                                    .replace('\n', '\\n')
-                                    .replace('\r', '\\r'))
+        return self.splitstring(dedent(doc).replace('"', '\\"')
+                                           .replace('\n', '\\n')
+                                           .replace('\r', '\\r'))
+
+    def splitstring(self, doc):
+        if len(doc) == 0:
```
serge-sans-paille: could be ``if not doc``

serge-sans-paille:
@diorcety perfect! can youjust squash your commits?

diorcety:
done

serge-sans-paille:
Thanks!

PR #1307, opened 2019-06-14
\o/ was not intended as a backslash-escape.  Tell python.

DWesl:
Avoid SyntaxWarnings due to invalid backslash-escapes.


serge-sans-paille:
\o/ LGTM \o/

PR #1314, opened 2019-06-21
Do not outline local definitions when building static_if

### LINKED ISSUE #1171: if is None produces and unknown identifier error
If the comment inside this code is uncommented, the pythran returns the following error:

> CRITICAL: I am in trouble. Your input file does not seem to match Pythran's constraints...
> E: identifier b unknown, either because it is an unsupported intrinsic, the input code is faulty, or... pythran is buggy. (line 4)

```
#pythran export test(int):
def test(i):
    a = 0 
    c = None
    if i % 2 == 0:
    #    c = 1
    #if c is not None:
        a = 2 
        b = 3 
    return a

```

serge-sans-paille:
Fix #1171

### DIFF
--- a/pythran/analyses/is_assigned.py
+++ b/pythran/analyses/is_assigned.py
@@ -2,7 +2,6 @@
 
 from pythran.passmanager import NodeAnalysis
 
-from collections import defaultdict
 import gast as ast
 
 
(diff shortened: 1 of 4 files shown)


diorcety:
I manage to reduce the issue. And this is very strange:

```
import numpy as np

#pythran export abcd(str:str dict)
def abcd(headers):
    def get_header():
            return None

    lpix = float(headers.get('a'))

    diameter = None
    print("TEST %s" % (lpix))
    if lpix is not None and lpix > 0.0:
        print("TEST2 %s" % (lpix))
        diameter = 20.0
        print("TEST3 %s" % (diameter))
    return diameter
```

```
import aa

print(aa.abcd({'a': '10'}))
```

Comment the unused function get_header and the abcd function returns 20.0 as it should do

serge-sans-paille:
@diorcety I included your test case to the test bed and fixed the issue. Thanks *a lot*  for the reproducer

diorcety:
Seems good now 👍

PR #1332, opened 2019-08-01
Fix compilation with Python < 3.6.1

diorcety:
PySlice_Unpack is only supported since 3.6.1


serge-sans-paille:
Thanks @diorcety

PR #1333, opened 2019-08-01
Run validation with assert ON

serge-sans-paille:
This finds more bugs, and a fix for numpy_expr is attached.


serge-sans-paille:
@jeanlaroche this should fix the bug we discussed about on IRC.
As a side note: compile your native builds with `-DNDEBUG`

jeanlaroche:
That's awesome. Let me try now... Thanks!!!

jeanlaroche:
I confirm that this fixes my problem. No more assert!

PR #1349, opened 2019-09-03
Implement option arguments

jeanlaroche:
Not sure whether we want the command-line arguments to override the config flags or not. But this is a start.

### DIFF
--- a/pythran/config.py
+++ b/pythran/config.py
@@ -13,9 +13,9 @@ logger = logging.getLogger('pythran')
 
 
 def get_paths_cfg(
-    sys_file='pythran.cfg',
-    platform_file='pythran-{}.cfg'.format(sys.platform),
-    user_file='.pythranrc'
+        sys_file='pythran.cfg',
+        platform_file='pythran-{}.cfg'.format(sys.platform),
+        user_file='.pythranrc'
 ):
     """
     >>> os.environ['HOME'] = '/tmp/test'
@@ -44,7 +44,7 @@ def get_paths_cfg(
     return {"sys": sys_config_path, "platform": platform_config_path, "user": user_config_path}
 
 
-def init_cfg(sys_file, platform_file, user_file):
+def init_cfg(sys_file, platform_file, user_file, config_args=None):
     paths = get_paths_cfg(sys_file, platform_file, user_file)
     sys_config_path, platform_config_path, user_config_path = paths["sys"], paths["platform"], paths["user"]
 
@@ -52,6 +52,19 @@ def init_cfg(sys_file, platform_file, user_file):
     for required in (sys_config_path, platform_config_path):
         cfgp.read([required])
     cfgp.read([user_config_path])
+
+    if config_args is not None:
+        # Override the config options with those provided on the command line e.g. compiler.blas=pythran-openblas
+        for arg in config_args:
+            try:
+                lhs, rhs = arg.split('=',maxsplit=1)
+                section, item = lhs.split('.')
+                if not cfgp.has_section(section):
+                    cfgp.add_section(section)
+                cfgp.set(section, item, rhs)
+            except Exception as e:
+                pass
+        pass
 
     return cfgp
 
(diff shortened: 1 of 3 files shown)


serge-sans-paille:
I think we can come up with a better design. I'll do a preparatory commit for it, then wave at you.

pythran/config.py
```diff
@@ -163,7 +181,9 @@ def parse_define(define):
             extension['include_dirs'].append(numpy.get_include())
 
         # blas dependency
-        user_blas = cfg.get('compiler', 'blas')
+        user_blas = get_from_ext('compiler.blas')
```
serge-sans-paille: this shouldn't be needed if you updated the config correctly using the args

pythran/config.py
```diff
@@ -149,7 +157,17 @@ def parse_define(define):
 
     for k, w in extra.items():
         extension[k].extend(w)
-    if cfg.getboolean('pythran', 'complex_hook'):
+
```
serge-sans-paille: Instead of adding the complexity here, I think you should udpate ``init_cfg`` to accept the extra config argument. And as you asked previously: they are more important than any config file.

jeanlaroche: OK I"ll try that.

jeanlaroche: mmm maybe I'm missing something but init_cfg(sys_file, platform_file, user_file) is called at module load time 
```
cfg = init_cfg('pythran.cfg',
               'pythran-{}.cfg'.format(sys.platform),
               '.pythranrc')
```
A that point, I don't see how I can pass the extra config arguments.

jeanlaroche: OK Take a look at it now. I think you're right, it's a lot cleaner.

pythran/config.py
```diff
@@ -53,6 +53,21 @@ def init_cfg(sys_file, platform_file, user_file):
...
+    if config_args is not None:
+        # Override the config options with those provided on the command line e.g. compiler.blas=pythran-openblas
+        for arg in config_args:
+            try:
+                lhs,rhs = arg.split('=')
+                section,item = lhs.split('.')
+                if not cfgp.has_section(section):
+                    cfgp.add_section(section)
+                cfgp.set(section,item,rhs)
+                cfgp._sections[section][item]=rhs
```
serge-sans-paille: this looks redundant

pythran/config.py
```diff
@@ -53,6 +53,21 @@ def init_cfg(sys_file, platform_file, user_file):
         cfgp.read([required])
     cfgp.read([user_config_path])
 
+    if config_args is not None:
+        # Override the config options with those provided on the command line e.g. compiler.blas=pythran-openblas
+        for arg in config_args:
+            try:
+                lhs,rhs = arg.split('=')
```
serge-sans-paille: please run pep8 on your modification :-)

pythran/config.py
```diff
@@ -104,6 +119,11 @@ def lint_cfg(cfgp, **paths):
 
 def make_extension(python, **extra):
 
+    # load platform specific configuration then user configuration
+    cfg = init_cfg('pythran.cfg',
+                   'pythran-{}.cfg'.format(sys.platform),
+                   '.pythranrc',extra.get('config',None))
```
serge-sans-paille: this looks better with one line per argument (yes that's neat-picking, but the logic looks good)

serge-sans-paille:
App veyor complains on your change, you need to remove the config argument once it's processed

pythran/config.py
jeanlaroche: oops

pythran/config.py
jeanlaroche: ah yes!

pythran/config.py
```diff
@@ -13,9 +13,9 @@
 
 
 def get_paths_cfg(
-    sys_file='pythran.cfg',
-    platform_file='pythran-{}.cfg'.format(sys.platform),
-    user_file='.pythranrc'
+        sys_file='pythran.cfg',
+        platform_file='pythran-{}.cfg'.format(sys.platform),
+        user_file='.pythranrc'
```
serge-sans-paille: this diff is unrelated to this commit, right?

pythran/config.py
```diff
@@ -53,6 +53,19 @@ def init_cfg(sys_file, platform_file, user_file):
         cfgp.read([required])
     cfgp.read([user_config_path])
 
+    if config_args is not None:
+        # Override the config options with those provided on the command line e.g. compiler.blas=pythran-openblas
+        for arg in config_args:
+            try:
+                lhs, rhs = arg.split('=')
```
serge-sans-paille: you should specify `maxsplit=1` here in case we pass `compiler.cxxflags=-fsanitize=address` or similar stuff

serge-sans-paille:
Thanks for the contribution @jeanlaroche o/

PR #1331, opened 2019-07-28
Adding new special funcs.

Chronum94:
Working on special functions support. Starting with spherical bessels.

PR #1366, opened 2019-09-19
Adding new special funcs.

serge-sans-paille:
sph_harm, spherical_jn and spherical_yn, only for integer types as first
argument, based on boost/math implementation.


serge-sans-paille:
This obsoletes #1331, but with the eval_legendre function pruned (boost & scipy implementation seem to behave differently, or there are too much constraints on the types)

PR #1390, opened 2019-10-12
Remove duplicate information for requirements

serge-sans-paille:
doc requirements are already available through doc/requirements.txt
dev requirements are to the dev discretion, no enforcement there


paugier:
Then, also change this: https://pythran.readthedocs.io/en/latest/DEVGUIDE.html#editable-installation-and-additional-packages

`python -m pip install -e .[dev]` won't work!

serge-sans-paille:
Done, I've removed this section :-) I'm not using these recommendations, I don't see a reason to advertise them to other. Let anyone dev the way he wants!

Issue #1381, opened 2019-10-08
installation manual on windows


gdementen:
Would it be possible to add some instructions on how to get a working Pythran installation on Windows (preferably using conda)? It would be highly appreciated. Yes, pythran installs fine and even translates my python code to C++ but then it tries to compile the generated C++ using my installed VC++ compiler which fails. I know VC++ is not a supported compiler but what should I use and how do I configure that?

serge-sans-paille:
We have a problem with Visual Compiler... Would that be ok to rely on clang on windows? I can try that path...

gdementen:
Sorry, I don't know (hence the question). I am an experienced Python programmer, but once it crosses the C barrier, it is out of my comfort zone.

serge-sans-paille:
Can you test branch #1391 ?

gdementen:
I just tested it and it works fine for my **simple** tests when setting the CXX variable in pythran config file (I never managed to do it using the config command line option, but that's probably just me), though it spewed a lot of warnings. However, I just also realized that a recent version of VC++ also appears to work fine (I just upgraded my old Visual Studio environment to the 2019 version, which includes VC++ 14.2) ! Thanks a lot for your work.

serge-sans-paille:
Cool, closing the issue then.

Issue #1339, opened 2019-08-12
example of Linux-OK Windows-KO code (::pythonic::numpy::sort?)


piotrbartman:
```python
import numpy as np

# pythran export test(int64[:], int)
def test(idx, length):
    idx[:length] = np.sort(idx[:length])
```

Compiles OK on Linux (gcc 8.3.0)

Fails on Windows (VS 2017, MSVC 14.14) with (sorry for non-English error messages, but changing the locale for Visual Studio seems to be beyond stack-overflow knowledge base...):

```
.../pythran\pythonic/python/core.hpp(64): warning C4244: "inicjowanie": konwersja z "Py_ssize_t" do "long", mozliwa utrata danych
.../pythran\pythonic/python/core.hpp(94): warning C4244: "inicjowanie": konwersja z "npy_intp" do "long", mozliwa utrata danych
.../pythran\pythonic/python/core.hpp(100): warning C4244: "*=": konwersja z "const npy_intp" do "long", mozliwa utrata danych
.../pythran\pythonic/python/core.hpp(146): warning C4244: "inicjowanie": konwersja z "Py_ssize_t" do "long", mozliwa utrata danych
.../pythran\pythonic/types/int.hpp(175): warning C4244: "return": konwersja z "long" do "unsigned char", mozliwa utrata danych
.../pythran\pythonic/types/int.hpp(176): warning C4244: "return": konwersja z "long" do "signed char", mozliwa utrata danych
.../pythran\pythonic/types/int.hpp(177): warning C4244: "return": konwersja z "long" do "unsigned short", mozliwa utrata danych
.../pythran\pythonic/types/int.hpp(178): warning C4244: "return": konwersja z "long" do "short", mozliwa utrata danych
.../pythran\pythonic/__builtin__/id.hpp(28): warning C4244: "return": konwersja z "uintptr_t" do "long", mozliwa utrata danych
.../pythran\pythonic/__builtin__/id.hpp(33): warning C4244: "return": konwersja z "uintptr_t" do "long", mozliwa utrata danych
.../pythran\pythonic/__builtin__/id.hpp(38): warning C4244: "return": konwersja z "uintptr_t" do "long", mozliwa utrata danych
.../pythran\pythonic/include/utils/shared_ref.hpp(99): warning C4521: "`anonymous-namespace'::pythonic::utils::shared_ref<std::vector<bool
,std::allocator<_Ty>>>": okreslono wiele konstruktor�w kopiujacych
        with
        [
[...]

PR #1391, opened 2019-10-14
clang-cl.exe support

serge-sans-paille:


serge-sans-paille:
I'll leave a day or two for #1339 or #1381 and merge that one, that's probably a big step forward without much drawbacks.

PR #1410, opened 2019-11-21
fix compile without ENABLE_PYTHON_MODULE

byronhe:
add #ifdef ENABLE_PYTHON_MODULE for
pythran -e and g++ compile without ENABLE_PYTHON_MODULE


serge-sans-paille:
Looks nice thanks!

PR #1429, opened 2019-12-31
0.9.5 - Japan

serge-sans-paille:


serge-sans-paille:
@hroncok : I've tested this branch on koji and it seems to build fine without the two existing patches, this should make your packaging work easier.

PR #1430, opened 2020-01-02
Remove py2 support

serge-sans-paille:


serge-sans-paille:
@scoder : this patch changes the naming of the builtin namespace from __builtin__ to builtins. I'll propose a patch in cython to reflect that soon.

PR #1446, opened 2020-01-29
No inlining of global variable if their initializer has a side effect

serge-sans-paille:
That would duplicate the side effect at each use site.


serge-sans-paille:
cc @jeanlaroche

Issue #1449, opened 2020-01-30
Feature request: disable pythran optimization/analyses to speed up compilation


jeanlaroche:
Title says it all. When debugging a large script by trial and error (which is often necessary) speeding up the pythran compilation would be tremendously helpful. Perhaps disabling some of the analysis would help make the compilation faster.
Note that passing the -O0 flag to pythran makes the C++ compilation faster, so that's already a small time saving.

serge-sans-paille:
`pythran --config 'pythran.optimizations='` should do the trick, but it's not. i'll investigate

serge-sans-paille:
Closed by #1450

PR #1450, opened 2020-01-31
More generic handling of compil options

serge-sans-paille:
It's now possible to specify --config 'pythran.optimizations=' on the CLI.


serge-sans-paille:
Fix #1449

PR #1452, opened 2020-01-31
Bugfix/cxxtypes recursion generate

rfiischer:
A infinite recursion was being generated due to the recursion logic using the exception handler. A fix was proposed.

### DIFF
--- a/pythran/cxxtypes.py
+++ b/pythran/cxxtypes.py
@@ -225,21 +225,18 @@ std::declval<bool>()))
                 return Type.__add__(self, other)
 
             def generate(self, ctx):
+                import sys
+                current_recursion_limit = sys.getrecursionlimit()
                 try:
                     return 'typename __combined<{}>::type'.format(
                         ','.join(ctx(t) for t in self.types))
                 except RuntimeError:
                     # this is a situation where we accept to somehow extend
                     # the recursion limit, because of degenerated trees
-                    import sys
-                    current_recursion_limit = sys.getrecursionlimit()
-                    try:
-                        return self.generate(ctx)
-                    except RuntimeError:
-                        sys.setrecursionlimit(current_recursion_limit * 2)
-                        return self.generate(ctx)
-                    finally:
-                        sys.setrecursionlimit(current_recursion_limit)
+                    sys.setrecursionlimit(current_recursion_limit * 2)
+                    res = self.generate(ctx)
+                    sys.setrecursionlimit(current_recursion_limit)
+                    return res
 
         class ArgumentType(Type):
             """


pythran/cxxtypes.py
```diff
@@ -233,13 +233,10 @@ def generate(self, ctx):
...
-                        return self.generate(ctx)
-                    except RuntimeError:
-                        sys.setrecursionlimit(current_recursion_limit * 2)
-                        return self.generate(ctx)
-                    finally:
-                        sys.setrecursionlimit(current_recursion_limit)
+                    sys.setrecursionlimit(current_recursion_limit * 2)
+                    result = self.generate(ctx)
+                    sys.setrecursionlimit(current_recursion_limit)
+                    return result
```
serge-sans-paille: this will fail to reset the recursion limit to its original value. What about

```
def generate(self, ctx):
    current_recursion_limit = sys.getrecursionlimit()
    try:
        return 'typename __combined<{}>::type'.format(
                        ','.join(ctx(t) for t in self.types))
    except RuntimeError:
        sys.setrecursionlimit(current_recursion_limit * 2)
        return self.generate(ctx)
    finally:
        sys.setrecursionlimit(current_recursion_limit)
```

serge-sans-paille:
Thanks for the PR! I proposed an upgrade, I'd be happy to read your thoughts.

pythran/cxxtypes.py
rfiischer: The suggested implementation unfortunately doesn't work :(
I don't understand why, but I think it has to do with the way Python deals with the `try\except\finally` clauses with return statements distributed among them. I tried running the following code:
```python
def generate(self, ctx, counter=0):
    import sys
    current_recursion_limit = sys.getrecursionlimit()
    print('called {}'.format(counter))
    try:
        return 'typename __combined<{}>::type'.format(
            ','.join(ctx(t) for t in self.types))
    except RuntimeError:
        print('excepted {}'.format(counter))
        sys.setrecursionlimit(current_recursion_limit * 2)
        return self.generate(ctx, counter=counter + 1)
    finally:
        print('finalled {}'.format(counter))
        sys.setrecursionlimit(current_recursion_limit)
```
and the outputs don't make any sense. I get for an example
```
called 0
called 0
called 0
excepted 0
called 1
called 0
called 0
```
It means the function got called by the module lots of times, then ran into an exception and the exception called the function by recursion, indicated by `called 1`, but then, no finally clauses were executed. It gets crazier later with lots of finally clauses setting back the limit of recursion mixed with calls and excepts. Curiously enough the change I proposed was the only one that worked. I will resume debugging to see if I can find an efficient method to setting the recursion limit back.

rfiischer: Using the proposed code, the interpreter still runs at a stack overflow error caused by the same infinite recursion.

serge-sans-paille: I'm pretty confident about the way try/finally/return interact. That being said, what about this simpler and more explicit version:

```python
def generate(self, ctx):
    current_recursion_limit = sys.getrecursionlimit()
    try:
        return 'typename __combined<{}>::type'.format(
                        ','.join(ctx(t) for t in self.types))
    except RuntimeError:
        sys.setrecursionlimit(current_recursion_limit * 2)
        res =  self.generate(ctx)
        sys.set_recursionlimit(current_recursion_limit )
        return res
````

Comment (serge-sans-paille) [reply pythran/cxxtypes.py]: @rfiischer: up?

Comment (rfiischer) [reply pythran/cxxtypes.py]: Correcting the typo and adding `import sys` makes the code work perfectly! Already pushed another commit

Comment (rfiischer) [reply pythran/cxxtypes.py]: I tried to debug the code with print statements, since my debugger crashes when I try to place a breakpoint, and the outputs make no sense. Pythran runs and generates a fully functional pythranized module out of the function I need converted, though. I honestly have no idea how I could debug better this code :( but the proposed correction works nicely on this module I am working on. I have also no idea to what triggers this behavior, is it related to the number of arguments of the function? Mine has 9, with the following signatures:
```
(float[:, :], float[:], float[:], int[:, :, :], int[:, :], int[:, :], int, int, float)
```
All the arrays are numpy.

Comment (serge-sans-paille) [reply pythran/cxxtypes.py]: The annotations look good. Without a reproducer, it's hard for me to tell... Anyway, the proposed version is much better, can you just squash the three commits in one? Thanks!

Comment (rfiischer) [reply pythran/cxxtypes.py]: I arranged a script that produces the bug. Before giving final closure, it would be nice if you could analyse and test yourself so that we can propose the best solution and really find the source of the error. And also, test if the recursion limit is being set back to the original one. 

```python
"""
Script that bugs cxxtypes.py
"""

import numpy as np

# pythran export _core_decode(float[:, :], float[:], float[:], int[:, :, :], int[:, :], int, int, float)
def _core_decode(arg0, arg1, arg2, arg3, arg4, arg5, arg6, arg7):

    var0 = np.zeros((arg5, 16), dtype=float)
    var1 = np.zeros((arg5, 16, 4), dtype=float)

    for k in range(arg5):
        var2 = [arg2[k], arg1[k]]

        for i in range(16):
            var1[k, i, 0] = np.dot(arg3[i, 0], var2)

            for j in range(1, 4):
                var1[k, i, j] = arg0[k, j - 1] + np.dot(arg3[i, j], var2)

    var3 = 0
    var4 = np.zeros(16, dtype=float)
    var5 = False
    while var3 < arg6 and not var5:
        for k in range(arg5):
            for j in range(16):
                var4[j] = np.max([var0[k, arg4[j, 0]] +
                                  var1[k, arg4[j, 0], 0],
                                  var0[k, arg4[j, 1]] +
                                  var1[k, arg4[j, 1], 1],
                                  var0[k, arg4[j, 2]] +
                                  var1[k, arg4[j, 2], 2],
                                  var0[k, arg4[j, 3]] +
                                  var1[k, arg4[j, 3], 3]])

            var4 = var4 - var4[0]

            if k == arg5 - 1:
                if np.max(np.abs(var4 - var0[0, :])) < arg7:
                    var5 = True
                    break

                var0[0, :] = var4

            else:
                if np.max(np.abs(var4 - var0[k + 1, :])) < arg7:
                    var5 = True
                    break

                var0[k + 1, :] = var4

        var3 += 1

    return var1, var0
```

When run with the original implementation on rel 0.9.5, it bugs. Replacing the `generate` function at line 227 with the one from my last commit everything works perfectly.

If you find any hints or need any kind of debugging I can make, I'm happy to help!

serge-sans-paille:
Thanks for providing a test case! I've tested current patch and it works great, if you can squash it, I'll be happy to merge it.

I've also run the following diff to check that the recursion limit doesn't change

```
-            def generate(self, ctx):
+            def generate_helper(self, ctx):
+                import sys
+                current_recursion_limit = sys.getrecursionlimit()
                 try:
                     return 'typename __combined<{}>::type'.format(
                         ','.join(ctx(t) for t in self.types))
                 except RuntimeError:
                     # this is a situation where we accept to somehow extend
                     # the recursion limit, because of degenerated trees
-                    import sys
-                    current_recursion_limit = sys.getrecursionlimit()
-                    try:
-                        return self.generate(ctx)
-                    except RuntimeError:
-                        sys.setrecursionlimit(current_recursion_limit * 2)
-                        return self.generate(ctx)
-                    finally:
-                        sys.setrecursionlimit(current_recursion_limit)
+                    sys.setrecursionlimit(current_recursion_limit * 2)
+                    res = self.generate_helper(ctx)
+                    sys.setrecursionlimit(current_recursion_limit)
+                    return res
+
+            def generate(self, ctx):
+                import sys
+                current_recursion_limit = sys.getrecursionlimit()
+                out = self.generate_helper(ctx)
+                assert sys.getrecursionlimit() == current_recursion_limit
+                return out
+
+
```

And it's indeed the case, so the patch really looks good. I'll still investigate the reason why we get that call graph, but that's another story :-)

rfiischer:
Squash done!

serge-sans-paille:
Thanks a lot for the detailed test case and the follow-ups!

PR #1461, opened 2020-02-06
Fix a.reshape(-1)

jeanlaroche:
There's a bug in the current implementation of a.reshape that causes a segmentation fault for calls like a.reshape(-1)
This PR fixes it, and adds a test.

### DIFF
--- a/pythran/pythonic/include/types/tuple.hpp
+++ b/pythran/pythonic/include/types/tuple.hpp
@@ -1115,23 +1115,28 @@ namespace sutils
   using push_front_t = concat_t<types::pshape<T>, P>;
 
   template <class S>
-  long find(S &s, long v, std::integral_constant<size_t, 0>)
+  long find(S &s, long v, std::integral_constant<size_t, 0>, long start,
+            bool comp(long, long))
   {
-    return v == std::get<0>(s) ? 0 : -1;
+    return comp(std::get<0>(s), v) && 0 < start ? 0 : -1;
   }
   template <class S, size_t I>
-  long find(S &s, long v, std::integral_constant<size_t, I>)
+  long find(S &s, long v, std::integral_constant<size_t, I>, long start,
+            bool comp(long, long))
   {
-    return v == std::get<I>(s)
+    return comp(std::get<I>(s), v) && I < start
                ? I
-               : find(s, v, std::integral_constant<size_t, I - 1>());
+               : find(s, v, std::integral_constant<size_t, I - 1>(), start,
+                      comp);
   }
 
   template <class S>
-  long find(S &s, long v)
+  long find(S &s, long v, long start = std::tuple_size<S>::value,
+            bool comp(long, long) = [](long a, long b) { return (a == b); })
   {
-    return find(
-        s, v, std::integral_constant<size_t, std::tuple_size<S>::value - 1>());
+    return find(s, v,
+                std::integral_constant<size_t, std::tuple_size<S>::value - 1>(),
+                start, comp);
   }
 
   template <class S, class B>
--- a/pythran/pythonic/numpy/ndarray/reshape.hpp
+++ b/pythran/pythonic/numpy/ndarray/reshape.hpp
@@ -6,6 +6,7 @@
 #include "pythonic/utils/functor.hpp"
 #include "pythonic/utils/numpy_conversion.hpp"
 #include "pythonic/types/ndarray.hpp"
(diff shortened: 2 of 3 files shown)


pythran/pythonic/numpy/ndarray/reshape.hpp
```diff
@@ -27,7 +28,20 @@ namespace numpy
                             types::ndarray<T, NpS>>::type
     reshape(types::ndarray<T, pS> const &expr, NpS const &new_shape)
     {
-      long where = sutils::find(new_shape, -1);
+      long where = -1;
```
serge-sans-paille: I understand why you're doing that, but would that be possible to extend `sutils::find`` so that it accepts an extra `start` argument. In that case you could just chain two calls to `sutils::find` to detect errors... You could also make `sutils::find` accept a lambda predicate instead of a value to check for any negative value. I want to avoid creating a temporary array like you do, that's not super efficient.

jeanlaroche: Yes, I don't like it either, I like your implementation of sutils::find and I was thinking about extending it to test for < ... I'll try your ideas.

jeanlaroche:
OK Done, I exteded sutils::find to accept a start value (and end value in fact), and a lambda for the comparison.

pythran/pythonic/include/types/tuple.hpp
```diff
@@ -1115,23 +1115,32 @@ namespace sutils
...
-               : find(s, v, std::integral_constant<size_t, I - 1>());
+               : find(s, v, std::integral_constant<size_t, I - 1>(), start,
+                      comp);
   }
 
   template <class S>
-  long find(S &s, long v)
+  long
+  find(S &s, long v,
+       long start = std::integral_constant<size_t, std::tuple_size<S>::value>(),
```
serge-sans-paille: shouldn't statr default to `0` ?

pythran/pythonic/include/types/tuple.hpp
```diff
@@ -1115,23 +1115,32 @@ namespace sutils
   using push_front_t = concat_t<types::pshape<T>, P>;
 
   template <class S>
-  long find(S &s, long v, std::integral_constant<size_t, 0>)
+  long find(S &s, long v, std::integral_constant<size_t, 0>, long start,
+            std::function<bool(long, long)> comp)
```
serge-sans-paille: instead of a std::function, just use a template parameter, that way we avoid the std::function overhead.

pythran/pythonic/include/types/tuple.hpp
jeanlaroche: no because you go backward in the template expansion

pythran/pythonic/numpy/ndarray/reshape.hpp
```diff
@@ -27,7 +28,18 @@ namespace numpy
                             types::ndarray<T, NpS>>::type
     reshape(types::ndarray<T, pS> const &expr, NpS const &new_shape)
     {
-      long where = sutils::find(new_shape, -1);
+      std::function<bool(long, long)> comp = [](long a, long b) {
+        return a <= b;
+      };
+      long where = sutils::find(
+          new_shape, -1,
+          std::integral_constant<size_t, std::tuple_size<NpS>::value>(), comp);
```
serge-sans-paille: why do you start from the end?

pythran/pythonic/include/types/tuple.hpp
jeanlaroche: (i.e. you start with len(tuple)-1)

serge-sans-paille: And no need to use integral_constant here.

pythran/pythonic/numpy/ndarray/reshape.hpp
jeanlaroche: That's what you did... I didn't change that part

pythran/pythonic/include/types/tuple.hpp
jeanlaroche: Actually yes, because you need the integral_constant constant to finish the expansion (you have two overlads of find(), one for I=0 and one for I != 0, this has to be known at compile time). Or at least that's what I understand. (remove integral_constant breaks compilation by the way)

serge-sans-paille:
Thanks for the efforts @jeanlaroche !

PR #1463, opened 2020-02-09
Honor shape when folding empty array

### LINKED ISSUE #1447: Not easy to return an empty array
Assume we have this:
```
import numpy as np

#pythran export test(int)
def test(n):
    if n > 10:
        return np.ones((n,n,n))
    else:
#        return np.ones((0,0,0*n)) # <- this compiles
        return np.ones((0,0,0))
```
This does not compile unless I replace 0 by 0*n. I'm assuming pythran is doing early optimization, this can be verified with the -P flag:
```
return __pythran_import_numpy.array((), __pythran_import_numpy.float64)
```
versus:
```
return __pythran_import_numpy.ones(builtins.pythran.make_shape(0, 0, (0 * n)))
``` 
Is there a way to make this work without the ugly workaround? I.e. a construct which pythran will not optimize out, to return a n-dimensional empty array?

serge-sans-paille:
Fix #1447

### DIFF
--- a/pythran/conversion.py
+++ b/pythran/conversion.py
@@ -74,13 +74,22 @@ def size_container_folding(value):
             values = [to_ast(elt) for elt in value.values()]
             return ast.Dict(keys, values)
         elif isinstance(value, np.ndarray):
-            return ast.Call(func=ast.Attribute(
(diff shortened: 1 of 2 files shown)


jeanlaroche:
Thanks!!!

PR #1472, opened 2020-02-16
Avoid redundancy withing type combiners

serge-sans-paille:
This generates much cleaner code.


serge-sans-paille:
This also makes the test case from #1452 generate very simple code.

PR #1473, opened 2020-02-18
Fix combiner for ndarray expression

### LINKED ISSUE #1287: Wrong return size when two paths can return different arrays
This is a regression from 0.9.0
The code below compiles, but does not run right.
test(20) should return a 20 element array, but it does not, it return a 2 element array (with the "correct" values in it, from np.arange() but the size is truncated to two
```
import numpy as np

#pythran export test(int)
def test(N):
    if N<10:
        return np.array([.1,.1])
    A = np.arange(N)
    return A
```

serge-sans-paille:
Fix #1287


jeanlaroche:
> Fix #1287

Thanks Serge, this one has been a pain in the butt for a long while! I'll verify!

PR #1483, opened 2020-03-04
Allow %{ext} substitution in output filename

### LINKED ISSUE #1481: Can't choose the new filename without pointing out the extension
I wanted to pass the `output_file` argument to `pythran.compile_pythranfile`, but I want to make my code compile both on Linux and Windows, so I don't know beforehand the extension of the file. It would be nice to have an option where the extension is set to automatic, and one can change only the module's name.

serge-sans-paille:
Fixes #1481

### DIFF
--- a/docs/MANUAL.rst
+++ b/docs/MANUAL.rst
@@ -383,6 +383,10 @@ the Pythran context.
 Advanced Usage
 --------------
 
+One can use ``-o <filename>`` or ``--output=<filename>`` to control the name of
+the generated file. If ``<filename>`` contains the ``%{ext}`` pattern, it is
+replaced by the extension that matches your current platform.
+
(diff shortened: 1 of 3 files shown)


serge-sans-paille:
@rfiischer can you check that this matches your need?

rfiischer:
Yes, it does! Thank you very much!

rfiischer:
Actually, I couldn't use this feature. Neither with `-o`, `--output` or using the `pythran.compile_pythranfile` method. With the `-o` option I get a compile error, with `--output` I get the file with the same name as the `.py`, and with the  `pythran.compile_pythranfile` I get the same compile error. I checked the `pythran.compile_pythranfile` method, and it seems that the `module_name` is going with the `%{ext}` modifier to internal functions.

PR #1484, opened 2020-03-07
Fix module_name inference from custom output filename

serge-sans-paille:


serge-sans-paille:
@rfiischer should work better with that patch. Can you confirm?

rfiischer:
Yes! Worked both using the command line `-o` and the `pythran.compile_pythranfile` function! Just the command `--output=<filename>` that was added on #1483 on the current documentation doesn't produce any changes on the output file name.

serge-sans-paille:
> --output=<filename> that was added on #1483 on the current documentation doesn't produce any changes on the output file name

What do you mean?

PR #1507, opened 2020-04-08
Fix/complex default parameters

serge-sans-paille:


serge-sans-paille:
cc @jeanlaroche

jeanlaroche:
Thanks! This fixes the issue I was seeing! Awesome!

PR #1511, opened 2020-04-12
Bump beniget requirements

### LINKED ISSUE #1509: Pythran outputs incorrect C++
The following code compiles and runs, but does not do the right thing!
```
#pythran export test(int, int[])
import numpy as np
def test(X, Y):
    for ii in range(0, 2):
        if Y[ii] > 2:
            print('Error, Y[ii] should be less than 2 ', Y[ii])
            X = Y[ii] #<- This is optimized away by pythran!!!!
            continue
        X = Y[ii]
    return X

# python -c "from test import *; import numpy as np;  print(test(1,np.array([1,10])))"
```
In fact, looking at the C++ code generated: you can see that X_ = Y.fast(ii); is missing before the continue.
```
      for (long  ii=0L; ii < 2L; ii += 1L)
      {
        if ((pythonic::operator_::gt(Y.fast(ii), 2L)))
        {
          pythonic::__builtin__::functor::print{}(pythonic::types::str("Error, Y[ii] should be less than 2 "), Y.fast(ii));
          continue;
        }
        X_ = Y.fast(ii);
      }
```
Running the code in python and pythran does not output the same value...
`python -c "from test import *; import numpy as np;  print(test(1,np.array([1,10])))"
`

serge-sans-paille:
Fix #1509

### DIFF
--- a/pythran/tests/test_optimizations.py
+++ b/pythran/tests/test_optimizations.py
@@ -301,6 +301,11 @@ def foo():
         ref = init
         self.check_ast(init, ref, ["pythran.optimizations.ForwardSubstitution"])
 
+    def test_forwarding1(self):
+        init = 'def f(i):\n while i:\n  if i > 3: x=1; continue\n  x=2\n return x'
+        ref = 'def f(i):\n    while i:\n        if (i > 3):\n            x = 1\n            continue\n        x = 2\n    return x'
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
@jeanlaroche can you confirm this fixes your issue?

jeanlaroche:
Yes, confirmed fixed thanks!

PR #1512, opened 2020-04-12
Fix optional argument handling in spec

### LINKED ISSUE #1508: Compilation error with fromFile
This does not compile with the latest pythran. BUT, remove the extra input Z in test() and it compiles!!!
```
import numpy as np

def B(dataFile):
    W = np.fromfile(dataFile, np.float64, 1000, '', 0).reshape((100, 10))
    self = tuple([W])
    return self

def A(modelBinFile):
    X = B(modelBinFile)
    self = tuple([X])
    return self

# pythran export test(str?,int?)
def test(fileName='data.bin', Z=0):
    TD = A(fileName)
```

serge-sans-paille:
Default argument list was split in the reverse order...
Modify a rosetta test case to use that spec syntax.

Fix #1508

### DIFF
--- a/pythran/spec.py
+++ b/pythran/spec.py
@@ -286,8 +286,7 @@ class SpecParser(object):
         elif len(p) == 4:
             p[0] = tuple((t,) + ts for t in p[1] for ts in p[3])
         else:
-            non_defaults = [t for t in p[4] if len(t) == len(p[4][0])]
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
@jeanlaroche #1508 should be fine with that one, thanks for spotting the error *and* reporting it.

PR #1590, opened 2020-07-20
Add a compilation flag THREAD_SAFE_REF_COUNT that makes the ref count…

jeanlaroche:


serge-sans-paille:
I think you can inconditionnaly use the atomic version by default. Better be safe by design :-)

jeanlaroche:
mmm I'm not sure why the Travis build failed

serge-sans-paille:
Please add an entry in the user documentation for that macro variable, as it' something users may be facin :-) all looks good otherwise, I'll just make sure it's ok performance wise, but that's on me.

docs/MANUAL.rst
```diff
@@ -393,6 +393,14 @@ However, if unset through ``-UNDEBUG``, all ``assert`` are executed and
...
 done and may fail with a C-ish assertion.
 
+Thread safety
+-------------
+
+By default Pythran dones not generate thread-safe code: reference counting for automatic
+deletion of objects is not done atomically by default. It's possible to force pythran
+to generate thread-safe reference counting by defining the flag ``THREAD_SAFE_REF_COUNT``
+via ``-DTHREAD_SAFE_REF_COUNT``. There is a small performance penalty associated with this.
+
```
serge-sans-paille, line 403: + Note that if OpenMP is activated, Pythran automatically switches to  thread-safe reference counting.

PR #1593, opened 2020-07-21
Fix time comparison for cpp file creation

diorcety:
The current os.stat comparison is not correct

### DIFF
--- a/pythran/dist.py
+++ b/pythran/dist.py
@@ -147,7 +147,7 @@ class PythranExtension(Extension):
             output_file = base + '.cpp'  # target name
 
             if os.path.exists(source) and (not os.path.exists(output_file)
-               or os.stat(output_file) < os.stat(source)):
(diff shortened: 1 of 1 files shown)


diorcety:
`touch -d '1 Jan 2006 11:34' a && touch -m -d '1 Jan 2006 12:35' a && touch -d '1 Jan 2006 11:35' b  && touch -m -d '1 Jan 2006 12:34' b && python3.7 -c "import os; print(os.stat('a') < os.stat('b'))"`
return true
`touch -d '1 Jan 2006 11:34' a && touch -m -d '1 Jan 2006 12:35' a && touch -d '1 Jan 2006 11:35' b  && touch -m -d '1 Jan 2006 12:34' b && python3.7 -c "import os; print(os.path.getmtime('a') < os.path.getmtime('b'))"`
return false

serge-sans-paille:
LGTM, thanks!

PR #1604, opened 2020-07-24
Change pdqsort.h to pdqsort.hpp

jeanlaroche:


serge-sans-paille:
Great! you now need to run clang-format-3.6 -i on pdqsort.hpp, or skip it in ``pythran/tests/test_format.py:19``. Can you do that?

jeanlaroche:
Oh, I see, I imagine it was skipped before but since the name has changed it's not any longer. I'll get it to be skipped, better than changing the code.

serge-sans-paille:
>  I'll get it to be skipped, better than changing the code.

Agreed, thanks!

PR #1621, opened 2020-08-06
builtins::bin: avoid use of std::osstream

lsix:
The final length of the string representation can be easily computed
head of time so a stream string is a bit overkill. Prefer a regular
string with storage reserved in advance.

[Quickbench](https://quick-bench.com/q/teJB_p3MpMNwPDqXNnCaPqVng6w) also says this is more efficient (but to be honest, this code will never end up in a critical compute intensive path).

### DIFF
--- a/pythran/pythonic/builtins/bin.hpp
+++ b/pythran/pythonic/builtins/bin.hpp
@@ -6,29 +6,39 @@
 #include "pythonic/types/str.hpp"
 #include "pythonic/utils/functor.hpp"
 
-#include <sstream>
+#include <algorithm>
+#include <cmath>
 
 PYTHONIC_NS_BEGIN
 
 namespace builtins
 {
   template <class T>
-  types::str bin(T const &v)
+  typename std::enable_if<std::is_scalar<T>::value, types::str>::type
+  bin(T const &v)
   {
-    long unsigned int i = 1L << (8 * sizeof(T) - 1);
-    while (i && !(v & i))
-      i >>= 1;
-    if (!i)
+    using UT = typename std::make_unsigned<T>::type;
+    if (v == T{0})
       return "0b0";
     else {
-      std::ostringstream oss;
-      oss << "0b";
+      // Due to rounding errors, we cannot use std::log2(v)
+      // to accuratly find length.
+      size_t len = (8 * sizeof(UT)) - 1;
+      UT i{UT{1} << len};
+      while (!(i & v)) {
+        i >>= 1;
+        len--;
+      }
+      types::str res;
+      res.reserve(2 + len);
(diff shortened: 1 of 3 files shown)


lsix:
Wait, this does not work for negative numbers (but the original code did not either).

But to be honest, the reference python behavior is somehow strange:

```python
>>> bin(-8)
'-0b1000'
```

I’ll fix it tomorrow, so we actually mimic the python behavior.

basically:

```python
def strangebin(a):
    if a < 0:
        return "-" + bin(-a)
    else:
        return bin(a)
```

pythran/pythonic/builtins/bin.hpp
```diff
@@ -6,29 +6,47 @@
...
+  bin(T const &v)
   {
-    long unsigned int i = 1L << (8 * sizeof(T) - 1);
-    while (i && !(v & i))
-      i >>= 1;
-    if (!i)
+    using UT = typename std::make_unsigned<T>::type;
+    if (v < T{0})
+      if (v == std::numeric_limits<T>::min()) {
+        // In this special case, calling -v would overflow so
```
serge-sans-paille, line 23: it would be great to have a test case for that path.

pythran/pythonic/builtins/bin.hpp
```diff
@@ -6,29 +6,47 @@
...
+        // a special case is needed.
+        types::str res;
+        res.reserve(8 * sizeof(T) + 3);
+        auto &backend = res.get_data();
+        backend.append("-0b1");
+        for (size_t i = 1; i < 8 * sizeof(T); i++) {
+          backend.append("0");
```
serge-sans-paille, line 30: btw this could be an `std::fill`

serge-sans-paille:
You'll need to rebase on master to cope with the new gast version.

pythran/pythonic/builtins/bin.hpp
lsix: I have added one, using `sys.maxsize,` however it causes tests to fail on windows.

During the tests, the parameter it treated as a `long`. And long range is different on windows:  it goes from  -2147483648 to 2147483647, when it goes from -9223372036854775808 to 9223372036854775807 on my linux system. Basically, to have consistent  behavior, `long long` should be used.

Basically as it is, this test case (`test_bin3`) should be platform dependent. I guess you have had this kind of question raised in the past. What is your take on it?

lsix: I have pushed a test (`pythran.tests.test_base.test_3`) for it, but disabled for windows.

serge-sans-paille:
Thanks o/

PR #1622, opened 2020-08-08
Feature/py39 support

serge-sans-paille:


serge-sans-paille:
python 3.9 support coming! The ast as changed a bit, so has gast, let's adapt to these changes.

PR #1623, opened 2020-08-10
types::str: remove get_data since it duplicates chars

lsix:
`types::str::get_data` and `types::str::chars` basically does the same thing. Remove `types::str::get_data` and
only use chars in order to avoid duplicated API.


serge-sans-paille:
Thanks o/

PR #1629, opened 2020-09-01
Fix pythran-config with cl.exe or clang-cl.exe

diorcety:
The arguments returned by pythran-config are incorrect if the compiler is not gcc (cl.exe or even clang-cl.exe).
Using ccompiler methods for getting valid arguments


serge-sans-paille:
LGTM, thanks.

The travis error is unrelated, I'll fix it in another commit.

PR #1631, opened 2020-09-05
Use range values to fasten slice computation

### LINKED ISSUE #1603: Avoiding wraparound test using assert()
I was reading https://serge-sans-paille.github.io/pythran-stories/basic-value-range-analysis.html and trying to find a way to avoid the wraparound code in an expression like this:
`x[a:b] * y[a:b]`
where we can know for sure that 'a>=0'
I was assuming that doing:
```
assert a>=0
x[a:b] * y[a:b]
```
would cause pythran to remove the wraparound code and create a tighter loop.
I generate the C++ code with -e but I can see no difference whether I add `assert a>=0` or not (other than the presence of the assert() call in the C++

serge-sans-paille:
This introduces a new type of slices, fast_contiguous_slice, with faster
methods. A direct consequence is the rewrite of most slicing methods to
accommodate this new type.

Fix #1603

### DIFF
--- a/pythran/backend.py
+++ b/pythran/backend.py
@@ -1007,8 +1007,11 @@ class CxxFunction(ast.NodeVisitor):
             args.append(arg)
 
         if node.step is None or (isnum(node.step) and node.step.value == 1):
-            return "pythonic::types::contiguous_slice({},{})".format(args[0],
-                                                                     args[1])
+            if self.all_positive(node.lower) and self.all_positive(node.upper):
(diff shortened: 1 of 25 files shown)


serge-sans-paille:
cc @jeanlaroche I haven't tested the performance impact yet!

jeanlaroche:
Dude! I can't wait to try it! Let me take a look and try to understand what you did.

serge-sans-paille:
I have a better implementation locally, please wait before testing :-)

PR #1635, opened 2020-09-14
Travis CI: Test on Python 3.9

hroncok:


hroncok:
This was supposed to be opened inside my fork, so I can test things. Sorry fo the noise.

PR #1641, opened 2020-09-21
Handle typing of rvalues in a decent way

### LINKED ISSUE #1572: Problem still exist with resizing an array previously declared in a tuple.
Pythran still misbehaves with the code below:
self[0] is initializes as a (1,1) array, then when an attempt is made to give it a shape of (100,100) this fails.
```
# pythran export test()
import numpy as np

def A():
    self_xx = np.zeros((1,1))
    self = tuple([self_xx])
    return self

def B(self, D, C):
    print(D, C)
    self[0] = np.zeros((D, C))
    print('self[0]', self[0].shape)

def test():
    fr = A()
    B(fr, 100, 100)
```

serge-sans-paille:
Introduce an explicit cast on rvalues to deal with combiner applied on
non-reference nodes.

Fix #1572

### DIFF
--- a/pythran/backend.py
+++ b/pythran/backend.py
@@ -916,7 +916,19 @@ class CxxFunction(ast.NodeVisitor):
         return " and ".join(op(x, y) for x, op, y in all_cmps)
 
     def visit_Call(self, node):
-        args = [self.visit(n) for n in node.args]
+        args = []
+        args_cast = []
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
I won't land that one in release 0.9.7, but it's worth testing it, cc @jeanlaroche

PR #1652, opened 2020-10-16
Fix inconsistency in TupleInterval Handling

serge-sans-paille:
When performing an union, intersect or add, a TupleInterval must return a tuple
interval, even of unknown range

### DIFF
--- a/pythran/interval.py
+++ b/pythran/interval.py
@@ -386,13 +386,13 @@ class IntervalTuple(object):
 
     def union(self, other):
         if isinstance(other, Interval):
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
@jeanlaroche this is a new fix for the bug we've been discussing by email. Please confirm it works correctly for you!

jeanlaroche:
I confirm that pythran no longer crashes on my file, and that the resulting C++ code compiles! Thanks a bunch!

serge-sans-paille:
Great!

PR #1653, opened 2020-10-23
Nanmin nanmax

ChristopheCoue:
numpy.nanmin and numpy.nanmax do not respect Numpy specification : "When all-NaN slices are encountered a RuntimeWarning is raised and Nan is returned for that slice."
https://numpy.org/doc/stable/reference/generated/numpy.nanmin.html

It seems there is also a problem with all-PINF slices but not checked in details

### DIFF
--- a/pythran/pythonic/numpy/nanmax.hpp
+++ b/pythran/pythonic/numpy/nanmax.hpp
@@ -15,27 +15,36 @@ namespace numpy
   namespace
   {
     template <class E, class F>
-    void _nanmax(E begin, E end, F &max, utils::int_<1>)
+    bool _nanmax(E begin, E end, F &max, utils::int_<1>)
     {
+      bool found = false;
       for (; begin != end; ++begin) {
         auto curr = *begin;
-        if (!functor::isnan()(curr) && curr > max)
+        if (!functor::isnan()(curr) && curr >= max) {
           max = curr;
+	  found = true;
+	}
       }
+      return found;
     }
     template <class E, class F, size_t N>
-    void _nanmax(E begin, E end, F &max, utils::int_<N>)
+    bool _nanmax(E begin, E end, F &max, utils::int_<N>)
     {
+      bool found = false;
       for (; begin != end; ++begin)
-        _nanmax((*begin).begin(), (*begin).end(), max, utils::int_<N - 1>());
+        found |= _nanmax((*begin).begin(), (*begin).end(), max, utils::int_<N - 1>());
+      return found;
     }
   }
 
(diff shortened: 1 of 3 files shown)


pythran/pythonic/numpy/nanmax.hpp
```diff
@@ -15,27 +15,38 @@ namespace numpy
...
   }
 
   template <class E>
   typename E::dtype nanmax(E const &expr)
   {
+    bool found = false;
     typename E::dtype max = std::numeric_limits<typename E::dtype>::lowest();
-    _nanmax(expr.begin(), expr.end(), max, utils::int_<E::value>());
+    found = _nanmax(expr.begin(), expr.end(), max, utils::int_<E::value>());
+    if (!found)
```
serge-sans-paille, line 48: Let me propose an alternative implementation: set `max` to `nan`, do not track `found` and return `max` in the end. For that trick to work, one need to change the test

```
if (!functor::isnan()(curr) && curr > max)
```
to

```
if (!functor::isnan()(curr) && !(curr <= max))
```

it makes the code cleaner, and probably faster, based on two limit cases:
- https://quick-bench.com/q/zJrmiX0h3SsJERLQ9l97lHLMy-g (no NaN)
- https://quick-bench.com/q/AYPj5AXwjBT_kUxgs43ZgK-XvLw (all NaN)

serge-sans-paille: Oh and by the way: thanks for working on this, and for providing test cases, clang-formating the code etc. that's much appreciated ;-)

serge-sans-paille:
up?

pythran/pythonic/numpy/nanmax.hpp
ChristopheCoue: ok for me. I thought it was costly to compare with nan but it seems it is not the case any more

serge-sans-paille: I was surprised too :-)

serge-sans-paille:
@ChristopheCoue do you plan to update that PR, or do you want me to finish it? I'm fine with both ;-)

ChristopheCoue:
I am not rellay used to your way of working, so maybe it is better if you finish it

serge-sans-paille:
Some good news! I've been doing some extra benchmark with your and my implementation, using the pythran implementation and not a micro benchmark, and yours is always faster. Merging your contribution as is :-)

ChristopheCoue:
ok, great. Thanks for your work

PR #1656, opened 2020-10-26
variant_functor: Correctly destroy constructed objects

diorcety:


pythran/pythonic/types/variant_functor.hpp
```diff
@@ -107,6 +107,13 @@ namespace types
       fun = new (mem) Type(other);
     }
 
+    template <class Type>
+    variant_functor_impl<Type>::~variant_functor_impl<Type>()
+    {
+      if (fun)
+        fun->~Type();
```
serge-sans-paille, line 114: nice catch! Maybe you should st fun to `nullptr` just for easier debugging?

pythran/pythonic/types/variant_functor.hpp
```diff
@@ -107,6 +107,13 @@ namespace types
       fun = new (mem) Type(other);
```
serge-sans-paille, line 107: one should also check if `mem` is not `nullptr` here.

serge-sans-paille: Same around line 57

serge-sans-paille:
Thanks!

Issue #107, opened 2013-06-08
Use annotations instead or with the current system to parse the function's signature


thedrow:
Python 3.x has the [annotations](http://www.python.org/dev/peps/pep-3107/) feature.
If you can support it instead of comments for python 3.x it will make the code much more readable.

diorcety:
It can be usefull now that pythran supports Python 3

paugier:
Note that with [Transonic](https://bitbucket.org/fluiddyn/transonic/), it's already possible to use type annotations to use Pythran.

You can put anything in python type annotations, in particular, types defined with `typing` or `numpy` or even strings (so you can use something like `"float[:, :] or int[:, : ,:]"`, which works with Transonic).

However, there is only one "signature" with type annotations, so to support the equivalent of multi-signatures in Pythran (which is really useful), we need a type system more powerful than what exists today in Python. I think we would need something like

```
from pythran import Array, Union

U = Union[float, Array[float, "3d"]]

def my_func(arg: U):
    return arg**2
```

But this is bad because then the Pythran file cannot be run without Pythran.

it is much better to use `from transonic import Array, Union`, because transonic is a much smaller (pure Python) run-time dependency.

serge-sans-paille:
See #2251 for python annotation support in local variables.

PR #1665, opened 2020-11-09
[107] Use Python 3 type annotations

seertaak:
**NB. This is a WIP.** I am creating this PR to get feedback relatively early.

[Issue 107](https://github.com/serge-sans-paille/pythran/issues/107) calls for the use of PEP 484 _type annotations_ in lieu of the Pythran's comment directives as a means to convey function signature to the Pythran compiler. This PR offers an experimental implementation of these.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ docs/AUTHORS.rst
 docs/SUPPORT.rst
 
 docs/_build/*
+.venv/
+.idea/
--- a/pythran/analyses/__init__.py
+++ b/pythran/analyses/__init__.py
@@ -18,6 +18,7 @@ from .cfg import CFG
 from .constant_expressions import ConstantExpressions
 from .dependencies import Dependencies
 from .extended_syntax_check import ExtendedSyntaxCheck
+from .extract_function_type_annotations import ExtractFunctionTypeAnnotations
 from .fixed_size_list import FixedSizeList
 from .global_declarations import GlobalDeclarations
 from .global_effects import GlobalEffects
--- a/pythran/analyses/extract_function_type_annotations.py
+++ b/pythran/analyses/extract_function_type_annotations.py
@@ -1,0 +1,98 @@
+# Copyright whatever. I don't care.
+
+from pythran.passmanager import ModuleAnalysis
+from pythran.tables import MODULES
+from pythran.spec import Spec
+from pythran.typing import List, Set, Tuple, NDArray
+from pythran.log import logger as log
+from pythran import types
+from pythran import spec
+from astpretty import pprint
+from typing import ClassVar
+
+import gast as ast
+
+import numpy as np
+
+dtypes = {
+    'bool': np.bool,
+    'byte': np.byte,
+    'complex': np.complex,
+    'int': np.int,
+    'float': np.float,
+    'uint8': np.uint8,
+    'uint16': np.uint16,
+    'uint32': np.uint32,
+    'uint64': np.uint64,
+    'uintc': np.uintc,
+    'uintp': np.uintp,
+    'int8': np.int8,
+    'int16': np.int16,
+    'int32': np.int32,
+    'int64': np.int64,
+    'intc': np.intc,
+    'intp': np.intp,
+    'float32': np.float32,
+    'float64': np.float64,
+    'float128': np.float128,
+    'complex64': np.complex64,
+    'complex128': np.complex128,
+    'complex256': np.complex256,
+    'str': str,
+    'None': type(None),
+    'bytes': bytes,
+}
+
+
+def type_annot_to_spec(node: ast.AST):
+    if isinstance(node, ast.Name):
+        return dtypes[node.id]
+    elif isinstance(node, ast.Slice):
+        start = node.lower.value if node.lower else 0
+        stop = node.upper.value if node.upper else -1
+        step = node.step.value if node.step else 1
+        s = slice(start, stop, step)
+        return s
+    elif isinstance(node, ast.Constant):
+        return node.value
+    elif isinstance(node, ast.Tuple):
+        return tuple(type_annot_to_spec(n) for n in node.elts)
+    elif isinstance(node, ast.Subscript):
+        if isinstance(node.value, ast.Name):
+            tfm = dict(list=List, set=Set, tuple=Tuple, ndarray=NDArray)
+            if node.value.id not in tfm:
+                raise Exception(f"Instantiation of undefined template '{node.value.id}'.")
+            container = tfm[node.value.id]
+            args = type_annot_to_spec(node.slice)
+            if container is NDArray:
+                elt_ty = args[0]
+                if not isinstance(elt_ty, type):
+                    raise Exception(f"Invalid type argument to template type ndarray; first argument is expected to be a type, got: {elt_ty} of type {type(elt_ty)}.")
+                new_args = [elt_ty]
+                for arg in args[1:]:
+                    if isinstance(arg, int):
+                        new_args.append(slice(0, arg, 1))
+                    elif isinstance(arg, slice):
+                        new_args.append(arg)
+                    else:
+                        raise Exception(f"Invalid type argument to template type ndarray; expected slice or int literal, got: {arg.__name__} of type {type(arg)}.")
+                args = tuple(new_args)
+            return container(args)
+        else:
+            raise Exception(f"Expected template type name, got '{node.value}'.")
+    return node
+
+
+class ExtractFunctionTypeAnnotations(ModuleAnalysis):
+    def __init__(self: 'ExtractFunctionTypeAnnotations'):
+        self.result = dict()
+        self.update = False
+        self.functions = set()
+        ModuleAnalysis.__init__(self)
+
+    def run(self, node):
+        super(ModuleAnalysis, self).run(node)
+        return spec.Spec(functions=self.result)
+
+    def visit_FunctionDef(self: 'ExtractFunctionTypeAnnotations', fn: ast.FunctionDef) -> None:
+        self.result[fn.name] = (tuple(type_annot_to_spec(arg.annotation) for arg in fn.args.args),)
--- a/pythran/analyses/extract_function_type_annotations.py
+++ b/pythran/analyses/extract_function_type_annotations.py
@@ -7,7 +7,6 @@ from pythran.typing import List, Set, Tuple, NDArray
 from pythran.log import logger as log
 from pythran import types
 from pythran import spec
-from astpretty import pprint
 from typing import ClassVar
 
 import gast as ast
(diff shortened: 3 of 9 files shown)


serge-sans-paille:
Thanks for triggering that discussion. I think that @paugier will have an opinion on this.
Some early thoughts: doe a function signature implies a pythran export? how would we handle overloads? It would be nice to have a few test cases too.

paugier:
First naive questions: @seertaak did you have a look at [Transonic](https://transonic.readthedocs.io) and how one can use Pythran with Python 3 type annotations with it?

With your proposal, how would you define array types? In strings as in  `def fun(a: "float[]")`?

It would be nice to present examples of what your proposal improves and how things would be written.

A comment: multi-signatures are a bit more precise than annotations as soon as one uses more than one overloads. See for example the difference between:
```
export fun(int, int)
export fun(float, float)
```
and 
```
export fun(int, int)
export fun(float, float)
export fun(int, float)
export fun(float, int)
```
In this case, it seems anecdotal but when there are more arguments, it can make a huge difference in term of compilation time.  One needs to be clear how annotations are converted to signatures.

To express the former in Transonic one can write
```
T = Union[int, float]
def fun(a: T, b: T): ...
```
In contrast, 
```
def fun(a: Union[int, float], b: Union[int, float]): ...
# or
def fun(a: "int or float", b: "int or float"): ...
```
should correspond to the later.

seertaak:
## PR Goals

1. Try to implement annotations "conservatively", with minimum disruption. Concretely, pythran directives are still supported; PEP 484 annotations are optional.
2. If a function is annotated but has no pythran directive, it is **not exported**. Just like before.
3. If a function is annotated _and_ has a pythran directive, verify that the argument types match (throw error if not).
4. If a function _foo_ is annotated _and_ has "empty" pythran directive (e.g. `#pythran export foo`), the **export the function** using signature gleaned from the type annotations.

## 30,000ft View

1. Right at the the start of processing - close to where the specs are parsed, we use a custom analysis which extracts annotations from FunctionDecl AST objects, and converts them into the spec format.
2. Next, we attempt to reconcile the specs obtained through comment directives and those obtained through type annotations, as described above.
3. In the middle end, a transformation pass is added to remove all type annotations. This prevents the constant folding pass from failing (it recursively traverses the type annotations and fails to find the names associated with the meta types).

## Support/Limitations

- no support for overloading currently. However, it should be straightforward to implement now that I understand the model is basically C++'s model. (I thought it was HM under the hood which doesn't support overloads IIUC.) To fix this I would need to amend the analysis step (1) above to maintain _overload sets_ rather than single function signatures. The reconciliation would then compare both sets of signatures rather than looking at individual ones. 
- currently the return type annotation (`-> int`) is ignored. We should verify that the type matches that outputted by HM.
- array isn't supported, currently. I wasn't sure if that was a c array or `std::array` or something different, this is why I didn't implement it. It should be straightforward to implement.
- no test cases; these are definitely needed. **If you have thoughts on how to create really _evil_ input, let me know.**

### Misuse: checking

```py
#pythran export foo(str)
def foo(xs: str):
    return xs
# Error produced: "E: Type annotations don't match pythran spec directive for 'foo': expected ((<class 'int'>,),); got ((<class 'str'>,),)."

#pythran export bar
def bar(xs: ndarray[int, :, uint32]):
    return xs
# Error produced:  (Still needs some work e.g. replacing list for uint32.)
# Exception: Invalid type argument to template type ndarray; expected slice or int literal, got: uint32 of type <class 'type'>.                                                                                      
```

### Primitive Types 

Every primitive type except float128 and complex256. For those, I got an undefined error in numpy; not sure why.
e.g.
```py

def times_two(x: int):
    return 2*x

def twice(x: str):
    return 2*x

def blah(z: float32):
    return 2.0*x
```

### Compound Types

Here's what's supported:
```
import numpy as np

#pythran export foo
def foo(xs: list[int]) -> list[int]:
    return [2*x for x in xs]

#pythran export bar
def bar(strs: set[str]) -> bool:
    return 'hello' in strs

#pythran export baz
def baz(x: tuple[int, str]) -> str:
    return str(x[0]) + x[1]

#pythran export bingo
def bingo(M: ndarray[float32, :, 2]) -> ndarray[float32, :]:
    print(len(M))
    x = np.zeros(len(M))
    for i in range(len(M)):
        x[i] = sum(m_i for m_i in M[i, :])
    return x
```


## Design Questions

1. _Should the `#pythran export...` directive still be required?_ I have some code which removes the need for the directive altogether. When I shared my PR I decided to be conservative; I thought there may be a preference for the directives to stay, and I didn't want to rock the boat too much. My personal preference is for the directives to be optional. I would like two annotations: `@private` and `@export`, and I would prefer if the default was to _export_ rather than make the function private - just like python. The compiler should complain when type annotations are missing on exported functions.
2. How should overloading be handled? We should use python's [overload annotation](https://docs.python.org/3/library/typing.html#typing.overload).
3. The introduction of [the Annotated type](https://docs.python.org/3/library/typing.html#typing.Annotated) in python 3.9 means that going forward generic type parameters can be values. This is of obvious relevance in the context of generating C++. (`std::aray` or other fixed-size containers). My view is that we should avoid inventing new syntax at the parser level. Any syntax Pythran uses should be compatible with plain Python, and should _feel_ like plain python. I believe my design follows this idea. But I should stress that I'm not particularly wedded to the syntax I'm using!
4. I think some discussion is perhaps warranted on the type system front. Pythran is an interesting hybrid betweem more a more ML approach and of course C++. I'm not sure I have the correct intuition for the type system yet, I'm afraid.
5. As an aside, I think the project may benefit from a more unified approach to error handling. It would be nice if a pass could output an error and automatically the location/file etc would be printed. I tried to do my best here, but I wasn't quite sure what exception to throw, and didn't want to get too bogged down.
6. Final aside, the PR is currently a bit of a mess due to a mess-up on git. I'll fix that.

paugier:
About the difference with what we can currently do with Transonic-Pythran (I'm just curious) ?

seertaak:
First, transonic looks really interesting, congrats on that project. It goes without saying that I find pythran quite delightful.

> About the difference with what we can currently do with Transonic-Pythran (I'm just curious) ?

Apart from C (?) arrays, my implementation supports the gamut of types supported by the spec syntax itself. Hopefully this answers your question. Or maybe you wonder if the typing is really recursive (can you make a list of list? _Yes_).

serge-sans-paille:
Thanks for the detailed answers!

One quick note before I dive more in the details:

> If a function foo is annotated and has "empty" pythran directive (e.g. #pythran export foo), the export the function using signature gleaned from the type annotations.

```
 #pythran export foo
```

already means you're exporting a global variable named `foo`. To avoid conflict with existing notation, I suggest

```
 #pythran export foo(...)
```

I'm obviously open to alternative :+1: 

And again, thanks for triggering this delightful discussion ;-)

serge-sans-paille:
Thanks for the detailed answers!

One quick note before I dive more in the details:

> If a function foo is annotated and has "empty" pythran directive (e.g. #pythran export foo), the export the function using signature gleaned from the type annotations.

```
 #pythran export foo
```

already means you're exporting a global variable named `foo`. To avoid conflict with existing notation, I suggest

```
 #pythran export foo(...)
```

I'm obviously open to alternative :+1: 

And again, thanks for triggering this delightful discussion ;-)

seertaak:
> To avoid conflict with existing notation, I suggest
> ```
>  #pythran export foo(...)
> ```

Thanks for pointing this out. I'm happy to use your proposed syntax.

Are you ok with the use of `@overload` annotation? Even if we don't use other annotations (like `@export`), I will need some way to signal, syntactically, that a function is overloaded. I think annotations are the most natural approach, but I'm open to other suggestions.

To manage expectations, the next slab of time I'll have will be this weekend.

serge-sans-paille:
I'm ok with `@overload` but that can fit in another PR to keep this review focused and simple. Some more thoughts:

```python
def bingo(M: ndarray[float32, :, 2]) -> ndarray[float32, :]:
```

without the proper imports, this line is no longer valid, you'll get unbound identifier for `ndarray`, `float32` etc. you should probably use directly the value from `pythran.typing.*`. Plus that may simplify part of your code.

paugier:
> you should probably use directly the value from pythran.typing.*

But then you need Pythran to run the code even when it's not compiled, isn't it?

serge-sans-paille:
yeah :-/

seertaak:
## Questions About Implementation

@serge-sans-paille if I understand correctly, what you would like is that:

1. Instead of parsing the type annotations, the user is required to add a `from pythran.typing import *` (or similar). 
2. Next, the code that reads the type annotations looks for top-level imports, and evaluates the type annotation AST in an environment that recreates the module imports.
3. Next, those types are used to generate the specs used for the rest of compilation.

Please correct me where I'm wrong :smile: I actually thought of this approach myself, but had already begun work on the parsing method I used, and wanted not to get bogged down. However, what I like about the approach I think you suggest is that it keeps you honest: by evaluating the type annotation using Python itself, you ensure that the module can be parsed and evaluated using only Python (and of course the Pythran module it refers to - see next point!).

## Type Annotations Implies Pythran Dependency

> But then you need Pythran to run the code even when it's not compiled, isn't it?

You would need to import the relevant Pythran module, yes, but the alternative embodied by my current changes is that you need to _compile the module using Pythran_. If you really don't want to require Pythran at all, then there are two options:

1. For types like `list` and `set` you could get away with using Python's own generic types. The semantics of Pythran would be that these Python lists are converted to pythran lists, etc. This obviously wouldn't work for NDArray, which is probably the type that most of Pythran's users want in the first place.
2. You drop type annotations and proceed with the older comment directives.

My opinion is that the import to Pythran's typing module is not too onerous.

## Rules Surrounding Pythran Comment Directives When Type Annotations Exist

It would be great if you could clarify the following:

1. For an overload set of size _n_, how many Pythran comment directives should be required? _n_? Or _k_ <= _n_? Or 1?
2. Should the user be allowed to specify a subset of comment directives fully (i.e. with all the argument types)? Or should all the directives be either fully specified, or `(...)`? Obviously, if they are all `(...)` then you will have _n_ identical export directives. I'm not saying that's bad, I'm just clarifying that that's what the input will be.
3. Currently proximity between pythran comment directive and the function it relates to is not checked. I'm worried about the potential for confusion; for example, assume that the user fully specifies in a comment directive one overload, but mistakenly attaches it to _another_ overload - or somewhere completely different. Should the compiler warn or fail in this case?

For now those are all my questions. Going forward it would make a lot of sense for type annotations to permit _return types_ to be specified, but that's for another day.

serge-sans-paille:
Hey @steertaak, what's the status of this PR? Looks like it's in pause, which is fine with me - I mean, there's no hurry, just wanted to get some news ;-)

ashwinvis:
With NumPy v1.20, there is now an official way of annotating types:

https://numpy.org/devdocs/release/1.20.0-notes.html#numpy-is-now-typed

paugier:
I'm a bit skeptical about using `numpy.typing` for Pythran (or other Numpy accelerators). `ArrayLike` is an union of many different types from which an array can be created (any sequences, any objects implementing the `__array__` protocol, etc.).

To use type annotations for Pythran and other Numpy accelerators, one needs much precise concrete types, and also fused types. See what is doable with `transonic.typing` (https://transonic.readthedocs.io/en/latest/generated/transonic.typing.html#transonic.typing.Array).

PR #1673, opened 2020-11-25
Fix bug in unwrap function

jeanlaroche:
The formula for unwrapping is y += 2*pi * round((x-y)/(2*pi)) (where x is the previous value)
The current formula was y += 2*pi * int((x-y)/(pi)) which is bad on two counts (pi instead of 2pi and not using round)


serge-sans-paille:
Thanks for your continuous usage, and debugging, of the numpy functions ;-)

PR #1679, opened 2020-12-04
Support static size value as numpy.ones/empty/zeros parameters

### LINKED ISSUE #1677: Initialization without int() causes compilation error
Serge, I'm sure you'll say that I'm using pythran in an un-orthodox manner but I keep running into this issue when using pythran to generate C++.
See the example below. I get a compilation error with this trying to run pythran on it.
Things work if I use int(30) instead of 30.
I'm sure you're going to tell me that `self[3] = -100.0 * np.ones(self[0])` wouldn't work in python since self is a tuple, but it does work when I use the generated code in C++.
I'd like to understand what's going on here (i.e., why using int(30) makes the entire thing compile) as this is something I tend to run into on a regular basis.

```
#pythran export test()
import numpy as np


def A___init__(sRate):
    self_N = 30 # <- int(30) makes the compilation work
    self_bufferSize = 200
    self_Y = np.zeros((self_N, self_bufferSize))
    self_prevM = -100.0 * np.ones(self_N)
    self = tuple([self_N, self_bufferSize, self_Y, self_prevM])
    A_reset(self)
    return self

def A_reset(self):
    self[3] = -100.0 * np.ones(self[0]) # <- Where things go wrong.
    pass

def B___init__(x):
    self_AAA = A___init__(x)
    self = tuple([self_AAA])
    return self

def B_reset(self, x):
    A_reset(self[0])

def runSignal(x):
    X = B___init__(x)
    B_reset(X, x)
    print(X[0][0])


def test():
    x = 44100.0
    runSignal(x)
```

serge-sans-paille:
Should fix #1677

### DIFF
--- a/pythran/pythonic/include/numpy/empty.hpp
+++ b/pythran/pythonic/include/numpy/empty.hpp
@@ -17,6 +17,11 @@ namespace numpy
   types::ndarray<typename dtype::type, types::pshape<long>>
   empty(long size, dtype d = dtype());
 
+  template <long N, class dtype = functor::float64>
+  types::ndarray<typename dtype::type,
+                 types::pshape<std::integral_constant<long, N>>>
+  empty(std::integral_constant<long, N>, dtype d = dtype());
+
   DEFINE_FUNCTOR(pythonic::numpy, empty);
 }
 PYTHONIC_NS_END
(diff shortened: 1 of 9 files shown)


serge-sans-paille:
@jeanlaroche : can you confirm it fixes your issue ?

PR #1688, opened 2020-12-24
Fix arguments to rfft and irfft

jeanlaroche:


cycomanic:
This fails if you only specify the `axis` kwarg (I've encountered this as well).

jeanlaroche:
> 
> 
> This fails if you only specify the `axis` kwarg (I've encountered this as well).

Ah ok, I'll fix and add a test.

jeanlaroche:
I fixed the issue that arose when one of the arguments was missing.

serge-sans-paille:
Thanks!

PR #1689, opened 2020-12-26
[WIP] pocketfft integration

cycomanic:
This is still work in progress. I'd just like some feedback if there's nothing really stupid I'm doing here.

## Status

- [X] implement np.fft.fft
- [X] *n, axis, norm* kwargs working.
- [x] np.fft.ifft
- [ ] real 2 complex and complex 2 real functions
- [ ] add tests
- [ ] extend to 2D
- [ ] extend to ND
- [x] refactor for code-reuse

## Tests
- [X] simple tests for np.fft.fft are equal to results from numpy with axis, n, norm keywords
- [X] **Fixed** I see a difference to numpy when compiling with `-march=native` (no other CFLAGS) and zero-padding and axis set to a different value than -1. Need to investigate if this is also present in pocketfft. Would be good if someone else could test this

## Possible issues/questions

- I could not find a way to get the strides directly from the input array, so I generate the strides assuming that the array is contiguous, so this might cause issues.
- how to specify the number of  worker threads. Shall we follow scipy convention and specify a keyword? or alternatively just set it to 0 (which implies number of threads=number of cpus. Threading does require openmp though.


cycomanic:
Some follow-up on the `-march=native` differences. I narrowed it down to `-mavx` and `-mavx2` compile options. Also I compiled scipy.fft (which also uses cxx pocketfft) and it does not exhibit the same differences, so it seems to be something in the pythran code which causes this. this is a bit beyond my skill-level I think.

cycomanic:
When writing some tests I caught another bug, that I can't quite explain. Defining:
```python
import numpy as np
#pythran export testfft(complex128[][], int)
def tesffft(x, a):
    return np.fft.fft(x, axis=a)
```
fails. However if I had a `n=None` or a `norm=None` to the parameters it succeeds. On the other hand I can define a single keyword parameter using only `n` or `norm`

Issue #1690, opened 2020-12-26
Properly set filename when raising a PythranSyntaxError


serge-sans-paille:
When using pythran through the distutils extension, the `filename` field is not properly set, which makes it more difficult to debug. See https://github.com/scipy/scipy/pull/8306

serge-sans-paille:
Fixed by #1698

PR #1698, opened 2021-01-03
Improve error reporting

serge-sans-paille:
- Force file location, even when using pythran in library mode
- Improve file location for spec file


serge-sans-paille:
Related to https://github.com/scipy/scipy/issues/13308
Fixes #1690

PR #1705, opened 2021-01-06
Pocketfft complex to complex ffts

cycomanic:
Create complex to complex fft and ifft based on pocket fft.


cycomanic:
@serge-sans-paille I've split up the integration of pocketfft. This is the complex fft and ifft functions. They pass all the tests that I've come up with and are ready to merge I believe.

pythran/pythonic/include/numpy/fft/fft.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+    template <class T, class pS>
+    types::ndarray<std::complex<T>,
+            types::array<long, std::tuple_size<pS>::value>>
+    fft(types::ndarray<std::complex<T>, pS> const &a, types::none_type n , long axis =-1,
+          types::none_type norm = types::none_type{});
+
+    NUMPY_EXPR_TO_NDARRAY0_DECL(fft);
+    DEFINE_FUNCTOR(pythonic::numpy::fft, fft);
+  }
```
serge-sans-paille, line 40: this should work for non-complex floats too right?

pythran/pythonic/include/numpy/fft/ifft.hpp
```diff
@@ -0,0 +1,45 @@
...
+            types::array<long, std::tuple_size<pS>::value>>
+    ifft(types::ndarray<std::complex<T>, pS> const &a, long n, long axis,
+          types::none_type norm );
+
+    template <class T, class pS>
+    types::ndarray<std::complex<T>,
+            types::array<long, std::tuple_size<pS>::value>>
+    ifft(types::ndarray<std::complex<T>, pS> const &a, types::none_type n , long axis =-1,
+          types::none_type norm = types::none_type{});
+
```
serge-sans-paille, line 37: Same here, should work for non-complex types too, I guess

pythran/pythonic/numpy/fft/c2c.hpp
```diff
@@ -0,0 +1,122 @@
...
+    using pocketfft::stride_t;
+    using pocketfft::shape_t;
+    using ldbl_t = typename std::conditional<sizeof(long double)==sizeof(double), double, long double>::type;
+
+    template<typename T> T norm_fct(int inorm, size_t N)
+    {
+        if (inorm==0) return T(1);
+        if (inorm==2) return T(1/ldbl_t(N));
+        if (inorm==1) return T(1/sqrt(ldbl_t(N)));
+            throw std::invalid_argument("invalid value for inorm (must be 0, 1, or 2)");
```
serge-sans-paille, line 34: What kind of error is raised in Python? you should use the same here, e.g. `types::ValueError`

pythran/pythonic/numpy/fft/c2c.hpp
```diff
@@ -0,0 +1,122 @@
...
+{
+  namespace fft
+  {
+    using pocketfft::stride_t;
+    using pocketfft::shape_t;
+    using ldbl_t = typename std::conditional<sizeof(long double)==sizeof(double), double, long double>::type;
+
+    template<typename T> T norm_fct(int inorm, size_t N)
+    {
+        if (inorm==0) return T(1);
```
serge-sans-paille, line 31: I assume the cast is not needed here, because it's already part of the return type.

serge-sans-paille: Could be a `switch` btw

pythran/pythonic/numpy/fft/c2c.hpp
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    template <class T , class pS>
+    types::ndarray<std::complex<T>, types::array<long, std::tuple_size<pS>::value>>
+    c2c(types::ndarray<std::complex<T>, pS> const &in_array, long n, long axis,
+          types::str const &norm, bool forward)
+    {
+      long i;
+      auto constexpr N = std::tuple_size<pS>::value;
+      int inorm;
```
serge-sans-paille, line 67: this `inorm` variable should probably be an `enum`

pythran/pythonic/numpy/fft/c2c.hpp
```diff
@@ -0,0 +1,122 @@
...
+      if (norm == "ortho") {
+        inorm = 1;
+      } else if (!forward){
+        inorm = 2;
+      }else {
+        inorm = 0;
+      }
```
serge-sans-paille, line 74: based on that code, it can even be an `enum class` which makes sure that the exception above cannot happen.

serge-sans-paille (commented):
A first round of review!

pythran/pythonic/include/numpy/fft/fft.hpp
cycomanic: How do I convert an ndarray? I've tried using astype, but am failing.

pythran/pythonic/numpy/fft/c2c.hpp
cycomanic: Fixed, although I'm not sure I'm using enum classes correct (never used it before).

pythran/pythonic/numpy/fft/c2c.hpp
cycomanic: Fixed

pythran/pythonic/numpy/fft/c2c.hpp
cycomanic: using enum class now so removed the exception

pythran/pythonic/include/numpy/fft/ifft.hpp
cycomanic: Fixed (see above)

pythran/pythonic/include/numpy/fft/fft.hpp
cycomanic: Fixed, fft and ifft now work for non-complex inputs.

cycomanic: I needed to include quite a few overloaded functions, maybe there is a more elegant way of doing this, however I could not find one.

cycomanic: Sorry I should have said it works for floating point inputs. I need to look into what  numpy does for integer inputs (It seems it's not just a conversion to complex.), but I would argue that integer inputs is quite uncommon and can maybe put off to another merge request

cycomanic:
@serge-sans-paille I think I've addressed all your comments. I have not implemented integer inputs, but considering that this is quite uncommon anyway maybe we can put it off for now?

cycomanic:
Adding integer types was not much additional work so I've done that as well now. There are a lot of overloaded functions now however.

cycomanic:
just a note, at the moment the real input ffts are suboptimal (I'm just converting to complex instead of using r2c), I have some preliminary code already but want to include that with the r2c and c2r functions.

cycomanic:
@serge-sans-paille the recent changes also implement rfft and irfft using pocketfft. It passes all the previous (i)rfft tests for me. I've also added a lot of tests for fft and ifft (I have not changed the r(i)fft tests yet, but plan to add more tests of several things that I encountered previously. 

## What works:
- 1D fft, ifft, rfft and irfft
- n, axis and norm keywords, they can also be set to None
- floating point and integer type itputs into fft and ifft as well as irfft (for irfft I currently just convert to complex before, not sure if there is a faster way of doing this)

## Not implemented
- 2D and ND transforms
- hfft and ihfft (I think that is straightforward to add I think)

## Notable differences to numpy
- string input does not work (I don't think this would be possible with pythran even?)
- numpy converts complex64 to complex128 when doing a fft, I deviated from that and return the same as the input type (similar for the  rfft and irfft) I believe that this is better.(i actually have some code that does complex64 calculations).

## Performance
- I have not looked into this in detail, but in my tests code is about 1.5 times faster than the current irfft and rfft and slightly faster than numpy.fft (very minor) essentially on par with scipy.fft (which is also based on the cxx version of pocketfft)

cycomanic:
@jeanlaroche considering that you wrote the previous rfft implementation, could you check this against your use-case?

serge-sans-paille:
>Notable differences to numpy
>
>    string input does not work (I don't think this would be possible with pythran even?)
Correct

>    numpy converts complex64 to complex128 when doing a fft, I deviated from that and return the same as the input type (similar for the rfft and irfft) I believe that this is better.(i actually have some code that does complex64 calculations).
That's not compatible with pythran approach. We need to stick to numpy behavior here, is that workable for you?

Apart from that, thanks *a lot* for the great work. I'm going to do an extra review pass soon ;-)

cycomanic:
@serge-sans-paille I would argue that following numpy's behaviour here only has downsides (slowing down code, required casting if one wants to do single precision calculations ...) while there should not be any detrimental effects for someone who expects the previous behaviour. 

I would argue that numpy's behaviour is in fact a bug and there are at least two bug reports about this: 
numpy/numpy#6012 and numpy/numpy#17801 are open reports where it is essentially acknowledged that numpy's behaviour is not ideal, but it's too difficult/low priority to change at the moment. 

Moreover there is precedent to deviate from traditional numpy behaviour cupy deliberately deviates from numpy because of this (mentioned in numpy/numpy#17801). 

In fact numpy compiled with intel_mkl also does preserve single precision and anaconda buiild with intel_mkl (which is the default numpy in anaconda now) adopts this behaviour see   IntelPython/mkl_fft#11241 for the motivation by the intel python devs, why they should return single precision on single precision input. 

I should also mention that numpy also downcasts to double from long double so this behaviour can also lead to loss of precision.

If my arguments have not convinced you :), as an alternative I could also implement scipy.fft as this does preserve precision (it would also offer the advantage of specifying workers).

serge-sans-paille:
> @serge-sans-paille I would argue that following numpy's behaviour here only has downsides (slowing down code, required casting if one wants to do single precision calculations ...) while preserving the precision should not have any detrimental effects for someone who expects the previous behaviour.
> 
> I would argue that numpy's behaviour is a bug (or at least not ideal) and there are at least two bug reports about this:
> [numpy/numpy#6012](https://github.com/numpy/numpy/issues/6012) and [numpy/numpy#17801](https://github.com/numpy/numpy/issues/17801) are open reports where it is essentially acknowledged that numpy's behaviour is not ideal, but it's too difficult/low priority to change at the moment.
> 
> Moreover there is precedent to deviate from traditional numpy behaviour cupy deliberately deviates from numpy because of this (mentioned in [numpy/numpy#17801](https://github.com/numpy/numpy/issues/17801)).
> 
> In fact numpy compiled with intel_mkl also does preserve single precision and anaconda buiild with intel_mkl (which is the default numpy in anaconda now) adopts this behaviour see IntelPython/mkl_fft#11241 for the motivation by the intel python devs, why they should return single precision on single precision input.
> 
> I should also mention that numpy also downcasts to double from long double so this behaviour can also lead to loss of precision.

OK, you got me. Can you just document that behavior in the code? basically copy/pasting your answer at the right place would be enough :-)

> If my arguments have not convinced you :), as an alternative I could also implement scipy.fft as this does preserve precision (it would also offer the advantage of specifying workers).

Leave that for another patch set ;-)

serge-sans-paille:
Once the minor nits have been addressed, you can squash your commits in a big one, or in a few ones if the split makes sense, and we're done!

pythran/pythonic/numpy/fft/c2c.hpp
```diff
@@ -0,0 +1,287 @@
...
+        return inorm;
+    }
+
+    template<typename T> T norm_fct(Inorm inorm, size_t N)
+    {
+        switch (inorm)
+        {
+            case Inorm::ortho : return T(1/sqrt(ldbl_t(N)));
+            case Inorm::backward : return T(1/ldbl_t(N));
+            case Inorm::forward : return T(1);
```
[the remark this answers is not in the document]
cycomanic: I did first use !types::is_complex<T> but then I need to add conditionals to distinguish between integer and floating scenarios, i.e. convert the integers to double, but not floats (because we want to preserve precision).

pythran/pythonic/numpy/fft/rfft.hpp
```diff
@@ -1,145 +1,93 @@
...
-    // This is kludgy, and I'm sure there's a better way to do this. Jeanl
-    bool testThis(types::none_type param)
+    template <class T , class pS>
+    types::ndarray<typename std::enable_if<std::is_integral<T>::value, std::complex<double>>::type,
+            types::array<long, std::tuple_size<pS>::value>>
+    rfft(types::ndarray<T, pS> const &in_array, types::none_type n, long axis,
+          types::str const &norm)
     {
-      return false;
+        auto tmp_array = _copy_to_double(in_array);
```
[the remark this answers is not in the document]
cycomanic: See my comment above I think the code is more explicit (albeit a bit more code) if 
we overload for integer and floating instead of merging them and using conditionals.

serge-sans-paille:
Great! Can you now squash and rebase your commits?

cycomanic:
Yes, a little swamped at work at the moment. I'll try to get to it by the weekend.

cycomanic:
Messed up the rebase a couple of times, should be correct now. not sure why it's showing the old commits in the updates above. 

@serge-sans-paille that should be ready to merge. I will create another pull request with (i)hfft and also the scipy implementation (which allows specifying the number of worker threads).

serge-sans-paille:
\o/ thanks a lot for the hard work!

PR #1712, opened 2021-01-25
[WIP] Feature/fma copyto

serge-sans-paille:


serge-sans-paille:
Obsoleted by #1723 for ``np.cppyto``, and the fma part may not be such a good idea.

PR #1713, opened 2021-01-25
Acknowledge the fact we don't know how to widen Intervals based on I…

### LINKED ISSUE #1707: Pythran crashes on this code with `AttributeError: 'Interval' object has no attribute 'values'`
This code crashes pythran (not the compilation, pythran itself) with `AttributeError: 'Interval' object has no attribute 'values'`
MacOS. python3.8.5
```
import numpy as np


def PolyPitch_removeHarms(cand1Hz, harm2Bin,S):
    for ii in range(1, 10000):
        idx, delta = findClosest(cand1Hz, harm2Bin)
        harm1Bin = 10
        getPeaks2(S[int(harm1Bin) :int(harm1Bin) + 1]) + int(harm1Bin)

def getPeaks2(signal, minSignal=-np.inf, horiz=1):
    if minSignal is None:
        minSignal = -np.Inf
    for ii in range(horiz):
        if signal[ii] < minSignal:
            continue

def findClosest(values, item):
    idx=0
    return idx, np.abs(values[idx] - item)
```

serge-sans-paille:
…nterval

Fix #1707

### DIFF
--- a/pythran/interval.py
+++ b/pythran/interval.py
@@ -416,6 +416,8 @@ class IntervalTuple(object):
         return out or UNKNOWN_RANGE
 
     def widen(self, other):
+        if isinstance(other, Interval):
+            return UNKNOWN_TUPLE_RANGE
         return IntervalTuple(s.widen(o) for s, o in zip(self.values,
                                                         other.values))
 


jeanlaroche:
I confirm that fix/intervals-widening fixes the issue for me! THANKS!

PR #1733, opened 2021-03-06
Feature/dot different dtype

serge-sans-paille:


nbecker:
confirm this appears to work for my test case

serge-sans-paille:
Merged then :-)

PR #1742, opened 2021-03-23
Honor CFLAGS

serge-sans-paille:


pythran/config.py
```diff
@@ -167,6 +167,11 @@ def parse_define(define):
         extension['cxx'] = cxx
         extension['cc'] = cc or cxx
 
+    # Honor CFLAGS
+    cflags = os.environ.get('CFLAGS', None)
+    if cflags is not None:
+        extension['extra_compile_args'].extend(cflags.split())
```
rgommers, line 173: Shouldn't this have been `CXXFLAGS` instead?

PR #1746, opened 2021-03-26
Fix pythonic interface for numpy.finfo

angus-g:
Looks like this was still using an older `getattr` API, and wasn't being caught by the test because it gets optimised away.


serge-sans-paille:
LGTM, thanks!

PR #1749, opened 2021-03-29
Make np.finfo.eps float for complex dtype

angus-g:
Numpy's `finfo` will first cast its argument dtype to a floating instance, particularly in the case of a complex dtype. We add a specialisation to the template to facilitate this.

@serge-sans-paille I must admit I didn't look around too much for whether partial specialisation is the right pattern to use here (or whether I've done it "properly" -- my template C++ is a bit rusty!), I'm happy to use something more project-specific if that's the way to go instead.


angus-g:
Maybe this should just use `types::is_complex` instead...?

PR #1750, opened 2021-03-29
Forward .real and .imag into numpy_vexpr

angus-g:
This is implemented for most other ndarray derivatives, but not a plain view. We should be able to transform the underlying object in the same way as e.g. `numpy_iexpr`.

1. Not sure if we should necessarily be accessing `data_` and `view_` on the `numpy_vexpr` directly, but it does seem to fit the same pattern as the other expressions.
2. I guess it's only really `.real` and `.imag` that need this specialisation, but the same pattern is applied a few times -- maybe it'd benefit from an interface to change the view across `iexpr`, `gexpr`, `texpr` and `vexpr`?

### DIFF
--- a/pythran/pythonic/include/types/ndarray.hpp
+++ b/pythran/pythonic/include/types/ndarray.hpp
@@ -853,6 +853,14 @@ namespace builtins
     return {getattr(types::attr::REAL{}, a.arg)};
   }
 
+  template <class T, class F>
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
1. yeah, your pattern is fine
2. feel free to propose a refactoring, I'll gladly review it - but don't force yourself into it.

serge-sans-paille:
And thanks for spotting *and* fixing this!

PR #1752, opened 2021-04-06
Update install instructions, no more `easy_install` & use Miniforge

rgommers:
These stood out to me. `easy_install` has been removed from `setuptools` and should always be avoided even if it's still available on some system. Miniforge is better than Miniconda now, more up-to-date and more fitting because Pythran comes from conda-forge.

[ci skip]


serge-sans-paille:
100% agree, thanks!

PR #1757, opened 2021-04-13
Update installation instructions to mention mamba

SylvainCorlay:


serge-sans-paille:
README.rst.orig looks like a left over. Please keep installation notice for both conda and mamba ;-)

PR #1765, opened 2021-04-17
Handle LDFLAGS, and use CXXFLAGS instead of CFLAGS

rgommers:
This is a follow-up to gh-1742 ~and should fix https://github.com/scipy/scipy/issues/13717.~

### DIFF
--- a/.github/workflows/parallel.yml
+++ b/.github/workflows/parallel.yml
@@ -25,7 +25,11 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r requirements.txt
         pip install ipython nbval pytest-xdist cython scipy wheel
+        # libgomp comes with gcc-5, for clang we need to install libomp
         sudo apt install libopenblas-dev ${{ matrix.cpp-version }}
+        if [[ ${{ matrix.cpp-version }} == 'clang-5.0' ]]; then
+          sudo apt install libomp5 libomp-dev;
+        fi
     - name: Setup
       run: |
         python setup.py install
--- a/.github/workflows/parallel.yml
+++ b/.github/workflows/parallel.yml
@@ -41,4 +41,5 @@ jobs:
         export CC=`echo ${{ matrix.cpp-version }} | sed -e 's/g++/gcc/'`
         export CXX=`echo ${{ matrix.cpp-version }} | sed -e 's/clang/clang++/'`
         export CXXFLAGS="-fopenmp"
+        export LDFLAGS="-fopenmp"
         pytest pythran/tests/test_cases.py -v --numprocesses=1
--- a/.github/workflows/parallel.yml
+++ b/.github/workflows/parallel.yml
@@ -33,7 +33,7 @@ jobs:
     - name: Setup
       run: |
         python setup.py install
-        printf '[commpiler]\nblas=openblas\n' > ~/.pythranrc
+        printf '[compiler]\nblas=openblas\n' > ~/.pythranrc
     - name: Testing vectorized
       run: |
         export CC=`echo ${{ matrix.cpp-version }} | sed -e 's/g++/gcc/'`
--- a/docs/MANUAL.rst
+++ b/docs/MANUAL.rst
@@ -428,7 +428,7 @@ Wants to try your own compiler? Update the ``CC`` and ``CXX`` fields from your
 ``pythranrc``, or set the same environment variables to the right compilers. Environment
 variables have greater precedence than configuration file.
 
-Pythran also honors the ``CFLAGS`` environment variable.
+Pythran also honors the ``CXXFLAGS`` and ``LDFLAGS`` environment variables.
 
 The careful reader might have noticed the ``-p`` flag from the command line. It
 makes it possible to define your own optimization sequence::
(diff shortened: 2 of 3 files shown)


serge-sans-paille:
I second the `s/CFLAGS/CXXFLAGS/' change and the fact we honor `LDFLAGS` but I don't understand why `-m32` should be sued by default.

rgommers:
> but I don't understand why `-m32` should be used by default.

Let me ask, is that because you don't think it's always needed for compiling on 32-bit Windows, or because of something else? One thing I failed to notice is that there's no `pythran-win64.cfg`, which may be a problem here. I had expected this config file to be for 32-bit Windows only.

We want behavior that works out of the box on both 32-bit and 64-bit Windows, and the difference between them right now is that the latter works for SciPy and the former is missing `-m32`.

Right now config files are loaded with `'pythran-{}.cfg'.format(sys.platform)`. I'm not sure if it's better to change this, or to make a difference between 32/64-bit Windows some other way.

rgommers:
Sorry my above comment is a bit confused, I meant: 32-bit Python on Windows. CI on this repo has only one Windows job, I'll have a look at whether it's possible to add 32-bit Python. On Appveyor it's `PYTHON_ARCH: 32` (or `64`, default) that controls this. Unfortunately there are still way to many users that use 32-bit Python.

rgommers:
This should be easy with https://github.com/actions/setup-python, just using `architecture: 'x86'`. Current there's only one Windows CI job, it uses 64-bit Python 3.7. How about I add 32-bit Python 3.9 to it?

serge-sans-paille:
Oh, I understand the issue now. Pythran has only one windows config file, because it loads its config based on the output of `sys.platform`. I'll have a look at distutils internal to check why it's not adding the -m32 on its own when needed.
And +1 for the extra CI job on 32bit windows, that's probably the first thing to do.

serge-sans-paille:
NB: the openmp issue seems unrelated.

rgommers:
Updates needed:
- I'll remove the `-m32` part from this PR.
- I should also add to the docs that one can use `PYTHRAN_CXX_FLAGS` instead of `CXXFLAGS` for passing flags to Pythran only (can be useful, e.g. to use `-Wunused-function` for warning suppression).

I'm also wondering whether it's worth renaming `cflags` in the code and `.cfg` files to `cxxflags`. That'll be a bit disruptive short-term for users maybe, but it seems like the right thing to do. Maybe with a deprecation and transition period?

serge-sans-paille:
Can you also edit `.github/workflows/parallel.yml` and add an `LDFLAGS=-fopenmp` right under the `CXXFLAGS=-fopenmp` ? Looks like I behaved asif pythran was already honoring `CXXFLAGS` ... this is the reason why there's a failing test case here.

rgommers:
Good catch. It then wasn't actually running in parallel before (because `CXXFLAGS="-fopenmp"` was ignored till now), so hope it works.

Also, the actual test command is:
```
pytest pythran/tests/test_cases.py -v --numprocesses=1
```
I'm not sure that `--numprocesses=1` does anything, it's not handled in `test_cases.py` and `unittest` doesn't know that argument either I think. `pytest` does, but that's not invoked here.

rgommers:
Indeed, it actually cannot find `openmp`:
```
In file included from /tmp/tmpefqkv8i0.cpp:1:
In file included from /home/runner/work/pythran/pythran/pythran/pythonic/core.hpp:29:
In file included from /home/runner/work/pythran/pythran/pythran/pythonic/types/slice.hpp:5:
In file included from /home/runner/work/pythran/pythran/pythran/pythonic/types/NoneType.hpp:8:
In file included from /home/runner/work/pythran/pythran/pythran/pythonic/builtins/bool_.hpp:7:
In file included from /home/runner/work/pythran/pythran/pythran/pythonic/types/tuple.hpp:13:
In file included from /home/runner/work/pythran/pythran/pythran/pythonic/types/ndarray.hpp:4:
In file included from /home/runner/work/pythran/pythran/pythran/pythonic/include/types/ndarray.hpp:12:
/home/runner/work/pythran/pythran/pythran/pythonic/include/utils/broadcast_copy.hpp:7:10: fatal error: 'omp.h' file not found
#include <omp.h>
         ^~~~~~~
1 error generated.
WARNING: Compilation error, trying hard to find its origin...
WARNING: Nop, I'm going to flood you with C++ errors!
```
Nothing in the CI job is trying to install it - should be added to the `sudo apt install` line I think.

serge-sans-paille:
@rgommers probably need to add `libgomp1-dev` as a dep, indeed

rgommers:
Ubuntu makes it a little more complicated as usual. There's no `libgomp1-dev`, looks like there's `libgomp` for GCC (but it depends on GCC 8), and `libomp-dev` for `clang-5.0`.

CI for `g++5` was already passing though; so there's a spec file included in the `gcc-5` package: https://packages.ubuntu.com/bionic/amd64/gcc-5/filelist. So I think I just need to add `libomp-dev` here.

rgommers:
Okay it's not happy with that:
```

The following packages have unmet dependencies:
 libomp-dev : Depends: libomp5 (= 5.0.1-1)
E: Unable to correct problems, you have held broken packages.
```
https://packages.ubuntu.com/bionic/clang-5.0 recommends `libomp-dev`, but apparently you cannot mix `libomp5` with `gcc-5` here. Need some conditional install.

serge-sans-paille:
You did spend a lot of effort tracking my stupidly-settled CI job. I owe you one!

PR #1766, opened 2021-04-18
Setup testing -m32 on windows

serge-sans-paille:


serge-sans-paille:
cc @rgommers this confirms what I thouht: the easiest way to use pythran in multi-arch setup is to actually run a 32bit version of python. That way distutils properly links to the 32 bit version of libpython, and the appropriate flags are passed to the underlying compiler.

rgommers:
Thanks @serge-sans-paille. I'l a little confused why it works here, but not in Appveyor. The scipy-wheels setup uses the 32-bit version as well (simply specified by `PYTHON: C:\Python37` instead of `PYTHON: C:\Python37-x64`). 

From the CI build log here I can't see any compiler flags, so not sure what is different.

rgommers:
Also this PR dropped `-m32`: https://github.com/scipy/scipy/pull/13727/files. It was writing out these config values to `.pythranrc`:
```
CXX=clang-cl.exe
CC=clang-cl.exe
cflags=/std:c++14 -m32
```
All of that is now in `pythran-win32.cfg`, except `-m32`. I'll do a bit of digging in `distutils`; my impression is that passing along this flag is still flaky.

rgommers:
I think it's a MSVC vs Mingw32 issue. From SciPy's Azure config:
```
            # 32-bit build requires careful adjustments
            # until Microsoft has a switch we can use
            # directly for i686 mingw
            $env:NPY_DISTUTILS_APPEND_FLAGS = 1
            $env:CFLAGS = "-m32"
            $env:LDFLAGS = "-m32"
```
NumPy has the same thing.

I think it's fair to say that that flag then indeed does not belong in Pythran's config - if the user (or library) chooses to use Mingw32, they're responsible for putting `-m32` in `CFLAGS`.

serge-sans-paille:
> I think it's fair to say that that flag then indeed does not belong in Pythran's config - if the user (or library) chooses to use Mingw32, they're responsible for putting `-m32` in ~`CFLAGS`~ `CXXFLAGS`.

Agreed!

PR #1768, opened 2021-04-20
Improve OpenMP detection for different platforms and compilers

rgommers:
Currently the SciPy on Windows build gives warnings like:
```
clang-cl: warning: unknown argument ignored in clang-cl: '-print-file-name=libgomp.dylib' [-Wunknown-argument]
clang-cl: error: no input files
clang-cl: warning: unknown argument ignored in clang-cl: '-print-file-name=libgomp.so' [-Wunknown-argument]
clang-cl: error: no input files
```
for every Pythran invocation. This should improve that.

Also fix a str/bytes issue that resulted in invalid paths starting with `b/...` because of calling str() on a bytes object.

### DIFF
--- a/omp/__init__.py
+++ b/omp/__init__.py
@@ -2,7 +2,9 @@
 
 from ctypes.util import find_library
 from subprocess import check_output, CalledProcessError
-from numpy.distutils.misc_util import msvc_runtime_major
+from numpy.distutils.misc_util import (
+    msvc_runtime_major, get_shared_lib_extension
+)
 import ctypes
 import os
 import sys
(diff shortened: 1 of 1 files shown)


rgommers:
There's a couple of other issues/questions:

1. `omp` is called for every extension the way we use `PythranExtension` in SciPy, because of this monkey patching business:
```
    if int(os.environ.get('SCIPY_USE_PYTHRAN', 1)):
        # PythranBuildExt does *not* derive from npy_build_ext
        # Win the monkey patching race here and patch base class
        # before it's loaded by Pythran. This should be removed
        # once Pythran has proper support for base class selection.
        assert 'pythran' not in sys.modules
        import distutils.command.build_ext
        distutils_build_ext = distutils.command.build_ext.build_ext
        distutils.command.build_ext.build_ext = npy_build_ext
        try:
            from pythran.dist import PythranBuildExt as BaseBuildExt
        except ImportError:
            BaseBuildExt = npy_build_ext
        finally:
            distutils.command.build_ext.build_ext = distutils_build_ext
    else:
        BaseBuildExt = npy_build_ext
```
This also does the OpenMP check for every extension, even when Pythran isn't used. We should disable this.

2. Question: why is `omp` a separate top-level package next to `pythran`, rather than under it? Should this be moved into `pythran/openmp.py` instead?

omp/__init__.py
```diff
@@ -41,18 +43,43 @@ def init_msvc(self, ver):
             self.libomp = ctypes.CDLL(vcomp_path)
             self.version = 20
 
+    def get_libomp_names(self):
+        if cxx is None:
+            # Can't tell what compiler we're using, guessing we need libgomp
+            return 'libgomp'
```
serge-sans-paille, line 49: shouldn't that be `['libgomp']` to always return a list ?

rgommers: Yes indeed. Untested paths are dangerous .... fixed now.

serge-sans-paille:
LGTM, and thanks for the patch!

PR #1771, opened 2021-04-20
Fix combiner between nested structures

### LINKED ISSUE #1754: Re-sizing an array declared in a tuple does not work.
This code does not do what python does when compiled with pythran.

```
#pythran export test(int)
import numpy as np


def ___init__(N=44100):
    """Constructor for SilenceDetection"""
    self_N = N
    self_y = [np.zeros((N, 1)),]
    self = tuple([self_N, self_y])
    return self


def setSize(self, N):
    print('Before ', self[1][0].shape)
    print('N ', N)
    self[1][0] = np.zeros((N, 1))


def test(N):
    SD = ___init__(0)
    print('INIT', SD[1][0].shape)
    setSize(SD, N)
    print('After', SD[1][0].shape)
```
In python setSize(SD,N) resizes self[1].
In the pythran version, because __init__ is called with a value of 0, the size of self_y becomes fixed.

serge-sans-paille:
This requires a complex interplay between alias analyses, type inference and
combiners.

In `a[1][2]`, `a` is now considered as an indexable container indexed by an
integral constant, which make it possible to properly model tuple accesses;

This should fix #1754

### DIFF
--- a/pythran/analyses/aliases.py
+++ b/pythran/analyses/aliases.py
@@ -441,7 +441,8 @@ class Aliases(ModuleAnalysis):
                             continue
                     # FIXME: what if the index is a slice variable...
                     aliases.add(alias.containee)
-                elif isinstance(getattr(alias, 'ctx', None), ast.Param):
+                elif isinstance(getattr(alias, 'ctx', None), (ast.Param,
+                                                              ast.Store)):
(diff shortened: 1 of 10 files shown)


serge-sans-paille:
@jeanlaroche I've updated various part of this PR, and would appreciate your feedback, including impact on compilation time.

PR #1774, opened 2021-04-22
Fix syntax error in `omp`

stefanv:


serge-sans-paille:
Interestingly that's still valid Python, just not what we need here ;-) Thanks for spotting this.

stefanv:
Valid if `names` is defined ;)

PR #1783, opened 2021-05-07
Add specialization for double for interp function

jeanlaroche:
I specialized interp for numeric types. Let me know if the way I did it isn't righ.

### DIFF
--- a/pythran/pythonic/include/numpy/interp.hpp
+++ b/pythran/pythonic/include/numpy/interp.hpp
@@ -1,50 +1,95 @@
 #ifndef PYTHONIC_INCLUDE_NUMPY_INTERP_HPP
 #define PYTHONIC_INCLUDE_NUMPY_INTERP_HPP
 
+#include "pythonic/include/builtins/None.hpp"
+#include "pythonic/include/types/ndarray.hpp"
 #include "pythonic/include/utils/functor.hpp"
 #include "pythonic/include/utils/numpy_conversion.hpp"
-#include "pythonic/include/types/ndarray.hpp"
-#include "pythonic/include/builtins/None.hpp"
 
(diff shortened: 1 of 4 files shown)


pythran/pythonic/numpy/interp.hpp
```diff
@@ -1,20 +1,20 @@
-#ifndef PYTHONIC_NUMPY_INTERP_HPP
-#define PYTHONIC_NUMPY_INTERP_HPP
+#ifndef PYTHONIC_NUMPY_INTERPHPP
```
serge-sans-paille, line 1: This change is surprising

pythran/pythonic/include/numpy/interp.hpp
```diff
@@ -13,38 +13,76 @@ namespace numpy
 
   // None,None,None
   template <class T1, class T2, class T3>
-  types::ndarray<double, types::pshape<long>>
+  typename std::enable_if<!std::is_arithmetic<T1>::value,
```
serge-sans-paille, line 16: do you mean std::is_integral here?

serge-sans-paille: ok, after discussing it on IRC, this change looks good, but doesn't capture `std::complex`. Is that intended?

jeanlaroche: mmm no, I just didn't think of it. I should probably add a specialization for complex.

serge-sans-paille:
Just renamed out_type into interp_out_type, and removed the template alias redeclaration in the implementation file.

serge-sans-paille:
Great job, thanks a lot @jeanlaroche

PR #1790, opened 2021-05-23
Fix dev version string to be PEP 440 compliant

rgommers:
Using `0.9.12dev` actually broke SciPy CI, because Cython does this:
```
  pythran_is_pre_0_9_6 = tuple(map(int, pythran.__version__.split('.')[0:3])) < (0, 9, 6)
```
Add tests to check version string is compliant, taken over from NumPy

### DIFF
--- a/docs/CLI.rst
+++ b/docs/CLI.rst
@@ -29,7 +29,7 @@ The generated native ``.so`` module can then be used with the Python interpreter
 Pythran version can be dumped through ``--version``::
 
   $> pythran --version 2>&1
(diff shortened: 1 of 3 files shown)


serge-sans-paille:
thanks @rgommers. You need to update https://raw.githubusercontent.com/serge-sans-paille/pythran/master/docs/CLI.rst that references the version release too.

rgommers:
done!

serge-sans-paille:
thanks o/

PR #1806, opened 2021-06-16
Raise Memory Error upon (too) large numpy alloc

serge-sans-paille:
That's not perfect because numpy actually raises a _ArrayMemoryError that
derives from MemoryError, but that's enough for our case.


rgommers:
> That's not perfect because numpy actually raises a _ArrayMemoryError that derives from MemoryError, but that's enough for our case.

Agreed, standard `MemoryError` is expected and just fine. That's what SciPy does as well if it raises errors itself from C/Fortran code.

PR #1808, opened 2021-06-16
Expose pythran.get_include

serge-sans-paille:
This should ease integration in other toolchains

### DIFF
--- a/pythran/config.py
+++ b/pythran/config.py
@@ -13,6 +13,11 @@ import sys
 logger = logging.getLogger('pythran')
 
 
+def get_include():
+    # using / as separator as advised in the distutils doc
+    return (os.path.dirname(os.path.dirname(__file__)) or '.') + '/pythran'
+
+
 class silent(object):
     '''
     Silent sys.stderr at the system level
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
cc @rgommers

pythran/config.py
```diff
@@ -13,6 +13,11 @@
 logger = logging.getLogger('pythran')
 
 
+def get_include():
+    return os.path.join((os.path.dirname(os.path.dirname(__file__)) or '.'),
+                        'pythran')
```
rgommers, line 18: I'm not sure about what this does. I believe the include path should simply be `os.path.dirname(__file__)`? 

All includes in generated code are of the form `#include pythonic/xxx`, so we only need the directory that `pythonic/` is in right?

rgommers (commented):
Thanks @serge-sans-paille!

CI seems unhappy.

pythran/config.py
serge-sans-paille: I can't tell fore sure ^^! Looks like under some confiuration, `__file__` has a relative path which makes `direname` returns an empty path, but I can't recall which configuration leads to it. Anyway the above code is a no-op wrt previous behavior so that should be ok.

rgommers: That happens if you run it in IPython or directly as `python config.py`. If it's part of a package, it never fails to give the correct path AFAIK.

pythran/config.py
```diff
@@ -13,6 +13,11 @@
 logger = logging.getLogger('pythran')
 
 
+def get_include():
+    # using / as separator as advised in the distutils doc
+    return (os.path.dirname(os.path.dirname(__file__)) or '.') + '/pythran'
```
rgommers, line 18: Ah, the double `dirname` just strips off one directory level. And then you're adding it back with `/pythran`.

```python
>>> os.path.dirname(os.path.dirname('/Users/rgommers/Code/pythran/config.py'))
'/Users/rgommers/Code'
>>> os.path.dirname('/Users/rgommers/Code/pythran/config.py')
'/Users/rgommers/Code/pythran'
```

serge-sans-paille: exactly :-) and the `or` part is just here for the case where we have a relative path  (but when does it happen?)

PR #1814, opened 2021-06-21
Use pacman in ArchLinux

ashwinvis:
Pythran got officially packaged today, along with beniget, since it is a build time dependency for scipy now

https://archlinux.org/packages/community/any/python-pythran/


serge-sans-paille:
That's cool news!

PR #1817, opened 2021-06-23
(try to) Improve error message upon invalid spec file

serge-sans-paille:


serge-sans-paille:
@charlotte12l : this one improves the error message upon invalid spec,as discussed last monday.

PR #1823, opened 2021-06-27
Use iterator for index, fixes #1819

### LINKED ISSUE #1819: Can't index an 2d array like `a1[int, tuple]`
I tried to use pythran to improve the following function (it is largely simplified for better debugging):

```
import numpy as np


#pythran export calc_bin_more_pythran2(int, int[:], int[:])
def calc_bin_more_pythran2(Vdim, binnumbers, nbin):
    flatcount = np.bincount(binnumbers, None)
    a = np.nonzero(flatcount)

    result = np.empty([Vdim, nbin.prod()], float)
    result[0, a] = 1.0/flatcount[a]
    return result
```

And I got this error. This error message is wrong because I've tested that `np.bincount(binnumbers, None)` is right and can be compiled successfully.
```
(scipy-dev) charlotte@CHARLOTLIU-MB0 stats % pythran test_bincount.py
WARNING: Compilation error, trying hard to find its origin...
CRITICAL: You shall not pass!
E: test_bincount.py:6:16 error: Invalid argument type for function call to `Callable[[T0, none], ...]`, no overload found, tried:
Callable[[Iterable[bool], Iterable[float], int], Array[1d, int]]
Callable[[Iterable[bool], Iterable[float]], Array[1d, int]]
Callable[[Iterable[bool]], Array[1d, int]]
Callable[[Iterable[int], Iterable[float], int], Array[1d, int]]
Callable[[Iterable[int], Iterable[float]], Array[1d, int]]
Callable[[Iterable[int]], Array[1d, int]]
----
    flatcount = np.bincount(binnumbers, None)
                ^~~~ (o_0)
----
```

After much trying, I found it is related to 2d array indexing because if I modify `result` to a 1d array `b`, it works just fine.
```
import numpy as np


#pythran export calc_bin_more_pythran2(int, int[:], int[:])
def calc_bin_more_pythran2(Vdim, binnumbers, nbin):
    flatcount = np.bincount(binnumbers, None)
    a = np.nonzero(flatcount)

    b = np.empty([nbin.prod()], float)
    b[a] = 1.0/flatcount[a]
    return b
```

cc @rgommers

cycomanic:
I'm unsure if this is the correct fix (I don't think I properly understand the difference between begin and fbegin), however this fixes bincount if the weightsvector is a slice, which causes the error in #1819

### DIFF
--- a/pythran/pythonic/numpy/bincount.hpp
+++ b/pythran/pythonic/numpy/bincount.hpp
@@ -46,7 +46,7 @@ namespace numpy
                                 std::declval<typename E::dtype>()),
                        types::pshape<long>>>::type
     out(types::pshape<long>(length), 0L);
-    auto iweight = weights.fbegin();
+    auto iweight = weights.begin();
     for (auto iter = expr.fbegin(), end = expr.fend(); iter != end;
          ++iter, ++iweight)
       out[*iter] += *iweight;
--- a/pythran/tests/test_numpy_func2.py
+++ b/pythran/tests/test_numpy_func2.py
@@ -385,6 +385,9 @@ def test_copy0(x):
     def test_bincount2(self):
         self.run_test("def np_bincount2(a, w): from numpy import bincount; return bincount(a + 1,w)", numpy.array([0, 1, 1, 2, 2, 2]), numpy.array([0.3, 0.5, 0.2, 0.7, 1., -0.6]), np_bincount2=[NDArray[int,:], NDArray[float,:]])
 
+    def test_bincount3(self):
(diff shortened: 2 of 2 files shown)


serge-sans-paille:
Your change is correct: `fbegin` is for flat iterator (i.e. raw memory). It's more generic to use `begin()` and it's equivalent if the array is a 1D array, which is the case here. Could you just add a test case in `pythran/tests/test_numpy_func2.py` ?

cycomanic:
Done

serge-sans-paille:
Great! Thanks @cycomanic

PR #1826, opened 2021-06-29
Setup Python 3.10 testing

serge-sans-paille:


serge-sans-paille:
@hroncok I managed to get a green build (on x86_64) with the following specfile: https://sergesanspaille.fedorapeople.org/pythran3.10.spec

hroncok:
The magic is in `sed -i -e 's/-O0/-O1/g' -e 's/-Werror/-w/g' pythran/tests/__init__.py` right?

With the Fedora's flag, `-O0` gets a lot of warning. Is `-O1` required or would `-Og` work?

I'll give it a spin and also test build on other arches. Thanks!

serge-sans-paille:
-O0 gives warning because the fedora flags imply fortify source that requires
-O1. This flag is just for testing and doesn't really impact the validity of
test, so -O1 is fine.

PR #1832, opened 2021-07-06
Fix typos in Changelog

cclauss:
Fix typos in Changelog

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -264,7 +264,7 @@
 
 	* Fix np.transpose regression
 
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
Thanks for the check! I'm fine with the change, but I'd rather have a commit message body that sounds less like promotion for a tool. Could you update that? Thanks!

serge-sans-paille:
I slightly updated the commit body, thanks!

PR #1839, opened 2021-07-09
Move from networkx to a vendored solution

serge-sans-paille:
Pythran only uses a small subset of networkx, and networkx brings a lot of
dependencies with it, including scipy, which implies a circular dependency when
using pythran to compile scipy.

As quoted by « someone on the internet » it's a good property for a compiler to
have minimal dependencies because it's used to build other stuff.

As a side effect also prune the dependency on decorator and six.


rgommers:
Well that seems rather painless surgery - thanks @serge-sans-paille!

PR #1847, opened 2021-07-16
Setup 32bit testing

serge-sans-paille:


rgommers (commented):
Just changing `CXXFLAGS` will not cover all bases I believe. In SciPy we test on Azure on 32-bit Windows and Linux architecture. This will also give you 32-bit Python. The way it's done in this PR you still get `x86_64` Linux and 64-bit Python.

rgommers:
I guess that explains why this showed up on Fedora - it was a 32-bit compile on a 64-bit platform?

PR #1848, opened 2021-07-19
Improve error message when input is a view (was `(reshaped)`)

rgommers:
xref https://github.com/scipy/scipy/issues/14420. The `(reshaped)` bit has been confusing to me before.


serge-sans-paille:
That's a good first step! Thanks.

PR #1854, opened 2021-07-23
Allow upcasting from integer to floating point type during from_pytho…

serge-sans-paille:
…n conversion

This should limit the number of required signature. Still a WIP


serge-sans-paille:
cc @charlotte12l for the record

PR #1855, opened 2021-07-25
WIP: support keepdims in numpy mean

charlotte12l:
My current implementation, I haven't tested it and I'm not sure if my understanding is correct. I added four new functions in `numpy/mean.hpp `:

- `auto mean(E const &expr, types::none_type axis, dtype d, std::false_type keepdims)`: simply copy the codes in `auto mean(E const &expr, types::none_type axis, dtype d)`
- `auto mean(E const &expr, long axis, dtype d, std::false_type keepdims)`: simply copy the codes in `auto mean(E const &expr, types::none_type axis, dtype d)`
- `auto mean(E const &expr, types::none_type axis, dtype d, std::true_type keepdims)`: reshape the value array to support `keepdims`, but I'm not sure how can I specify the return type for this function.

```
>>> np.mean(np.array([[[1,2],[3,4]],[[5,6],[7,8]]]),  keepdims=False)
4.5
>>> >>> np.mean(np.array([[[1,2],[3,4]],[[5,6],[7,8]]]),  keepdims=True)
array([[[4.5]]])
>>> np.mean(np.array([[[1,2],[3,4]],[[5,6],[7,8]]]),  keepdims=False).reshape((1,1,1))
array([[[4.5]]])
```
- `auto mean(E const &expr, long axis, dtype d, std::true_type keepdims)`: expand the array `axis` dim to support `keepdims`.

```
>>> np.mean(np.array([[[1,2],[3,4]],[[5,6],[7,8]]]), axis=2, keepdims=True)
array([[[1.5],
        [3.5]],

       [[5.5],
        [7.5]]])
>>> np.expand_dims(np.mean(np.array([[[1,2],[3,4]],[[5,6],[7,8]]]), axis=2, keepdims=False), 2)
array([[[1.5],
        [3.5]],

       [[5.5],
        [7.5]]])
```

Some concerns: I found that `keepdims` is the 5th argument in `numpy.mean` while `out` is the fourth argument and we didn't support `out`.  Will it be a problem...? 

cc @serge-sans-paille @rgommers

### DIFF
--- a/pythran/backend.py
+++ b/pythran/backend.py
@@ -957,7 +957,7 @@ class CxxFunction(ast.NodeVisitor):
     def visit_Constant(self, node):
         if node in self.immediates and isinstance(node.value, bool):
             ret = "std::integral_constant<%s, %s>{}" % (
-                PYTYPE_TO_CTYPE_TABLE[type(node.value)], node.value)
+                PYTYPE_TO_CTYPE_TABLE[type(node.value)], str(node.value).lower())
         elif node.value is None:
             ret = 'pythonic::builtins::None'
         elif isinstance(node.value, bool):
--- a/pythran/backend.py
+++ b/pythran/backend.py
@@ -955,7 +955,10 @@ class CxxFunction(ast.NodeVisitor):
             return "{}({})".format(func, ", ".join(args))
 
     def visit_Constant(self, node):
-        if node.value is None:
+        if node in self.immediates and isinstance(node.value, bool):
+            ret = "std::integral_constant<%s, %s>{}" % (
+                PYTYPE_TO_CTYPE_TABLE[type(node.value)], node.value)
+        elif node.value is None:
             ret = 'pythonic::builtins::None'
         elif isinstance(node.value, bool):
             ret = str(node.value).lower()
--- a/pythran/analyses/immediates.py
+++ b/pythran/analyses/immediates.py
@@ -1,6 +1,8 @@
 """ Immediates gathers immediates. For now, only integers within shape are
 considered as immediates """
 
+import gast as ast
+from pythran.tables import MODULES
 from pythran.analyses import Aliases
 from pythran.passmanager import NodeAnalysis
 from pythran.utils import pythran_builtin, isnum
@@ -15,6 +17,12 @@ class Immediates(NodeAnalysis):
 
     def visit_Call(self, node):
         func_aliases = self.aliases[node.func]
+        if len(func_aliases) == 1 and next(iter(func_aliases)) is MODULES['numpy']['mean']:
+            if len(node.args) >= 5:
+                keep_dims = node.args[4]
+                if isinstance(keep_dims, ast.Constant):
+                    self.result.add(keep_dims)
+
         if len(func_aliases) == 1 and next(iter(func_aliases)) is _make_shape:
             self.result.update(a for a in node.args
                                if isnum(a)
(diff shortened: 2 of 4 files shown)


rgommers:
> Some concerns: I found that `keepdims` is the 5th argument in `numpy.mean` while `out` is the fourth argument and we didn't support `out`. Will it be a problem...?

I think `out` can be added in the Pythran signature - but it must be `None`. If the user passes anything but `None`, then raise an exception.

charlotte12l:
@serge-sans-paille Hey Serge, do you have any suggestions on declaring the return type for `auto mean(E const &expr, types::none_type axis, dtype d, std::true_type keepdims)`? ... I think we need to calculated the `out_shape`  so we can ` -> decltype(numpy::functor::asarray{}(sum(expr) / 
 typename dtype::type(expr.flat_size())).reshape(out_shape))` ...

pythran/backend.py
```diff
@@ -955,7 +955,10 @@ def visit_Call(self, node):
             return "{}({})".format(func, ", ".join(args))
 
     def visit_Constant(self, node):
-        if node.value is None:
+        if node in self.immediates and isinstance(node.value, bool):
+            ret = "std::integral_constant<%s, %s>{}" % (
+                PYTYPE_TO_CTYPE_TABLE[type(node.value)], node.value)
```
serge-sans-paille, line 960: `node.value` will print `True` instead of `true`. You should use `str(node.value).lower()` instead

serge-sans-paille:
> @serge-sans-paille Hey Serge, do you have any suggestions on declaring the return type for `auto mean(E const &expr, types::none_type axis, dtype d, std::true_type keepdims)`? ... I think we need to calculated the `out_shape` so we can ` -> decltype(numpy::functor::asarray{}(sum(expr) / typename dtype::type(expr.flat_size())).reshape(out_shape))` ...

We could speciailize the case where the `axis` is `None`, but a good start would be `types::ndarray<typename E::dtype, types::array<long, E::value>>`

serge-sans-paille:
Extra note : you should add a few test case in `pythran/tests/test_numpy_func0.py` with the explicit `keepdims`

charlotte12l:
Hi @serge-sans-paille, could you help look at why the checks always fail the tests if you got some time? I tried the following tests but they always failed with `no matching function for call to object of type`

```
    def test_mean6(self):
        self.run_test("def np_mean6(a): from numpy import mean ; from numpy import float64; return mean(a, 2, float64, None, False)", numpy.array([[[1, 2], [3, 4.]]]), np_mean6=[NDArray[float,:,:,:]])
```
or
```
def test_mean6(self):
        self.run_test("def np_mean6(a): from numpy import mean ; return mean(a, 2, keepdims=False)", numpy.array([[[1, 2], [3, 4.]]]), np_mean6=[NDArray[float,:,:,:]])
```

serge-sans-paille:
@charlotte12l some fixes here: https://github.com/serge-sans-paille/pythran/tree/fix/add_keepdims
The remaining parts to fix are probably beyond your scope, I'll try to do that today.

serge-sans-paille:
Update pushed for more reducer :-)

charlotte12l:
Thanks! I totally forgot to modify the file pythran/pythonic/include/numpy/mean.hpp correspondingly...

So smart to use `make_scalar_pshape`, I only knew creating an array and fill it lol.

I added some comments on the following codes- if node.value is int, can we still do str(node.value).lower()? 
```
if node in self.immediates:
            assert isinstance(node.value, int)
            return "std::integral_constant<%s, %s>{}" % (
                PYTYPE_TO_CTYPE_TABLE[type(node.value)], str(node.value).lower())
```

serge-sans-paille:
> I added some comments on the following codes- if node.value is int, can we still do str(node.value).lower()?

yes. `str(<some-int-value>)` will give a number as string, which is unaffected by `lower()`

serge-sans-paille:
Obsoleted by #1869

PR #1859, opened 2021-08-01
Correctly handle combiner with range

### LINKED ISSUE #1857: set.update(iterable) does not work as expected
I'm having some trouble with passing an iterable to set.update. I would expect it to update the set it's called on with the contents of the iterable, but it does not seem to have that effect. I've included a representative iPython session below.

Thanks for the hard work you put in on this project.

```
Python 3.7.4 (default, Aug 13 2019, 20:35:49)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.25.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import pythran

In [2]: pythran.__version__
Out[2]: '0.9.12.post1'

In [3]: %load_ext pythran.magic

In [4]: def foo():
   ...:     s=set()
   ...:     s.update(range(3))
   ...:     return s
   ...:

In [5]: foo()
Out[5]: {0, 1, 2}

In [6]: %%pythran
   ...: #pythran export foo()
   ...: def foo():
   ...:     s=set()
   ...:     s.update(range(3))
   ...:     return s
   ...:

In [7]: foo()
Out[7]: set()
```

serge-sans-paille:
Fix #1857

### DIFF
--- a/pythran/pythonic/builtins/range.hpp
+++ b/pythran/pythonic/builtins/range.hpp
@@ -54,6 +54,25 @@ namespace builtins
     return *this;
   }
 
+  range_iterator &range_iterator::operator--()
+  {
+    value_ -= step_;
+    return *this;
+  }
+
+  range_iterator range_iterator::operator--(int)
+  {
+    range_iterator self(*this);
+    value_ -= step_;
+    return self;
+  }
+
+  range_iterator &range_iterator::operator-=(long n)
+  {
+    value_ -= step_ * n;
+    return *this;
+  }
+
   bool range_iterator::operator!=(range_iterator const &other) const
   {
     return value_ != other.value_;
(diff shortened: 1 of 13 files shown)


phtown:
Sorry to be a bother, but I've also noticed these not working as expected:
```
s=set()
s.update(x*2 for x in range(3))
s=set()
s.update(itertools.permutations(range(2)))
```
They both seem to work fine as long as the set isn't empty, and directly constructing a set as in `set(x*2 for x in range(3))` or `set(itertools.permutations(range(2)))` seems to work fine as well. I haven't tried to track down a complete list of every way to construct an iterable, but I can try looking into that if that would be helpful.

serge-sans-paille:
Should fix both issues now.

serge-sans-paille:
Hell, I forgot `itertools.permutation`. But I've got the pattern I'll double check the codebase.

PR #1862, opened 2021-08-02
Change IRC reference from Freenode to OFTC

fuglede:
Cf. [this announcement](https://www.freelists.org/post/pythran/Moving-the-IRC-chan-from-freenode-to-oftc) on the mailing list.

We also secure a few HTTP references in the process.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 Pythran
 #######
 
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
Thanks!

PR #1865, opened 2021-08-04
Accept some views as regular plain arrays

### LINKED ISSUE #1864: Array sanitized with np.asarray(arr, order='C') can still cause a TypeError.
An array that is sanitized with `np.asarray(arr, order='C')` can still cause pythran to throw a `TypeError`. This can happen when one of the axes of the array has length 0 or 1. If an axis has length 0 or 1, its corresponding stride can be arbitrary even though the data is C-contiguous (as described [here](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html#numpy.ndarray.flags)). When the array has an arbitrary stride, it fails validation that pythran does here:  https://github.com/serge-sans-paille/pythran/blob/e314a76d34685fd2e41f653e9b35f70b9c504f42/pythran/pythonic/types/ndarray.hpp#L1542 (and potentially elsewhere). 

This should reproduce the issue:

```python
# mod.py

# pythran export func(float[:, :])
def func(arr):
    return
```

```python
# script.py

import numpy as np
import mod

arr = np.random.random((1, 3))[:, :2]
arr = np.asarray(arr, order='C')
mod.func(arr)
```

which will throw this error:
```
Traceback (most recent call last):
  File "script.py", line 6, in <module>
    mod.func(arr)
TypeError: Invalid call to pythranized function `func(float64[:, :] (reshaped))'
Candidates are:

    - func(float[:,:])
```

This is related to the scipy issue https://github.com/scipy/scipy/issues/14420

serge-sans-paille:
A.k.a correctly handle strides with dimensions 0 or 1

Fix #1864

### DIFF
--- a/pythran/pythonic/types/ndarray.hpp
+++ b/pythran/pythonic/types/ndarray.hpp
@@ -1536,18 +1536,21 @@ bool from_python<types::ndarray<T, pS>>::is_convertible(PyObject *obj)
   auto const *stride = PyArray_STRIDES(arr);
   auto const *dims = PyArray_DIMS(arr);
   long current_stride = PyArray_ITEMSIZE(arr);
-  for (long i = std::tuple_size<pS>::value - 1; i >= 0; i--) {
-    if (stride[i] == 0 && dims[i] == 1) {
-      // happens when a new dim is added though None/newaxis
-    } else if (stride[i] != current_stride)
+  if (PyArray_SIZE(arr)) {
+    for (long i = std::tuple_size<pS>::value - 1; i >= 0; i--) {
+      if (stride[i] == 0 && dims[i] == 1) {
+        // happens when a new dim is added though None/newaxis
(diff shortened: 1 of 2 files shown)


treverhines:
Thanks for addressing the issue. I think there may still be some issues with zero-sized arrays with this patch. If an array has no elements, then _any_ of the strides can be arbitrary, not just the stride for the axis with length 0 (contrary to what I said in https://github.com/serge-sans-paille/pythran/issues/1864 ).  For example, I believe `np.random.random((0, 2, 3))[:, :, :2]` would still cause pythran to throw an error despite numpy considering it C-contiguous (I am unfortunately not a a computer to test this).

serge-sans-paille:
What a nice cornercase :-)

serge-sans-paille:
Updated with test case! Thanks for the follow-up

PR #1866, opened 2021-08-04
Basic support for to_eval of np._globals._NoValueType

serge-sans-paille:
This is required to correctly compute the signature of numpy.mean, and should
unlock #1855


serge-sans-paille:
@charlotte12l once I merge this one, you can rebase #1855 on master branch

PR #1869, opened 2021-08-10
Feature/add keep dims

serge-sans-paille:


serge-sans-paille:
\o/ thanks @charlotte12l for the hard work!

Issue #1842, opened 2021-07-13
`unique` implementation struggles with overloads


rgommers:
```python
import numpy as np

#pythran export try_unique(float[:])
def try_unique(x):
    return np.unique(x, return_counts=False)
```

A simple test shows the number of return values is wrong:
```python
>>> x = np.ones(7, dtype=float)
>>> try_unique(x)
(array([1.]), array([0]), array([0, 0, 0, 0, 0, 0, 0]), array([7]))
```

This is in a similar boat as the `keepdims` keyword (see gh-1564, gh-1820): a boolean keyword that controls return type. I'd say `unique` is worse actually.

This is going to trip people up, it may be worth considering to support boolean keywords by making separate types for `bool_true` and `bool_false` that can get different overloads. If this is too hard, I think it makes sense to only support the single argument version of `unique`.

charlotte12l:
The returned value is the same as `np.unique(x, return_index=True, return_inverse=True, return_counts=True)`.

```
>>> np.unique(x, return_index=True, return_inverse=True, return_counts=True)
(array([1.]), array([0], dtype=int64), array([0, 0, 0, 0, 0, 0, 0], dtype=int64), array([7], dtype=int64))
```

Looking at the implementation, we'll find for key `return_counts` in `np.unique` , Pythran only supports the following:
```
  template <class E>
  std::tuple<types::ndarray<typename E::dtype, types::pshape<long>>,
             types::ndarray<long, types::pshape<long>>,
             types::ndarray<long, types::pshape<long>>,
             types::ndarray<long, types::pshape<long>>>
  unique(E const &expr, bool return_index, bool return_inverse,
         bool return_counts)
```
, and it will always think the keys are true.

serge-sans-paille:
@charlotte12l before working on the generalization of `numpy.mean` case, you could add a special case for `np.unique`. It's easier to generalize when you have two examples instead of one ;-)

serge-sans-paille:
Fixed by #1876

PR #1876, opened 2021-08-13
Support  boolean arguments in numpy unique

charlotte12l:
https://github.com/serge-sans-paille/pythran/issues/1842#issuecomment-896320261

Hard code boolean arguments in numpy unique, at first, I only consider the case `len(node.args) >= 4` & `return_index` is true & `return_inverse` is true & `return_counts` is true.  

Next:
- [done]consider the case when `len(node.args) == 3` or `2` 
- [done] support `return_index` is true/false & `return_inverse` is true/false & `return_counts` is true/false.  
- use offset to generalize

### DIFF
--- a/pythran/analyses/immediates.py
+++ b/pythran/analyses/immediates.py
@@ -23,6 +23,20 @@ class Immediates(NodeAnalysis):
                 if isinstance(keepdims, ast.Constant):
                     self.result.add(keepdims)
 
+        if len(func_aliases) == 1 and next(iter(func_aliases)) is MODULES['numpy']['unique']:
+            if len(node.args) >= 2:
+                return_index = node.args[1]
+                if isinstance(return_index, ast.Constant):
+                    self.result.add(return_index)
+            if len(node.args) >= 3:
+                return_inverse = node.args[2]
+                if isinstance(return_inverse, ast.Constant):    
+                    self.result.add(return_inverse)
+            if len(node.args) >= 4:
+                return_counts = node.args[3]
+                if isinstance(return_counts, ast.Constant): 
+                    self.result.add(return_counts)
+
         if len(func_aliases) == 1 and next(iter(func_aliases)) is _make_shape:
             self.result.update(a for a in node.args
                                if isnum(a)
--- a/pythran/pythonic/include/numpy/unique.hpp
+++ b/pythran/pythonic/include/numpy/unique.hpp
@@ -15,21 +15,83 @@ namespace numpy
   template <class E>
   std::tuple<types::ndarray<typename E::dtype, types::pshape<long>>,
              types::ndarray<long, types::pshape<long>>>
-  unique(E const &expr, bool return_index);
+  unique(E const &expr, std::true_type return_index);
+
+  template <class E>
+  types::ndarray<typename E::dtype, types::pshape<long>>
+  unique(E const &expr, std::false_type return_index);
+
+  template <class E>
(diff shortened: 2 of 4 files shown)


serge-sans-paille:
Looks like a good start! Can you rebase on master, then add a few test case?

charlotte12l:
@serge-sans-paille  Sure, done. 

Besides, I'm wondering how Pythran know the default value is false? e.g. for `unique(x, return_inverse=True)`, how does Pythran know it should call `unique(E const &expr, std::false_type return_index, std::true_type return_inverse)` rather than `unique(E const &expr, std::true_type return_index, std::true_type return_inverse)`?

serge-sans-paille:
> Besides, I'm wondering how Pythran know the default value is false?

It's using `inspect.gefullargspecs` to get introspection info and retrieve the default value from the actual code of `numpy.unique` ;-)

serge-sans-paille:
This looks very good to me :-) Can you squash these three commits in one and we're done!

charlotte12l:
> This looks very good to me :-) Can you squash these three commits in one and we're done!

Oh I think I didn't finish adding all the permutations, I made a table, and seems we need support 15 kinds:(?)


No. | return_index | return_inverse | return_counts |  
-- | -- | -- | -- | --
1 | N/A | N/A | N/A | already have
2 | TRUE | N/A | N/A | already have
3 | FALSE | N/A | N/A | I added
4 | TRUE | TRUE | N/A | already have
5 | FALSE | TRUE | N/A | I added
6 | FALSE | FALSE | N/A | I added
7 | TRUE | FALSE | N/A |  
8 | TRUE | TRUE | TRUE | already have
9 | TRUE | TRUE | FALSE |  
10 | TRUE | FALSE | FALSE |  
11 | TRUE | FALSE | TRUE |  
12 | FALSE | TRUE | FALSE |  
13 | FALSE | TRUE | TRUE |  
14 | FALSE | FALSE | FALSE |  
15 | FALSE | FALSE | TRUE |  

However, some of them will have the same results, e.g.  | FALSE | FALSE | FALSE |  vs. | FALSE | FALSE | N/A | vs. | FALSE | N/A | N/A |  vs. | N/A | N/A | N/A | but the function definition would be different  `unique(E const &expr, std::false_type return_index, std::false_type return_inverse, std::false_type return_counts);`  vs. `unique(E const &expr, std::false_type return_index, std::false_type return_inverse);` vs.  `unique(E const &expr, std::false_type return_index);` vs. `unique(E const &expr);` . In this way, we need to support 4 functions rather than 1, is that right?

charlotte12l:
@serge-sans-paille Now I've added support for all the possible permutations and tests, could you check if they look fine and then I can squash these commits in one?  Besides, need we also add tests for computation correctness? like the example in https://github.com/serge-sans-paille/pythran/issues/1842 :)

serge-sans-paille:
Computation corecteness is a side effect of the test: every test tests that the return value is the same as the one from Python code, and you don't need to do anything special for that to happen. I'll have a look to the implementation.

serge-sans-paille:
There may be room for some code syndication, but let's leave that for another commit, I'm happy with current state!

serge-sans-paille:
Can you just squash that into a single commit please?

charlotte12l:
@serge-sans-paille  Done :)

serge-sans-paille:
Thanks for the hard work o/

PR #1880, opened 2021-08-27
Allow implicit conversion for immediate parameter

serge-sans-paille:
This allows to pass keep_dims=1 even though a bool is expected.


serge-sans-paille:
Ooops™ - fixed

jeanlaroche:
Yes, I confirm that it now compiles! Yepeee!

PR #1882, opened 2021-08-29
README.rst: Remove networkx from debian instructions

mkoeppe:


serge-sans-paille:
Thanks for spotting that!

PR #1883, opened 2021-08-29
pythran/types/tog.py: Use 'import gast as ast' for uniformity

mkoeppe:


serge-sans-paille:
I'm filne with the change, although tog is likely to be obsoleted at some point :-/

PR #1889, opened 2021-09-07
Add pythran option -ftime-report to print out the time spent on optim…

jeanlaroche:
…izations

Not sure about the way I'm printing results out (using print, perhaps logging would be more appropriate).

### DIFF
--- a/docs/CLI.rst
+++ b/docs/CLI.rst
@@ -76,7 +76,7 @@ To know more options about Pythran, you can check::
   $> pythran --help
   usage: pythran [-h] [-o OUTPUT_FILE] [-P] [-E] [-e] [-v] [-w] [-V] [-p pass]
                  [-I include_dir] [-L ldflags] [-D macro_definition]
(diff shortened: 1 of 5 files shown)


serge-sans-paille:
Any reason why youre collecting in a dict and dumping the dict at the end, instead of just logging  within each pass?

jeanlaroche:
> Any reason why youre collecting in a dict and dumping the dict at the end, instead of just logging within each pass?

I didn't think of reporting it right off, but I can see two reasons
- That way I cumulate the times for 2 passes of the same algo (maybe not desirable!)
- That way I can clean up the dispay (who cares?)
No good reason, I can change that if you prefer.

serge-sans-paille:
ok, your arguments make sense. Can you update the `--help` output in docs/CLI.rst ?

jeanlaroche:
Done

serge-sans-paille:
Obsoleted by #1902 which is just a port + small fixes of this PR

PR #1895, opened 2021-09-11
Fix corner case when a numpy_gexpr is initialized from another gexpr

serge-sans-paille:
But with a different, static shape, which may trigger an irrelevant static
check. Replace the check by an assert.


serge-sans-paille:
cc @jeanlaroche

serge-sans-paille:
@jeanlaroche pelase tell me if this fixes your issue, if so I'll add that to the release.

jeanlaroche:
Just checked. It fixes the issue.

PR #1905, opened 2021-10-17
Update rvalue type when needed

### LINKED ISSUE #1897: Pythran still has problems with arrays initialized to fixed 0 size used in a tuple.
The following code does not compile. But if the fixed size initialization of self_X is replaced by the unknown-at-compile-time size int(N), then things compile.
```
#pythran export test(int?)
import numpy as np

def CL___init__(N):
    self_X = np.zeros((0, 0))
    # N = int(N)                      # <- Enable these two lines, and things compile.
    # self_X = np.zeros((0*N, 0*N))   # <- Enable these two lines, and things compile.
    return tuple([self_X])

def CL_test(self):
    self[0] = np.zeros((30, 30))
    Y = []
    X = self[0]
    for ii in range(2):
        Y.append(X[:, ii:ii+1])
    Y = np.stack(Y, axis=0)
    print(Y)

def test(N = 10):
    CL = CL___init__(N)
    CL_test(CL)
```

serge-sans-paille:
Should fix #1897


serge-sans-paille:
@jeanlaroche yes, it's real, This really is a fix for you :-)

PR #1917, opened 2021-11-15
Document how ``int`` type is implemented in pythran

### LINKED ISSUE #1916: Make the potential `int` size pitfall more prominent
I was perusing the docs when [I found among the Limitations](https://pythran.readthedocs.io/en/latest/MANUAL.html#limitations)
> There is no BigInt support. All integer operations are performed on long (probably int64_t on a 64 bit machine.)

If this is indeed the case, there's [an infamous pitfall that NumPy has to deal with](https://github.com/numpy/numpy/issues/9464), namely that a C `long` on 64-bit Windows is 32-bit. This causes terrible surprises, for instance `np.arange` giving you int32 values even on 64-bit Windows.

If the note about `long` is still true, perhaps a more prominent warning should be put next to that bullet point, raising awareness of this underlying issue (i.e. that you get 32-bit ints on a 64-bit system). I have a Windows virtual machine somewhere, I can also experiment myself to check if the behaviour is indeed the same.

Probably related issue: https://github.com/serge-sans-paille/pythran/issues/1811

serge-sans-paille:
Fix #1916

### DIFF
--- a/docs/MANUAL.rst
+++ b/docs/MANUAL.rst
@@ -296,8 +296,9 @@ Unfortunately it's not always possible:
 
 - Pythran does not support heterogeneous containers (except tuples).
 
-- There is no BigInt support. All integer operations are performed on ``long``
-  (probably ``int64_t`` on a 64 bit machine.)
+- There is no BigInt support. All integer operations are performed on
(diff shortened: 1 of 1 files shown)


adeak (approved):
Looks great! Thank you :)

PR #1936, opened 2021-11-29
Remove use of npy_double in interp

jeanlaroche:


serge-sans-paille:
clean and simple, thanks!

PR #1954, opened 2022-01-18
bug 1950: fix the test of periodic interp, so fp is indeed periodic i…

jeanlaroche:
…n the test.


serge-sans-paille:
Clean and simple, thanks @jeanlaroche

PR #1966, opened 2022-02-22
Instructions for Windows 10 users MKL / OpenBLAS properly formatted

mattslezak-shell:
Instructions given so OpenMP will work on Windows 10 with both MKL and OpenBLAS.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -89,13 +89,64 @@ Using ``dnf``::
 Windows
 =======
 
-Windows support is on going and only targets Python 3.5+ with either Visual Studio 2017 or, better, clang-cl::
+Windows support is on going and only targets Python 3.5+ with clang-cl::
 
     $> pip install pythran
 
-Note that using ``clang-cl.exe`` is the default setting. It can be changed
-through the ``CXX`` and ``CC`` environment variables.
+Note that using ``clang-cl.exe`` is the default setting. It can be changed through the ``CXX`` and ``CC`` environment variables, but should be left as-is.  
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -89,13 +89,64 @@ Using ``dnf``::
...
+    cflags=/std:c++14 /w -Xclang -fopenmp
+    ldflags=\libiomp5md.lib
+    blas=pythran-openblas
+    CC=clang-cl.exe
+    CXX=clang-cl.exe
+
+Note with the ``OpenBLAS`` package, you need to: ``pip install pythran-openblas`` so the library can be linked properly.
+
+
+Using the LLVM linker (optional):
```
serge-sans-paille, line 137: I think this is more than hacky, we cannot suggest to modify system files here :-) This is worth a configuration entry in the pythran config file though.

I'm curious about the motivation for using `lld` here?

mattslezak-shell:
Change whatever you feel is appropriate.  I did notice on a very large project with OpenBLAS I had to change some settings around to get OpenMP working like this:

library_dirs='C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.29.30133\lib\x64'
ldflags=\libomp.lib

So link the standard Windows Visual Studio OpenMP library instead of the LLVM one.  Why this works on larger projects (100 lines of code) and not the LLVM one I have no idea.  So rather than figure it out, I'd just provide it as an alternative if the first method fails...

PR #1970, opened 2022-03-02
Feature/faster argument effects

serge-sans-paille:


jeanlaroche:
I looks like all my projects do compile with the version on this branch. Kudos Serge! The C++ code generation is much much faster than before (almost twice as fast?)

PR #1971, opened 2022-03-04
Fix/faster constant folding

serge-sans-paille:


serge-sans-paille:
@jeanlaroche this one should please you and decrease C++ generation time

Issue #1962, opened 2022-02-12
GCC 12: compilation error


marxin:
Using the GCC 12 compiler, I see the following error:

```
[ 2985s] /home/abuild/rpmbuild/BUILD/pythran-0.11.0/pythran/pythonic/types/numpy_binary_op.hpp:15:1: note: in expansion of macro ‘NUMPY_BINARY_FUNC_NAME’
[ 2985s]    15 | NUMPY_BINARY_FUNC_NAME(E0 &&self, E1 &&other)
[ 2985s]       | ^~~~~~~~~~~~~~~~~~~~~~
[ 2985s] /home/abuild/rpmbuild/BUILD/pythran-0.11.0/pythran/pythonic/types/numpy_binary_op.hpp: In substitution of ‘template<class E0, class E1> typename std::enable_if<{anonymous}::pythonic::types::valid_numop_parameters<typename std::decay<_Tp>::type, typename std::decay<_Tp2>::type>::value, {anonymous}::pythonic::types::numpy_expr<{anonymous}::pythonic::operator_::functor::mul, typename {anonymous}::pythonic::types::adapt_type<E0, E1>::type, typename {anonymous}::pythonic::types::adapt_type<E1, E0>::type> >::type {anonymous}::pythonic::types::operator*(E0&&, E1&&) [with E0 = {anonymous}::pythonic::types::str; E1 = long int&]’:
[ 2985s] /home/abuild/rpmbuild/BUILD/pythran-0.11.0/pythran/pythonic/include/operator_/mul.hpp:13:57:   required by substitution of ‘template<class A, class B> decltype ((forward<A>(a) * forward<B>(b))) {anonymous}::pythonic::operator_::mul(A&&, B&&) [with A = {anonymous}::pythonic::types::str; B = long int&]’
[ 2985s] /tmp/tmpuz6vaowl.cpp:130:113:   required from ‘typename __pythran_test_functional_variant_assign0::functional_variant_assign0::type<argument_type0>::result_type __pythran_test_functional_variant_assign0::functional_variant_assign0::operator()(argument_type0&&) const [with argument_type0 = long int&; typename type<argument_type0>::result_type = {anonymous}::pythonic::types::str]’
[ 2985s] /tmp/tmpuz6vaowl.cpp:143:115:   required from here
[...]

PR #1976, opened 2022-03-14
Add default constructor to string_iterator

jwakely:
A type must be default constructible to meet the forward iterator requirements.
GCC 12 won't compile `std::reverse_iterator<string_iterator>` without
this change.


marxin:
Needed for #1962.

serge-sans-paille:
Patch successfully tested on Fedora/rawhide :  https://koji.fedoraproject.org/koji/taskinfo?taskID=84170045

PR #1977, opened 2022-03-14
Fix building with GCC 12

marxin:
Follow suggestions made by GCC's C++ maintainer:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104733#c5

Fixes: #1962


serge-sans-paille:
Probably a duplicate of #1979

PR #1978, opened 2022-03-14
Remove template argument list from destructor name

jwakely:
This is not valid since C++20, see https://wg21.link/cwg2237

### DIFF
--- a/pythran/pythonic/types/variant_functor.hpp
+++ b/pythran/pythonic/types/variant_functor.hpp
@@ -112,7 +112,7 @@ namespace types
     }
 
     template <class Type>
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
I'm 100% fine with the change, but I also wonder how you spotted that as pythran always compiles with `-std=c++11` on Linux

jwakely:
No, I didn't spot that. I was compiling a preprocessed `.ii` file that somebody provided to me (to debug the problem fixed by #1976) and decided to see what would happen if I added `-std=c++20`. So this isn't causing a problem for me (or Fedora), it was just something I noticed.

PR #1980, opened 2022-03-15
Update distutils testing

serge-sans-paille:


pythran/tests/test_distutils.py
```diff
@@ -34,8 +34,10 @@ def test_setup_build(self):
         libdir = os.path.join(base, 'lib')
         if not os.path.isdir(libdir):
             libdir = os.path.join(base, 'lib64')
-        check_call(['python', '-c', 'import demo'],
-                   cwd=os.path.join(libdir, python_version, 'site-packages'))
+        check_call(['python', '-c',
+                    'import sys; sys.path.append("{}"); import demo'.format(
```
paugier, line 38: Just by curiosity, what means `"{}"` for a path? Why do you change that?

serge-sans-paille: That was an attempt to fix fedora build after setuptools update. And it fails :-/
`{}` here just performs string substitution

serge-sans-paille:
This does not fix the build on Fedora as I expected, closing

PR #1985, opened 2022-04-04
Simplify global declaration processing

serge-sans-paille:
This speeds up things a bit: no need to compute all use-def chains here.


serge-sans-paille:
cc @jeanlaroche ;-)

PR #1986, opened 2022-04-07
Feature/faster compilation

serge-sans-paille:


serge-sans-paille:
Again, this should please @jeanlaroche

PR #1987, opened 2022-04-08
Improve pattern matching implementation

serge-sans-paille:
Instead of trying to match every node, dispatch based on the type.

Harder, Better, Faster, Stronger.


serge-sans-paille:
cc @jeanlaroche I can't stop, I'm having so much fun ;-)

PR #1992, opened 2022-05-17
Spelling typo nop -> nope

swallan:
I was using Pythran and noticed this warning message, it seemed to me like this was intended to be "Nope..." instead of "Nop". Thanks!


serge-sans-paille:
^^! thanks

PR #1996, opened 2022-06-06
Don't use removed sysconfig.get_config_vars()["SO"] in a doctest

hroncok:
This was removed from Python 3.11+, see https://github.com/python/cpython/issues/91670


serge-sans-paille:
Thanks for spotting this!

PR #1998, opened 2022-06-11
Feature/faster compilation 2

serge-sans-paille:


serge-sans-paille:
cc @jeanlaroche for the timing testing :-)

PR #2004, opened 2022-07-06
Introduce --config=backend.annotate=true

### LINKED ISSUE #2001: Provide debugging info when face with a C++ compilation error
At this point, for complex/long .py file, it is very difficult trace back the origin of a C++ compilation problem when one arise. 
It would be very useful to have a  mapping that can relate a line of the C++ code to the line in the .py that was at its origin.
This is especially true for the long list of typdefs that happens at the start of all C++ functions, which are generated from the lines of code in the .py function, but are nearly impossible to associate with the python code.

Ideally, when a C++ compilation error, pythran should be able to say: this was caused by line xxx of file.py

serge-sans-paille:
When on, this option generates chunk of python code as comments within the C++
code, making it easier to track failures.

Fix #2001

### DIFF
--- a/docs/MANUAL.rst
+++ b/docs/MANUAL.rst
@@ -627,6 +627,15 @@ Another internal setting stuff. This controls the accuracy of the typing phase. 
     # This algorithms generates code difficult to compile for g++, but not clang++
     enable_two_steps_typing = False
 
+``[backend]``
+*************
+
+This controls some behavior of the C++ backend, so the default should be safe::
+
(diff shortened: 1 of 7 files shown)


serge-sans-paille:
cc @jeanlaroche

jeanlaroche:
Wow this is going to change my life! So cool!

serge-sans-paille:
The typing issue is real but cannot be solved that way. #2005 shows on going work to cleanup typer code, and hopefully generate cleaner types!

PR #2008, opened 2022-07-19
Setup a directory for samples challenging icc limits

serge-sans-paille:


serge-sans-paille:
cc @paugier: your test case passes validation here

PR #2037, opened 2022-11-17
Set intersection fix

### LINKED ISSUE #2036: Set intersection causes segfaults
The intersection of two sets triggers a Segmentation fault:

```
#pythran export f(int)
def f(N):
    a = {i for i in range(N)}
    b = {3, 4, 5}
    return a.intersection(b)
```

Union, difference, and issubset work.

Both latest release and master. Python 3.9 running on Linux.

Fix coming up.

Dapid:
Fixes #2036


serge-sans-paille:
Nice catch :-)

PR #2042, opened 2022-11-24
Increase and parametrize the maximum number of heterogeneous type in …

### LINKED ISSUE #2040: dict cannot hold more than 8 functions
I have a pythran module with a dict that maps strings of function names to pythran functions.  However, it seems like this dict cannot hold more than 8 items. When I try to retrieve anything past the 8th item, I get the 2nd item instead.  Incidentally, I get the same value when using a key that does not exist in the dict.

I encounted this issue when trying to add another kernel to `scipy.interpolate.RBFInterpolator`. Specifically, adding to this dict https://github.com/scipy/scipy/blob/5b8490a2b80e73aec59af9ac93bca2ebd178fee1/scipy/interpolate/_rbfinterp_pythran.py#L39

Here is my module demonstrating the issue:
```python
  # mod.py

def one():
    return 1.0

def two():
    return 2.0

def three():
    return 3.0

def four():
    return 4.0

def five():
    return 5.0

def six():
    return 6.0

def seven():
    return 7.0

def eight():
    return 8.0

def nine():
    return 9.0

def ten():
    return 10.0

NAME_TO_FUNC = {
    "one": one,
    "two": two,
    "three": three,
    "four": four,
    "five": five,
    "six": six,
    "seven": seven,
    "eight": eight,
    "nine": nine,
    "ten": ten,
    }

# pythran export evaluate_func(str)
def evaluate_func(name):
    func = NAME_TO_FUNC[name]
    return func()
```

Running this script
```python 
import mod
print(mod.evaluate_func('one'))
print(mod.evaluate_func('two'))
print(mod.evaluate_func('three'))
print(mod.evaluate_func('four'))
print(mod.evaluate_func('five'))
print(mod.evaluate_func('six'))
print(mod.evaluate_func('seven'))
print(mod.evaluate_func('eight'))
print(mod.evaluate_func('nine'))
print(mod.evaluate_func('ten'))
print(mod.evaluate_func('not in dict'))
```
I get this output
```
1.0
2.0
3.0
4.0
5.0
6.0
7.0
8.0
2.0
2.0
2.0
```

serge-sans-paille:
…a sequence

This only partially fixes #2040.

It does increase the bound, and it is now possible to change it using --config typing.max_heterogeneous_sequence_size=32.

There is however a subtle error with invalid lookup that requires more care, and probably another commit!

### DIFF
--- a/docs/MANUAL.rst
+++ b/docs/MANUAL.rst
@@ -613,15 +613,20 @@ Another internal setting stuff. This controls the accuracy of the typing phase. 
 
     [typing]
 
-    # maximum number of container access taken into account during type inference
-    # increasing this value inreases typing accuracy
-    # but slows down compilation time, to the point of making g++ crash
-    max_container_type = 2
-
     # maximum number of combiner per user function
     # increasing this value inreases typing accuracy
     # but slows down compilation time, to the point of making g++ crash
     max_combiner = 2
+
+    # above this number of overloads, pythran specifications are considered invalid
+    # as it generates ultra-large binaries
+    max_export_overloads = 128
+
+    # to the notable exceptions of tuple, pythran sequences are homogeneous. It is
+    # however possible to store functions objects in a sequence and a variant
+    # functor is created. This value bounds the number of different types within
+    # a sequence
+    max_heterogeneous_sequence_size = 16
 
 ``[backend]``
 *************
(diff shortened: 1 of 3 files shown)


serge-sans-paille:
Obsoleted by #2060

Issue #2048, opened 2022-11-27
`numpy.distutils` is deprecated and may not be present, or broken - only use it conditionally


rgommers:
`numpy.distutils` wont be present on Python 3.12, and is fairly regularly broken by new `setuptools` releases even on older versions. So what is needed is to use it conditionally, only if it's present and it is actually needed.

I'll work on a PR for this.

PR #2049, opened 2022-11-27
Only use `numpy.distutils` conditionally

rgommers:
It's deprecated, is easily broken by new setuptools releases, and will not be present at all for Python >= 3.12

Closes gh-2048

### DIFF
--- a/omp/__init__.py
+++ b/omp/__init__.py
@@ -2,12 +2,10 @@
 
 from ctypes.util import find_library
 from subprocess import check_output, CalledProcessError, DEVNULL
-from numpy.distutils.misc_util import (
-    msvc_runtime_major, get_shared_lib_extension
-)
 import ctypes
 import os
 import sys
+import sysconfig
 
 try:
     # there may be an environ modification when loading config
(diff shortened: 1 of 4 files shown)


pythran/config.py
```diff
@@ -223,6 +221,7 @@ def parse_define(define):
 
     # numpy specific
     if python:
+        import numpy
```
serge-sans-paille, line 224: why moving the import here?

omp/__init__.py
```diff
@@ -18,6 +16,70 @@ def compiler():
 cxx = compiler()
 
 
+def get_shared_lib_extension(is_python_ext=False):
```
serge-sans-paille, line 19: Just in case: Is this function extracted from another codebase? If so, we should mention it somewhere.

pythran/config.py
```diff
@@ -246,18 +245,24 @@ def parse_define(define):
...
+                    # required to cope with atlas missing extern "C"
+                    extension['define_macros'].append('PYTHRAN_BLAS_{}'
+                                                      .format(user_blas.upper()))
+                    extension['libraries'].extend(numpy_blas.get('libraries', []))
+                    extension['library_dirs'].extend(
+                        numpy_blas.get('library_dirs', []))
+                    extension['include_dirs'].extend(
+                        numpy_blas.get('include_dirs', []))
+            except Exception:
+                # This may fail, because of setuptools backwards compat issues
```
serge-sans-paille, line 264: In that case, it should probably be an hard error, or at least a warning

pythran/dist.py
```diff
@@ -15,7 +15,13 @@
 
 from distutils.command.build_ext import build_ext as LegacyBuildExt
 
-from numpy.distutils.extension import Extension
+try:
+    # `numpy.distutils` is deprecated, and won't be present on Python >=3.12
+    # If it is installed, we need to use it though, so try-import it:
+    from numpy.distutils.extension import Extension
+except:
```
serge-sans-paille, line 22: `except ImportError` ?

serge-sans-paille:
@rgommers in the distuils-free world, is there an easy way to grab compilation-related stuff from numpy at all?

pythran/config.py
rgommers: ah no need, `numpy` is a hard requirement. this is the only place it's needed, and I thought there was a non-python case to worry about. but there's not, I'll move it back

omp/__init__.py
rgommers: yes, from `numpy.distutils`. I'll mention it.

pythran/config.py
rgommers: I agree. After some thinking about it, this is only opt-in functionality that isn't used by SciPy - and probably used very rarely. If it does not work, better for the user to know that.

pythran/dist.py
rgommers: Yes, I think that's fine here. Any breakage due to `setuptools` will still be translated to an import error.

rgommers:
> @rgommers in the distutils-free world, is there an easy way to grab compilation-related stuff from numpy at all?

We may still get numpy build-time info via `np.show_config()`. However, I'm not sure why we'd want that - what options the user's `numpy` install was built with should not be very relevant. Also, the info can be misleading. Both conda-forge and some Linux distros always build against Netlib BLAS in order to be able to switch BLAS libraries at runtime.

Is there anything you'd actually need to know from numpy?

rgommers:
Comments addressed. The question on whether any build-time info from `numpy` is needed is still relevant. At the moment some work is going into rewriting `np.show_config` for the Meson build, so anything that's needed we can easily add.

serge-sans-paille:
Thanks a lot @rgommers for this christmas gift. I'm a bit afraid of moving away from `distutils` because it has been hiding part of the low-level compiler complexity so far, but keeping it is a dead end anyway.

>  The question on whether any build-time info from numpy is needed is still relevant. 
I'd love to have a way to "link" with the same blas library as numpy from pythran. One way to do that is to use `dlopen` on current process and discover symbols at runtime. I don't know if that's feasible on Windows, and it assumes numpy as been imported in current process. Do you see an easier / more portable way?

rgommers:
> I'd love to have a way to "link" with the same blas library as numpy from pythran. One way to do that is to use `dlopen` on current process and discover symbols at runtime. I don't know if that's feasible on Windows, and it assumes numpy as been imported in current process. Do you see an easier / more portable way?

That is for CBLAS only I assume? (otherwise there will be problems)

The portable way to use BLAS is with `scipy.linalg.cython_blas`, which takes care of all the platform and compiler specific details. For introspection of both NumPy and SciPy, see https://github.com/joblib/threadpoolctl

PR #2052, opened 2022-12-03
BUG: Include pyconfig.h first to define visibility macros

DWesl:
Make sure Python's visibility macros are set before including any system headers, which are likely to include sys/features.h

My attempts to compile SciPy failed because `posix_memalign` was not declared.  This function is guarded by `_POSIX_C_SOURCE >= 200112`; my `pyconfig.h` defines `_POSIX_C_SOURCE = 200809`.  I think this is the simplest method to ensure Python's visibility macros are set before `sys/features.h` is included by a system header.

### DIFF
--- a/pythran/pythonic/core.hpp
+++ b/pythran/pythonic/core.hpp
@@ -31,6 +31,8 @@
 #ifdef ENABLE_PYTHON_MODULE
 // Define python's visibility macros
 #include "pyconfig.h"
+
+// Some version of python define that macro on Windows, and it breaks compilation of some C++ headers.
 #ifdef copysign
 #undef copysign
 #endif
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
This introduces a regression on windows due to xsimd:

```

D:\a\pythran\pythran/pythran\xsimd/arch/xsimd_scalar.hpp(42,16): error: no member named 'copysign' in namespace 'std'

    using std::copysign;
```

Probably an xsimd issue, I'll have a look. That being said, I'd rather have this change in `core.hpp` , and it shlould be guarded by `#ifdef ENABLE_PYTHON_MODULE`

serge-sans-paille:
@DWesl : I've added a small patch on top of yours and it now validates fine. Can you confirm I can squash and merge?

DWesl:
That works for me

serge-sans-paille:
Thanks!

PR #2067, opened 2023-01-05
Bug 2066 do not call t.begin() if t is empty

jeanlaroche:
This is only an issue on windows.


serge-sans-paille:
Fixed by 3f4907910c6fb87ed2f142436c29ebb7313b07ea

PR #2078, opened 2023-02-07
Fix bad interaction between broadcasting, indexing and numpy expr

serge-sans-paille:
Showcased by lttb_pt translated and provided by Stefan van der Walt, which I am thankful to for the effort.


serge-sans-paille:
cc @stefanv , I also had to put an extra `order(C)` to avoid extra typing issues in the return type (the return can be a transposed or not transposed matrix).

PR #2081, opened 2023-02-14
Provide entry points pythran.import_pythrancode and pythran.import_py…

### LINKED ISSUE #2080: Ability to trigger compiler on import
In Cython, you can do:

```
import pyximport
pyximport.install()
```

Thereafter, if you pull in a module with a .pyx extension, it gets compiled in the background.

This may be a bit much for Pythran (since you can't use the extension to check what type of file it is), but perhaps something like:

```
import pythran
my_module = pythran.compile('my_module.py')
```

serge-sans-paille:
…thranfile

These function calls perform a full parse & compile & import cycle, making it easy to integrate pythran in an existing codebase without relying on distutils / build system integration.

Fix #2080


stefanv:
That should do the trick for now, thanks!

PR #2091, opened 2023-03-18
Pass all arguments of generated function by value

serge-sans-paille:
This is faster for scalars, but not for objects, so we need to generate std::move accordingly


serge-sans-paille:
@jeanlaroche this patchset brings significant changes, could you give it a try on your codebase?

jeanlaroche:
I will be not before Sunday. Don't have my machine with me...

jeanlaroche:
I'm getting this error in one of my scripts. Do you need me to try to determine where exactly this comes from or can you intuit what the issue is?

```
/Users/jlaroche/ua_conda_env-Darwin.x86_64/lib/python3.8/site-packages/pythran/pythonic/utils/broadcast_copy.hpp:268:23: error: no member named 'buffer' in
      '(anonymous namespace)::pythonic::types::list<long>'
      std::copy(other.buffer, other.buffer + other.flat_size(), self.buffer);
```

jeanlaroche:
I'm also getting runtime errors :( divides by zero for example.

serge-sans-paille:
Your first example is enough for me to create a reproducer, thanks. We'll discuss the divide by zero aspect afterward :-)

jeanlaroche:
Ah great, let me know when I can dig further. I think the other issues are related to the fact that an integer scalar is not getting updated from 0, I think I see that in a lot of places, and that could also be the cause for the / 0 ...

jeanlaroche:
After pulling the latest version I get this error:
```
INFO: gcc: /var/folders/j6/7t8sg1vj4q97zhh9z5cdmxbm4rz935/T/tmpba0fmb0u.cpp
In file included from /var/folders/j6/7t8sg1vj4q97zhh9z5cdmxbm4rz935/T/tmpba0fmb0u.cpp:1:
In file included from /Users/jlaroche/ua_conda_env-Darwin.x86_64/lib/python3.8/site-packages/pythran/pythonic/core.hpp:46:
In file included from /Users/jlaroche/ua_conda_env-Darwin.x86_64/lib/python3.8/site-packages/pythran/pythonic/types/slice.hpp:5:
In file included from /Users/jlaroche/ua_conda_env-Darwin.x86_64/lib/python3.8/site-packages/pythran/pythonic/types/NoneType.hpp:6:
In file included from /Users/jlaroche/ua_conda_env-Darwin.x86_64/lib/python3.8/site-packages/pythran/pythonic/builtins/bool_.hpp:6:
In file included from /Users/jlaroche/ua_conda_env-Darwin.x86_64/lib/python3.8/site-packages/pythran/pythonic/types/tuple.hpp:13:
/Users/jlaroche/ua_conda_env-Darwin.x86_64/lib/python3.8/site-packages/pythran/pythonic/types/ndarray.hpp:1058:5: error: reference to non-static member function must be called
    data = utils::shared_ref<T>(other.begin(), other.end());
    ^~~~
/Users/jlaroche/ua_conda_env-Darwin.x86_64/lib/python3.8/site-packages/pythran/pythonic/include/types/list.hpp:330:12: note: possible target for call
    dtype* data() { return _data->data();}
           ^
/Users/jlaroche/ua_conda_env-Darwin.x86_64/lib/python3.8/site-packages/pythran/pythonic/include/types/list.hpp:331:18: note: possible target for call
    const dtype* data() const { return _data->data();}
```

serge-sans-paille:
Obsoleted by #2098

PR #2097, opened 2023-04-12
Feature/faster gexpr

serge-sans-paille:


hmaarrfk:
I just have to ask,

do you use any special tools that allow you to touch so many files so quickly? are some of these autogenerated?

serge-sans-paille:
Only the basic ones : ` git grep -l '<pattern>'` to get the list of files I need to update, then a decent command of `vim` and that's it :-/

hmaarrfk:
just trying to find tools for polygots !

Issue #2095, opened 2023-04-09
Slower in Pythran than numpy?


hmaarrfk:
Hey, thanks for making this cool library. I really do believe that the advantages you outline in terms of ahead of time compilation are valuable to those building powerful scientific computation libraries.

I was trying my hand at doing a "simple" image analysis task: rgb -> nv12 conversion.

nv12 seems to be similar to yuv420 (I420) but with the U and V channels interleaved instead of on distinct planes.

This should be a simple transpose operation, but as is typical, it is easy to do this operation very slowly based on how you do it.

```
mark@nano 
--------- 
OS: Ubuntu 22.10 x86_64 
Host: 20UN005JUS ThinkPad X1 Nano Gen 1 
Kernel: 5.19.0-38-generic 
Uptime: 7 days, 1 hour, 43 mins 
Packages: 4010 (dpkg), 23 (flatpak), 12 (snap) 
Shell: bash 5.2.2 
Resolution: 2160x1350 
DE: GNOME 43.1 
WM: Mutter 
WM Theme: Adwaita 
Theme: Yaru-viridian [GTK2/3] 
Icons: Yaru-viridian [GTK2/3] 
Terminal: kitty 
CPU: 11th Gen Intel i7-1180G7 (8) @ 4.600GHz 
GPU: Intel Tiger Lake-UP4 GT2 [Iris Xe Graphics] 
Memory: 6991MiB / 15704MiB
```

The operation amounts to something like transposing a 2D Array. From `[u_0, u_1, ... u_n, v_0, v_v1, v_n]` to `[u_0, v_0, u_1, v_1, ... u_n, v_n]`.

To setup the problem, lets start with:
```python
import numpy as np
N_iter = 200
image_shape = (3072, 3072)

uv = np.full(
    (image_shape[0] // 4 * 2, image_shape[1]),
    fill_value=0,
    dtype='uint8'
)
uv.shape = (-1,)
```

For some reason, I need to do the operation in place. In numpy, this would amount to:

```python
import numpy as np
def pack_np(uv):
    uv_copy = uv.copy()
    u_size = uv.shape[0] // 2
    uv[0::2] = uv_copy[:u_size]
    uv[1::2] = uv_copy[u_size:]
```

On my computer, I get about 700-750 iterations per second with this loop:
```python
import tqdm
N_iter = 200
for i in tqdm(range(N_iter)):
    pack_np(uv)
```

[...]

PR #2098, opened 2023-04-13
Pass scalar by values for user functions

serge-sans-paille:
passing scalar by ref (through universal reference) leads to worst code, so avoid the situation when we can.


serge-sans-paille:
@jeanlaroche following up on your comments in #2095 on that PR : yeah I do plan to merge this one, but it's the most ground-breaking one, so let's dive into your issues, would you be able to share the whole error log and/or a minimal reproducer?

jeanlaroche:
I'm confused. I haven't tried 2095, even though I commented in there... 
Should I?

And yes, I'll pasted my errors.

jeanlaroche:
Here's the full error:

https://pastebin.com/PgjxVGrs

It's likely connected to code that I have to call a tensorflow library 
from the C++ code generated by pythran, but just in case that's not it.

Jean

serge-sans-paille:
@jeanlaroche could you share the original python code?

jeanlaroche:
OK I took a closer look at the code. The issue is that the calling convention used by pythran has changed with this PR so I needed to update some patches accordingly. Once I do that the compilation error disappears.

jeanlaroche:
yes!

PR #2105, opened 2023-05-02
0.13.1 - bouc'hal levnet

serge-sans-paille:


paugier:
I appreciate the nickname. Thanks @serge-sans-paille for bouc'hal and bouc'hal levnet.

jeanlaroche:
I had to lookup bouc'hal, but I'm don't know about levnet! :D

Issue #1999, opened 2022-06-25
XSIMD is used unchecked: abs.hpp: fatal error: xsimd/xsimd.hpp: No such file or directory


drew-parsons:
The pythran documentation suggests that XSIMD is only activated when `-DUSE_XSIMD` is set, and so pythran should be functional on any platform, including the many architectures unsupported by XSIMD.

But many of the header files, especially in the numpy include subdir but also elsewhere, use XSIMD unguarded. For instance pythran/pythonic/include/numpy/abs.hpp
```
#ifndef PYTHONIC_INCLUDE_NUMPY_ABS_HPP
#define PYTHONIC_INCLUDE_NUMPY_ABS_HPP

#include "pythonic/include/types/numpy_op_helper.hpp"
#include "pythonic/include/utils/meta.hpp"
#include "pythonic/include/utils/numpy_traits.hpp"
#include "pythonic/include/utils/functor.hpp"

#include <xsimd/xsimd.hpp>

PYTHONIC_NS_BEGIN

namespace numpy
{

#define NUMPY_NARY_FUNC_NAME abs
#define NUMPY_NARY_FUNC_SYM xsimd::abs
#include "pythonic/include/types/numpy_nary_expr.hpp"
}
PYTHONIC_NS_END

#endif
```
I note there is also pythran/pythonic/numpy/abs.hpp which does _not_ include the xsimd.hpp file, so perhaps I am misunderstanding the way pythran is accessing XSIMD.

But the apparent inconsistency is preventing successful builds on architectures which do not have XSIMD support, for instance ppc64el.  An attempted build log is found at https://buildd.debian.org/status/fetch.php?pkg=pythran&arch=ppc64el&ver=0.11.0%2Bds-1&stamp=1656166968&raw=0

The build is attempted over Debian build architectures, you can find the full set of build logs at https://buildd.debian.org/status/package.php?p=pythran&suite=experimental . The ones marked in red ("Build-Attempted" = failed build) are the ones trying to build without XSIMD.

serge-sans-paille:
we use the scalar part of xsimd unchecked, not the vector part. xsimd should be able to run in scalar mode on all architecture, if that's not the case, we should fix it :-)

drew-parsons:
Hmm, but then it comes back to https://github.com/xtensor-stack/xsimd/issues/756
which seemed to conclude xsimd can't build on all arches

drew-parsons:
[...]

Issue #1394, opened 2019-10-17
Cross-compilation to target many CPU architectures on Linux?


paugier:
There is nothing on cross-compilation in Pythran documentation. I understand that it is a subject in itself but sometimes, adding `-march=native` really speedup computations, but then the binary is no longer "generalist".

Would it be possible to compile for different CPU architectures so that the binaries would be both able to run on "old" machines and able to use advanced instructions where they are available?

serge-sans-paille:
There's support for that through `ifunc` (see https://gcc.gnu.org/onlinedocs/gcc-4.7.2/gcc/Function-Attributes.html) in some compilers, I'm unsure of how easy that would be to wrap into pythran :-)

serge-sans-paille:
Better: target_clones attribute :-) https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-target_005fclones-function-attribute

rgommers:
Related question I have:
- is the `.py` -> `.cpp` step guaranteed to be the same independent of hardware/os/etc. target, so you can run it with the native toolchain and use the generated `.cpp` files in a cross build?
-  Same question for the headers in `pythran.get_include()` - is there anything in there that's not portable, or is it fine to get the headers from a native package during a cross build?

PR #2107, opened 2023-05-07
Add more docs about using Pythran in a Python package

rgommers:
Include:
- transitive dependencies (boost, xsimd)
- using Pythran as a transpiler
- change `distutils` to `setuptools`
- details related to cross compilation (xref gh-1394)

Also add notes on `xsimd` always being required, and `USE_XSIMD` turning on vectorization while the default is using `xsimd`'s scalar mode (xref gh-1999).


serge-sans-paille:
<3

Issue #2126, opened 2023-07-23
Recursive import issue with Pythran when building SciPy


rgommers:
We're hitting a recursive import error when building SciPy twice with `pip install -e . --no-build-isolation`. The error happens only for editable installs of SciPy, however the recursion is also there with normal builds. The SciPy build has to do:
```meson
import pythran
pythran.get_include()
```
in its build script to obtain Pythran's include directory. However, that import statement basically pulls in all of the Pythran code base, and this errors out at:
```
pythran/tables.py", line 4565, in save_arguments
    themodule = import_module(".".join(module_name))
```
What is happening there is that that code fails when `module_name` equals `'scipy.special'`, creating an import loop.

Full traceback (from https://github.com/scipy/scipy/issues/18900):

<details>

```
Traceback (most recent call last):
  File "<string>", line 3, in <module>
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pythran/__init__.py", line 50, in <module>
    from pythran.toolchain import (generate_cxx, compile_cxxfile, compile_cxxcode,
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pythran/toolchain.py", line 6, in <module>
    from pythran.backend import Cxx, Python
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pythran/backend.py", line 7, in <module>
    from pythran.analyses import LocalNodeDeclarations, GlobalDeclarations, Scope
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pythran/analyses/__init__.py", line 12, in <module>
    from .aliases import Aliases, StrictAliases
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pythran/analyses/aliases.py", line 6, in <module>
    from pythran.tables import functions, methods, MODULES
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pythran/tables.py", line 4598, in <module>
    save_arguments((), MODULES)
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pythran/tables.py", line 4561, in save_arguments
    save_arguments(module_name + (elem,), signature)
[...]

PR #2144, opened 2023-08-29
Import top-level submodules and functions lazily

rgommers:
This ensures that `import pythran` doesn't import a lot of code which then starts to actually do things (like trying to import numpy and scipy) which may fail or have undesirable side-effects like emitting warnings.

This should also help with Cython, which does a:
```
try:
    import pythran
except ImportError:
    pass
```
which currently yields Pythran warnings even though the user doesn't want to use Pythran but only has it installed in their environment.

In addition, it allows running `pip install -e . --no-build-isolation` in SciPy twice in a row; something that was not working before because of a recursive import issue (scipy needs pythran; pythran imports scipy).

Closes gh-2126


rgommers:
CI is still chewing on it, but this looks happy now. @serge-sans-paille I had to drop Python 3.6 support, because what we need here was introduced only in Python 3.7: https://peps.python.org/pep-0562/. Since py36 and py37 are both already beyond their end-of-life date, I assume that's no problem. I noticed that Python 3.11 isn't tested at all yet, so if you want a follow-up PR for that (extending the range, or dropping Python 3.7 too), I'd be happy to do that.

serge-sans-paille:
Super clean! Thanks. +1 for py 3.11 validation setup, thanks @rgommers

rgommers:
> +1 for py 3.11 validation setup

Just to confirm: do you want me to just spread out the jobs a bit more across 3.7-3.11, or drop 3.7 too?

serge-sans-paille:
> Just to confirm: do you want me to just spread out the jobs a bit more across
> 3.7-3.11, or drop 3.7 too?

Spreading.

Issue #2139, opened 2023-08-14
New warnings: "Overriding pythran description with argspec information"


rgommers:
I just noticed a ton of these warnings with pythran `master` and Cython 3.0:
```
[16/1588] Generating 'scipy/_lib/_ccallback_c.cpython-312-x86_64-linux-gnu.so.p/_ccallback_c.c'
WARNING: Overriding pythran description with argspec information for: numpy.random.binomial
WARNING: Overriding pythran description with argspec information for: numpy.random.bytes
```
I'm unsure if this is a Pythran or a Cython issue, and what to do with these warnings. Just opening it here in case this looks familiar, because the warning explicitly mentions `pythran`.

It's not breaking anything yet as far as I can tell, it's only build log pollution.

serge-sans-paille:
This probably means that a numpy update provided extra type information, so we no longer need to provide them in pythran. I'll have a look.

rgommers:
Ah, that helps. It's probably numpy 2.0 + cython 3.0, which cause Cython to now prefer https://github.com/numpy/numpy/blob/main/numpy/__init__.cython-30.pxd over the `numpy.pxd` that it has itself.

serge-sans-paille:
mmh what's that numpy 2.0 ? Not part of pypi, right? In that case I can't fix this unless i add a hook for the WIP version of numpy, which is not something I'd like to do.

rgommers:
It should be due to https://github.com/numpy/numpy/pull/21975, which also got backported to 1.26.0b1.

rgommers:
I'm working on a fix for this.

PR #2145, opened 2023-08-29
Update signatures for numpy.random

rgommers:
Closes gh-2139

### DIFF
--- a/pythran/tables.py
+++ b/pythran/tables.py
@@ -2545,6 +2545,17 @@ _numpy_array_signature = Union[
          _complex_signature], NDArray[complex, :, :, :, :]],
 ]
 
+
+def expand_numpy_2_args(args, defaults=None):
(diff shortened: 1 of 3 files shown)


pythran/tables.py
```diff
@@ -4506,6 +4465,92 @@ def partialsum(seq):
     },
 }
 
+if numpy.__version__[0] < '2':
```
serge-sans-paille, line 4468: Instead of duplicating the dict what about the following:

```
def expand_numpy_2_args(*args):
    if numpy.__version__[0] < '2':
          return {'args': args}
    else:
          return {}

# ...
binomial": FunctionIntr(global_effects=true, **expand_numpy_2_args('n', 'p', 'size')),
...
```

serge-sans-paille:
Thanks a lot! That's much cleaner now.

PR #2146, opened 2023-08-30
CI updates: add Python 3.11, and other minor maintenance

rgommers:
- Spread out jobs more across Python versions
- Update to the latest setup-python actions
- ~Run tests in `core.yml` on Ubuntu 22.04~

One other thing I considered but didn't do is to reduce the matrix in `core.yml` across Python versions and GCC/Clang. Now everything is double; it'd be much more efficient to test like:
- py37 : gcc
- py38 : clang
- py39: gcc
- etc.
Only PyPy is different enough that it should be tested with both GCC and Clang for the same PyPy interpreter version.

@serge-sans-paille happy to do that too if you want.

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -12,12 +12,12 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, 3.10-dev, pypy-3.9]
+        python-version: [3.7, 3.9, '3.11', pypy-3.9]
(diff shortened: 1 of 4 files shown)


rgommers:
Ah, `clang-7` doesn't exist on Ubuntu 22.04. Is the goal to only test with oldest supported compiler versions in CI (Clang 7, GCC 8)? If so, I'll revert to only using Ubuntu 20.04

serge-sans-paille:
Yeah, the goal is to maintain compatibility with old compilers as long as possible, it would be great to keep that feature.

PR #2155, opened 2023-11-28
Add support for clear for lists

jeanlaroche:
I added the implementation of clear() for list. Note that I'm really not sure that my code is best way to do that (call erase(x) repeatedly)

### DIFF
--- a/pythran/pythonic/types/list.hpp
+++ b/pythran/pythonic/types/list.hpp
@@ -580,6 +580,12 @@ namespace types
     erase(x);
     return res;
   }
+  template <class T>
+  void list<T>::clear()
+  {
+      _data->clear();	
+  }
+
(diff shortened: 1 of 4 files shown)


pythran/pythonic/types/list.hpp
```diff
@@ -580,6 +580,14 @@ namespace types
     erase(x);
     return res;
   }
+  template <class T>
+  void list<T>::clear()
+  {
+    long sz = size();
```
serge-sans-paille, line 586: why not just `_data->clear()` ?

jeanlaroche: ah ah ah yes, I did not realize that _data was a std::vector! Of course.

serge-sans-paille:
clean & simple, thanks \o

PR #2162, opened 2023-12-11
Move setup configuration to pyproject.toml

serge-sans-paille:
Related to https://github.com/scipy/scipy/pull/19670


pyproject.toml
```diff
@@ -0,0 +1,55 @@
...
+[project.scripts]
+pythran = "pythran.run:run"
+pythran-config = "pythran.config:run"
+
+[build-system]
+requires = [
+    "ply>=3.4",
+    "setuptools>=61.0",
+    "gast~=0.5.0",
+    "numpy",
```
rgommers, line 41: I don't think `numpy` is needed? I can build a wheel just fine with `python -m build` or `pip install .`, and I am not aware of the content of that wheel then depending on whether `numpy` was available in the build env.

pyproject.toml
```diff
@@ -0,0 +1,55 @@
...
+dynamic = ["version", "readme", "dependencies"]
+
+[project.urls]
+Homepage = "https://github.com/serge-sans-paille/pythran"
+Documentation = "https://pythran.readthedocs.io"
+Changelog = "https://pythran.readthedocs.io/en/latest/Changelog.html"
+
+[project.optional-dependencies]
+doc = ["numpy", "nbsphinx", "scipy", "guzzle_sphinx_theme"]
+test = ["ipython", "nbval", "cython", "wheel"]
```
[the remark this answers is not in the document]
serge-sans-paille: ack, I'll do that in another PR though.

serge-sans-paille:
@rgommers : to fit in the `pyproject.toml` paradigm without writing extensions, the patch also changes the installation dir of the bundled xsimd and boost headers (see https://github.com/serge-sans-paille/pythran/pull/2162/files#diff-76bdcb439914b0c3b773139a2b33e7dfab727904613033afb2678af0afeae691). Maybe you should reflect that in scipy's buildsystem if you're not using pythran primitives to compile?

rgommers:
That change will break not only SciPy, but also scikit-image and any other users of `pythran.get_include` as far as I can tell, since the result of that is no longer enough - it only contains one of two include dirs. If you do want to get rid of a separate `third_party/` top-level directory to avoid the copy action in `setup.py`, it seems better to simply move them to the location they currently get copied to (so `pythran/xsimd` and `pythran/boost`)?

tylerjereddy:
I noticed that a recent push was made to this branch, and confirmed locally that the wheel build scenario described above is now successful, at least on x86_64 Linux with Python `3.12`. Let me know if anything else needs testing or what the plan will be, obviously I'm happy to test things to prevent slippage in the SciPy release cycle (though this is better than hacking our wheel builds, probably).

pyproject.toml
rgommers: This comment still applies I think. For the rest, this PR now LGTM.

serge-sans-paille: Numpy is a required runtime dependency of Pythran. Isn't that what's captured by that field?

serge-sans-paille:
@rgommers does that look good to you know? If so @tylerjereddy could you do a final sanity check?

tylerjereddy:
The wheel build for SciPy works for me alongside this branch.

Actually, I spent a bit of time working with `pythran` `master` branch, and it seems that is "ok" as well--my original test for `master` was too strict I think (it was to build SciPy from source, with Pythran installed, but `setuptools` removed; but `pip wheel .` I think seems to be "ok" even with `master` of Pythran because I *believe* `setup.py` will still do the trick to pull `setuptools` in during the isolated build).

I think I got confused by one of the responses suggesting that `pyproject.toml` was mandatory for the transitive dependency. I'm not entirely sure that's true, but `pyproject.toml` is the modern way anyway I guess.

So this seems "good," I wonder about the non-wheel scenario a bit, where SciPy now only needs `setuptools` to be present at build time because of Pythran with `3.12`, but that's maybe less important for the immediate issue.

serge-sans-paille:
Thanks @rgommers & @tylerjereddy for the review and testing! :bow:

PR #2163, opened 2023-12-14
Fix tuple + indexable_container combiner

### LINKED ISSUE #2160: Unnecessary alloc/copy upon return
Check out this code (sorry I could not reduce it further)
```
import numpy as np
def A():
    AA = np.arange(0, 12, dtype=np.float32)
    BB = np.empty(100, dtype=np.int32)
    CC = np.empty(100, dtype=np.float32)
    self = tuple([AA, BB, CC])
    return self

def B(self):
    self[1][0] = self[0][0]                  # self[1] int32 array. self[0] float32 array
    self[2][0] = 1.1                             # self[2] float32 array
    self[1][0:1] = self[0][0]               # <- alloc/copy upon return
    #self[1][0:1] = np.int32(self[0][0])    # <- No alloc/copy upon return
    return self[1]

#pythran export test()
def test():
    VN = A()
    print(VN)
    print('11111')
    C=B(VN)
    return C
```
If you put a breakpoint in shared_ref.hpp where new() is called, you'll see a copy/allocation at the return of function B, even though there is no need for it.
However, if you cast self[0][0] as an int32 (the type of self[1]) there is no copy/allocation.

serge-sans-paille:
The order of the combination matters, and it wasn't respected.

Fix #2160

### DIFF
--- a/pythran/pythonic/include/types/tuple.hpp
+++ b/pythran/pythonic/include/types/tuple.hpp
@@ -812,7 +812,7 @@ struct __combined<container<t>, std::tuple<t0...>> {
 PYTHONIC_NS_BEGIN
 namespace details
 {
-  template <class T, class P, bool Same>
+  template <class T, class P, bool Combine>
   struct pick_combined;
   template <class T, class P>
   struct pick_combined<T, P, true> {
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
@jeanlaroche can you confirm this fixes #2160 ?

Issue #2164, opened 2023-12-20
`spicy` build error referring to `pythran` header: `xsimd_scalar.hpp: error: '__exp10f' was not declared in this scope; did you mean 'exp10'?`


barracuda156:
Building `scipy` on 10.6 revealed an apparent problem with `pythran` headers: `__exp10f` is undefined.
```
In file included from /opt/local/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pythran/xsimd/xsimd.hpp:55,
                 from /opt/local/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pythran/pythonic/include/numpy/exp.hpp:8,
                 from scipy/interpolate/_rbfinterp_pythran.cpp:26:
/opt/local/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pythran/xsimd/arch/xsimd_scalar.hpp: In function 'float xsimd::exp10(const float&)':
/opt/local/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pythran/xsimd/arch/xsimd_scalar.hpp:474:16: error: '__exp10f' was not declared in this scope; did you mean 'exp10'?
  474 |         return __exp10f(x);
      |                ^~~~~~~~
      |                exp10
/opt/local/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pythran/xsimd/arch/xsimd_scalar.hpp: In function 'double xsimd::exp10(const double&)':
/opt/local/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pythran/xsimd/arch/xsimd_scalar.hpp:478:16: error: '__exp10' was not declared in this scope; did you mean 'exp10'?
  478 |         return __exp10(x);
      |                ^~~~~~~
      |                exp10
[...]

PR #2165, opened 2023-12-20
xsimd_scalar.hpp: fix for missing __exp10 on macOS < 10.9

barracuda156:
Fixes: https://github.com/serge-sans-paille/pythran/issues/2164

Apparently `__exp10` and `__exp10f` do not exist on macOS until 10.9: https://lists.llvm.org/pipermail/llvm-commits/Week-of-Mon-20131216/199358.html
And indeed, on 10.6 existing code is broken.

Use existing fallback instead, fix the macros.

### DIFF
--- a/pythran/xsimd/arch/xsimd_scalar.hpp
+++ b/pythran/xsimd/arch/xsimd_scalar.hpp
@@ -24,6 +24,10 @@
 #include "xtl/xcomplex.hpp"
 #endif
 
+#ifdef __APPLE__
+#include <AvailabilityMacros.h>
+#endif
+
 namespace xsimd
 {
     template <class T, class A>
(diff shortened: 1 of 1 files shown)


barracuda156:
@serge-sans-paille Should we rather use GCC builtins when those are available though? It can be added in a macro too.

serge-sans-paille:
@barracuda156 we're just vendoring xsimd, could you submit your patch there https://github.com/xtensor-stack/xsimd/pulls ? I'll backport it afterward. Thanks!

barracuda156:
> @barracuda156 we're just vendoring xsimd, could you submit your patch there https://github.com/xtensor-stack/xsimd/pulls ? I'll backport it afterward. Thanks!

Done https://github.com/xtensor-stack/xsimd/pull/993
Added gcc builtins support too.

Issue #2170, opened 2024-01-08
Questions related to Meson+Pythran


paugier:
Happy new year Serge and @rgommers !

I'm trying to build my packages using Pythran through Transonic with Meson. I tried to follow https://pythran.readthedocs.io/en/latest/MANUAL.html#integration-into-a-python-package and what is done in Scipy. Overall we now have a working solution (described in particular in these documents: [Packaging when using Transonic](https://transonic.readthedocs.io/en/latest/packaging.html) and [Build Fluidsim](https://fluidsim.readthedocs.io/en/latest/build-from-source.html)), however, I'm still unsure about what should be done for few questions related to Pythran.

Warning for @rgommers: beware if you look at the ugly things we do in meson.build files with Transonic. Python source directories with their `meson.build` files are generated from other `meson.build` files. I realize that it does not follow the Meson philosophy but it works and it's convenient. Anyway, this issue is not about that.

Pythran's documentation is quite minimalist about Meson and it might be useful to add few notes.

### ~/.pythranrc

With Meson, Pythran is used only to produce C++ code. Hence, many of the options in the .pythranrc files are not used. Anyway, it seems reasonable that a build of a package would be reproducible and not depend at all on a ~/.pythranrc file. I don't know if it is possible to tell Pythran to not consider ~/.pythranrc. I tried with `PYTHRANRC="" pythran ...` but it seems that pythran is happy with this and continues to consider ~/.pythranrc.

### `[compiler]` section

It would be nice to have a short note (or just a link) explaining how to do with meson-python what is usually done with the `[compiler]` section of Pythran configuration files. What is recommended? Just environment variables?

### complex_hook

The only Pythran option that I really use affecting C++ code generation is `complex_hook`. So with Meson, we call Pythran with `--config pythran.complex_hook=pythran_complex_hook` and we have a Meson option `pythran-complex-hook`.

In https://pythran.readthedocs.io/en/latest/MANUAL.html#customizing-your-pythranrc there is

> complex_hook:
>
>   Set this to True for faster and still Numpy-compliant complex multiplications. Not very portable, but generally works on Linux.

I don't know if it means that it could be reasonable to build wheels uploaded on PyPI to set pythran-complex-hook to true on Linux and false elsewhere? Also I don't know what would be the best way to do that with meson-python...

### `USE_XSIMD` and `-Ofast -march=native`

[...]

PR #2171, opened 2024-01-09
Provide support for '--ignore-pythranrc'

serge-sans-paille:
Related to #2170

### DIFF
--- a/docs/MANUAL.rst
+++ b/docs/MANUAL.rst
@@ -552,6 +552,9 @@ environment variable ``PYTHRANRC``::
 
     PYTHRANRC=/opt/company/pythran/config.pythranrc pythran arc_distance.py
 
+When ``PYTHRANRC`` is set to the empty string, no user-site configuration is
+loaded. This can be helpful for reproducible builds.
+
 All the options in the ``.pythranrc`` file can be specified when running pythran by using the command line argument --config= .
 For example::
 
--- a/pythran/config.py
+++ b/pythran/config.py
@@ -56,17 +56,22 @@ def get_paths_cfg(
                                             "pythran-default.cfg")
 
     user_config_path = os.environ.get('PYTHRANRC', None)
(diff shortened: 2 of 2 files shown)


serge-sans-paille:
@paugier can you double check this matches your need?

paugier:
It seems buggy. `init_cfg` is called without `ignore_userfile` before the call from `run.py`. And then there is

```
    global cfg
    if cfg is not UninitializedConfig:
        return cfg
```

Moreover, it seems to me that it would be convenient for use in Meson to be able to activate this with an environment variable, for example with something like `PYTHRANRC=0` ?

It actually already "works" (~/.pythranrc is ignored), but with a warning "WARNING:     user file does not exist: 0". We could just remove the warning for a special value.

serge-sans-paille:
@paugier I like your idea! Implementation done

paugier:
Ah I get a traceback :-)

```
  File "/home/users/augier3pi/dev/pythran/pythran/config.py", line 79, in init_cfg
    paths = get_paths_cfg(sys_file, platform_file, user_file)
  File "/home/users/augier3pi/dev/pythran/pythran/config.py", line 74, in get_paths_cfg
    paths.update((user_config_paths,))
ValueError: dictionary update sequence element #0 has length 0; 2 is required
```

It works with something like:

```python
    paths = {
        "sys": sys_config_path, "platform": platform_config_path
    }

    user_config_path = os.environ.get('PYTHRANRC', None)
    if user_config_path is None:
        user_config_dir = os.environ.get('XDG_CONFIG_HOME', None)
        if not user_config_dir:
            user_config_dir = os.environ.get('HOME', None)
        if not user_config_dir:
            user_config_dir = '~'
        user_config_path = os.path.expanduser(
            os.path.join(user_config_dir, user_file))
    if user_config_path:
        paths.update(("user", user_config_path,))
    return paths
```

PR #2176, opened 2024-01-22
Centralize all pythran-managed allocation in one place

serge-sans-paille:
This should make it easier to instrument allocations.

As a side effect slightly improve some string manipulation.


serge-sans-paille:
cc @jeanlaroche

serge-sans-paille:
@jeanlaroche : I've added a flag `--trace-allocations` to automate the instrumentation of memory allocation, is that close to what you had in mind?

jeanlaroche:
This works really well and is super super useful for me!

serge-sans-paille:
I've added some extra documentation, will merge afterward

Issue #2178, opened 2024-01-29
Compilation error no member named 'baseid' during check of overlap.


jeanlaroche:
While testing the new flag --trace-allocations I ran into this issue, which appears to be somewhat new (code compiles fine with older versions of pythran)

```
import numpy as np

# pythran export test()
def test():
    N = 100
    B = np.zeros((N,N), dtype=int)
    A = [[0] for ii in range(N)]
    for ii in range(len(A)):
        a = len(A[ii])
        B[ii, 0:a] = A[ii][0:a]
```
Here's some of the error message
```
/Users/jlaroche/ua_conda-Darwin.x86_64/miniforge/envs/jean/lib/python3.8/site-packages/pythran/pythonic/types/numpy_gexpr.hpp:59:14: error: no member named 'baseid' in '(anonymous namespace)::pythonic::types::sliced_list<long, (anonymous namespace)::pythonic::types::fast_contiguous_slice>'
    return e.baseid();
           ~ ^
/Users/jlaroche/ua_conda-Darwin.x86_64/miniforge/envs/jean/lib/python3.8/site-packages/pythran/pythonic/types/numpy_gexpr.hpp:65:12: note: in instantiation of function template specialization '(anonymous namespace)::pythonic::types::baseid_helper<(anonymous namespace)::pythonic::types::sliced_list<long, (anonymous namespace)::pythonic::types::fast_contiguous_slice>>' requested here
    return baseid_helper(
           ^
/Users/jlaroche/ua_conda-Darwin.x86_64/miniforge/envs/jean/lib/python3.8/site-packages/pythran/pythonic/types/numpy_gexpr.hpp:72:26: note: in instantiation of function template specialization '(anonymous namespace)::pythonic::types::baseid<(anonymous namespace)::pythonic::types::sliced_list<long, (anonymous namespace)::pythonic::types::fast_contiguous_slice>>' requested here
    return baseid(e0) == baseid(e1);
                         ^
[...]

PR #2179, opened 2024-02-02
Add missing baseid() in lists.

jeanlaroche:
Fixes bug #2178


serge-sans-paille:
Thanks for the PR. This is missing a test case though, ideally one that would be failing before that patch and working now.

jeanlaroche:
Ah yes, I'll put the example of the bug that this refers to.

PR #2193, opened 2024-03-31
drop python<=3.7 support

kloczek:
According to https://endoflife.date/python python 3.7 has been EOSed 27 Jun 2023.
Filter all code over `pyupgrade --py38-plus`.


docs/papers/iop2014/xp/numba/nqueens.py
```diff
@@ -43,8 +43,8 @@ def n_queens(queen_count):
     cols = range(queen_count)
     #for vec in permutations(cols):
     for vec in permutations(cols,None):
-        if (queen_count == len(set(vec[i]+i for i in cols))
```
serge-sans-paille, line 46: don't touch the input of paper examples.

pythran/cxxgen.py
```diff
@@ -69,7 +68,7 @@ def inline(self):
         if tp_decl is None:
             return tp_lines
         else:
-            return "%s %s" % (tp_lines, tp_decl)
+            return "{} {}".format(tp_lines, tp_decl)
```
serge-sans-paille, line 71: why not using fstrings here too?

serge-sans-paille (changes requested):
I'm fine with most changes - thanks for going through this. Please just don't change the papers/* files, they are associated with published papers and should stay as they are.
I've mentioned using more f-strings instead of sometimes using the format method, don't feel obliged to do so, that can be left for another round.

Issue #2141, opened 2023-08-23
Using Pythran to implement the compiled kernel of a NumPy ufunc


steppi:
Hi,

I've been investigating the possibility of using Pythran to implement the compiled kernels for [ufuncs](https://numpy.org/doc/stable/reference/ufuncs.html) in SciPy's `special` submodule. In `special`, SciPy vendors several compiled special function libraries, one in particular, an F77 library [specfun.f](https://github.com/scipy/scipy/blob/main/scipy/special/specfun/specfun.f) written for the book "Computation of Special Functions", 1996, John Wiley & Sons, Inc. by Shanjie Zhang and Jianming Jin, is problematic due to a high number of correctness bugs. Rather than trying to debug and fix the F77 code, I think it would be better to rewrite the offending functions in a modern language. I've already begun this effort for the Gaussian hypergeometric function [hyp2f1](https://github.com/scipy/scipy/blob/main/scipy/special/_hyp2f1.pxd) using Cython.

Being able to write the ufunc kernel in Pythran seems like it could be a good option, since there's a shortage of bandwidth for maintainers and contributors with simultaneous experience in both the relevant mathematics and in programming with compiled languages. At the moment Pythran is an optional dependency, but there is talk of making it required https://github.com/scipy/scipy/issues/18683, after which it might become viable to start using Pythran to generate compiled code for use in other compiled code in SciPy.

I've created a toy example of a ufunc where the compiled kernel was implemented with Pythran here, https://github.com/steppi/pythran_ufunc_example. I'd be curious if you have any feedback on whether the approach I've used is viable, or whether you anticipate it might be prone to breaking upon updates to Pythran. 

In this example I create a capsule function for the exponential integral E1, by directly translating the Fortran code in `specfun.f`.

The below snippet is from https://github.com/steppi/pythran_ufunc_example/blob/main/ufunc_example/_exp1.py

```python

import numpy as np

# pythran export capsule _exp1(float64)
def _exp1(x):
    EPS = np.finfo('float').eps
    GA = 0.5772156649015328
    if x == 0.0:
        return float("inf")
    if x <= 1:
        result = 1.
        term = 1.
        for k in range(1, 25):
            term = -term*k*x/(k + 1)**2
            result += term
            if abs(term) <= abs(result) * EPS:
                break
        return -GA - np.log(x) + x * result
    M = int(20 + np.trunc(80 / x))
    t0 = 0.0
    for k in range(M, 0, -1):
[...]

PR #2199, opened 2024-04-08
Add basic support for generation of ufunc

serge-sans-paille:
Related to #2141

### DIFF
--- a/docs/MANUAL.rst
+++ b/docs/MANUAL.rst
@@ -417,6 +417,28 @@ is the ``pythran-config`` tool - only use that if ``pythran`` is installed in
 the host architecture (i.e., the architecture on which the produced binaries
 need to run).
 
+Generating Ufuncs
+-----------------
+
+Pythran can be used to turn a function operating on scalars into a `ufunc
+<https://numpy.org/doc/stable/reference/ufuncs.html>`_ that can operate on
+arrays and scalar, honoring Numpy's broadcasting rules. A special spec
+decoration is used to achieve that goal::
+
+    #pythran export ufunc foo(double, double)
+
+This line commands Pythran to generate a *ufunc* that takes two array-like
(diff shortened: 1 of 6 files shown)


docs/MANUAL.rst
```diff
@@ -417,6 +417,22 @@ is the ``pythran-config`` tool - only use that if ``pythran`` is installed in
...
+<https://numpy.org/doc/stable/reference/ufuncs.html>`_ that can operate on
+arrays and scalar, honoring Numpy's broadcasting rules. A special spec
+decoration is used to achieve that goal::
+
+    #pythran export ufunc foo(double, double)
+
+This line commands Pythran to generate a *ufunc* that takes two array-like
+inputs whose *dtype* are double precision floats.
+
+Pythran only supports generation of *ufunc* for function that produce a single
```
paugier, line 433: functionS that produce

steppi:
This is awesome! I can’t wait to try it out. Thanks @serge-sans-paille. On a related note, we’re more than halfway through translating all of the currently existing ufuncs in `scipy.special` into C++. The plan is to make this into a separate C++ library that SciPy and other array libraries can depend on. How much work would it be to make the scalar versions of functions in `scipy.special` available inside ufunc kernels written in Pythran, if the scalar kernels are already written in C++? Would you be interested in supporting more of `scipy.special` in Pythran and if so, is this something I could help with?


@mdhaber @tirthasheshpatel. Check out how cool this is.

mdhaber:
Very cool - like exactly what we [wrote about](https://github.com/scipy/archive/blob/main/proposals/grant_proposals/NumFOCUS_SDGs/2023_NumFOCUS_SDG_Round1_Special_Fcuntions.pdf)! That plan changed, but this is still great for things that won't be in `special` or for things that are built on top of `special`.

PR #2208, opened 2024-05-14
0.16.0 - brok

serge-sans-paille:


serge-sans-paille:
As usual, @rgommers @paugier @martibosch @drew-parsons @stefanv @hroncok  : as usual, if you can give a try to that release candidate, that would prevent bad surprises later in the pipeline. Thanks for the effort!

serge-sans-paille:
and @jeanlaroche :-)

martibosch:
all good from my end: https://github.com/martibosch/pylandstats/actions/runs/9080664342/job/24952689523

note that the ci workflow includes not only the build and tests but also running all the example notebooks, so it should be quite solid as far as pylandstats' use of pythran goes.

rgommers:
No time for dedicated testing (holiday, and then post-holiday catch up), but SciPy does have a CI job that tests against Pythran master (https://github.com/scipy/scipy/blob/58e66faba645ea3b21b3f82de98d98c39c706866/.github/workflows/linux.yml#L74), so things should be in decent shape from SciPy's point of view.

serge-sans-paille:
@h-vetinari : could you give this one  a try?

h-vetinari:
Looks fine in https://github.com/serge-sans-paille/pythran/pull/2208 (not running the full test suite, but the same testing we've done for a long time).

drew-parsons:
Tests are passing on debian stable.

When building the wheel (`python3.12 -m build --wheel`...), it gives absent package warnings, e.g.
```
/usr/lib/python3/dist-packages/setuptools/command/build_py.py:204: _Warning: Package 'pythran.boost' is absent from the `packages` configuration.
!!

        ********************************************************************************
        ############################
        # Package would be ignored #
        ############################
        Python recognizes 'pythran.boost' as an importable package[^1],
        but it is absent from setuptools' `packages` configuration.
        
        This leads to an ambiguous overall configuration. If you want to distribute this
        package, please make sure that 'pythran.boost' is explicitly added
        to the `packages` configuration field.

        Alternatively, you can also rely on setuptools' discovery methods
        (for example by using `find_namespace_packages(...)`/`find_namespace:`
        instead of `find_packages(...)`/`find:`).

        You can read more about "package discovery" on setuptools documentation page:
        
        - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

        If you don't want 'pythran.boost' to be distributed and are
        already explicitly excluding 'pythran.boost' via
        `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
        you can try to use `exclude_package_data`, or `include-package-data=False` in
        combination with a more fine grained `package-data` configuration.
        
        You can read more about "package data files" on setuptools documentation page:
        
        - https://setuptools.pypa.io/en/latest/userguide/datafiles.html

        
        [^1]: For Python, any directory (with suitable naming) can be imported,
              even if it does not contain any `.py` files.
              On the other hand, currently there is no concept of package data
              directory, all directories are treated like packages.
        ********************************************************************************
              
!!
  check.warn(importable)
```

likewise for 'pythran.pythonic', 'pythran.tests', 'pythran.xsimd'

serge-sans-paille:
> Tests are passing on debian stable.
> 
> When building the wheel (`python3.12 -m build --wheel`...), it gives absent package warnings, e.g.
> 
> ```
> /usr/lib/python3/dist-packages/setuptools/command/build_py.py:204: _Warning: Package 'pythran.boost' is absent from the `packages` configuration.
> !!
> 
>         ********************************************************************************
>         ############################
>         # Package would be ignored #
>         ############################
>         Python recognizes 'pythran.boost' as an importable package[^1],
>         but it is absent from setuptools' `packages` configuration.
>         
>         This leads to an ambiguous overall configuration. If you want to distribute this
>         package, please make sure that 'pythran.boost' is explicitly added
>         to the `packages` configuration field.
> 
>         Alternatively, you can also rely on setuptools' discovery methods
>         (for example by using `find_namespace_packages(...)`/`find_namespace:`
>         instead of `find_packages(...)`/`find:`).
> 
>         You can read more about "package discovery" on setuptools documentation page:
>         
>         - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html
> 
>         If you don't want 'pythran.boost' to be distributed and are
>         already explicitly excluding 'pythran.boost' via
>         `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
>         you can try to use `exclude_package_data`, or `include-package-data=False` in
>         combination with a more fine grained `package-data` configuration.
>         
>         You can read more about "package data files" on setuptools documentation page:
>         
>         - https://setuptools.pypa.io/en/latest/userguide/datafiles.html
> 
>         
>         [^1]: For Python, any directory (with suitable naming) can be imported,
>               even if it does not contain any `.py` files.
>               On the other hand, currently there is no concept of package data
>               directory, all directories are treated like packages.
>         ********************************************************************************
>               
> !!
>   check.warn(importable)
> ```
> 
> likewise for 'pythran.pythonic', 'pythran.tests', 'pythran.xsimd' and their submodules.

Yeah, I couldn't find a way around these problems. Patch welcome :-)

serge-sans-paille:
Everything looks good, I'll probably do a release next monday. Thank you all for the testing effort.

stefanv:
All good with skimage too, thanks @serge-sans-paille.

PR #2212, opened 2024-05-27
Release/0.16.1

serge-sans-paille:


serge-sans-paille:
@rgommers / @stefanv / @jeanlaroche if you built package with pythran 0.16.0, please rebuild them with 0.16.1 once it's released, it generates native code with massive memory leaks :-/

serge-sans-paille:
Also cc @martibosch

PR #2217, opened 2024-06-26
Setup python 3.13 validation

serge-sans-paille:


serge-sans-paille:
scipy wheel don't look ready for this

PR #2218, opened 2024-06-28
Add support for scipy-openblas64

serge-sans-paille:


serge-sans-paille:
@mattip thanks a lot for providing scipy-openblas64!

mattip:
Whohoo - another user!

PR #2222, opened 2024-07-13
Make sure we have a decent optimization level for the C++ compilation…

### LINKED ISSUE #2219: Performance hit with Honer's algorithm
Consider this code to evaluate polynomials (and make pretty fractals):

```
import sys

import numpy as np
import matplotlib.pyplot as plt

import tqdm

accelerate = '--aot' in sys.argv

if accelerate:
    from extern import newton_step

else:
    def eval_poly(coeffs, x):
        # Taken from numpy.polynomial.polynomial._val
        c0 = coeffs[-1] + x*0
        for i in range(2, len(coeffs) + 1):
            c0 = coeffs[-i] + c0*x
        return c0

    def newton_step(z, poly_coeffs, der_coeffs):
        return z - eval_poly(poly_coeffs, z) / eval_poly(der_coeffs, z)

N = 1024
iter = 4

extent = 1.5
real = np.linspace(-extent, extent, N)
imag = np.linspace(-extent, extent, N)

xv, yv = np.meshgrid(real, imag)
z = xv + 1j * yv

roots = [1, np.exp(2j * np.pi / 3), np.exp(-2j * np.pi / 3)]

polynomial = np.polynomial.Polynomial.fromroots(roots)
deriv = polynomial.deriv()

poly_coeffs = polynomial.coef
der_coeffs = deriv.coef

for it in tqdm.trange(iter):
    z = newton_step(z, poly_coeffs, der_coeffs)

# -- This is just to make it pretty
frame = np.zeros((N, N, 3), dtype=np.float64)

for i, root in enumerate(roots):
    dist = np.abs(z - root)
    frame[:, :, i] = 1/(dist ** 2 + 1)

frame_norm = frame / np.linalg.norm(frame, axis=-1)[:, :, None]
plt.imshow(frame_norm)
plt.show()
```

with `extern.py`:

```def eval_poly(coeffs, x):
    c0 = coeffs[-1] + x*0
    for i in range(2, len(coeffs) + 1):
        c0 = coeffs[-i] + c0*x
    return c0

#pythran export newton_step(complex128[:, :], complex128[:], complex128[:])
def newton_step(z, poly_coeffs, der_coeffs):
    return z - eval_poly(poly_coeffs, z) / eval_poly(der_coeffs, z)
```
Running in pure Python mode, I get over 7 it/s, but with Pythran I get over 4 *s/it*. This is with both release and master, Python 3.11, on Fedora 40 (GCC 14.1.1).

I don't know what is going on, but this _should_ be the kind of problems Pythran shines at.

Restricting ourselves to reals we get an even larger slow down, from 17 it/s in pure Python, down to 3.8 s/it with Pythran:

```
z = xv + yv

roots = [1, np.exp(2j * np.pi / 3), np.exp(-2j * np.pi / 3)]

polynomial = np.polynomial.Polynomial.fromroots(roots)
deriv = polynomial.deriv()


poly_coeffs = np.real(polynomial.coef).copy()
der_coeffs = np.real(deriv.coef).copy()

for it in tqdm.trange(iter):
    z = newton_step(z, poly_coeffs, der_coeffs)
```
and `#pythran export newton_step(float64[:, :], float64[:], float64[:])`

serge-sans-paille:
… step

This used to be enforced by distutils but it seems to no longer be the case.

Fix #2219

### DIFF
--- a/pythran/config.py
+++ b/pythran/config.py
@@ -190,6 +190,11 @@ def make_extension(python, **extra):
         "extra_objects": []
     }
 
+    # In case the extension doesn't include any optimization level, make sure we
+    # have a decent default. Later options have prcedence so this is still
+    # customizable by users.
+    extension['extra_compile_args'].insert(0, "-O2")
+
     if python:
         extension['define_macros'].append('ENABLE_PYTHON_MODULE')
     extension['define_macros'].append(


serge-sans-paille:
cc @rgommers / @stefanv your builds may be impacted

rgommers:
Thanks for the ping. This doesn't affect us, since we only use Pythran to produce C++ source files, not extension modules.

Issue #2228, opened 2024-08-20
pythran fails to build with setuptools 73.0.0


tommilligan:
Found while building `scipy` which depends on `pythran`

```log
#28 59.45 pdm.termui: [651/1475] Generating 'scipy/stats/_stats_pythran.cpython-311-x86_64-linux-gnu.so.p/_stats_pythran.cpp'
#28 59.45 pdm.termui: FAILED: scipy/stats/_stats_pythran.cpython-311-x86_64-linux-gnu.so.p/_stats_pythran.cpp 
#28 59.45 pdm.termui: /tmp/pdm-build-env-4jkfvkat-shared/bin/pythran -E ../scipy/stats/_stats_pythran.py -o scipy/stats/_stats_pythran.cpython-311-x86_64-linux-gnu.so.p/_stats_pythran.cpp
#28 59.45 pdm.termui: Traceback (most recent call last):
#28 59.45 pdm.termui:   File "/tmp/pdm-build-env-4jkfvkat-shared/bin/pythran", line 8, in <module>
#28 59.45 pdm.termui:     sys.exit(run())
#28 59.45 pdm.termui:              ^^^^^
#28 59.45 pdm.termui:   File "/tmp/pdm-build-env-4jkfvkat-shared/lib/python3.11/site-packages/pythran/run.py", line 181, in run
#28 59.45 pdm.termui:     pythran.compile_pythranfile(args.input_file,
#28 59.45 pdm.termui:     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
#28 59.45 pdm.termui:   File "/tmp/pdm-build-env-4jkfvkat-shared/lib/python3.11/site-packages/pythran/__init__.py", line 127, in __getattr__
#28 59.45 pdm.termui:     import pythran.toolchain
#28 59.45 pdm.termui:   File "/tmp/pdm-build-env-4jkfvkat-shared/lib/python3.11/site-packages/pythran/toolchain.py", line 11, in <module>
#28 59.45 pdm.termui:     from pythran.dist import PythranExtension, PythranBuildExt
#28 59.45 pdm.termui:   File "/tmp/pdm-build-env-4jkfvkat-shared/lib/python3.11/site-packages/pythran/dist.py", line 139, in <module>
#28 59.45 pdm.termui:     class PythranBuildExt(PythranBuildExtMixIn, LegacyBuildExt, metaclass=PythranBuildExtMeta):
[...]

PR #2231, opened 2024-08-20
Remove dependency on setuptools for `pythran -E`

serge-sans-paille:
This should simplify the life of scipy maintainers, see #2228.


serge-sans-paille:
@agriyakhetarpal / @rgommers is the last commit of this patchset enough for scipy? DO you need a minor release with it?

rgommers:
Thanks @serge-sans-paille! This looks like it does the job indeed.

PR #2237, opened 2024-09-20
fix typos in MANUAL.rst

gdementen:


serge-sans-paille:
Thanks! I wish I could say that I made mistake just to check that someone is reading the documentation ^^!

Issue #2258, opened 2024-11-25
adding a pythran CLI flag to support free-threaded CPython


rgommers:
For free-threaded CPython (xref [PEP 703](https://peps.python.org/pep-0703/), https://py-free-threading.github.io/), Pythran seems to be working quite well already. We've tested Pythran-generated extension modules reasonably well within SciPy now with GIL-free multi-threaded tests, and so far everything looks healthy. There is one thing to do in Pythran though, namely adding a compile option to allow the user to request that the extension module is marked as compatible with free-threading.

That is pretty straightforward, it looks like this (see https://py-free-threading.github.io/porting/#declaring-free-threaded-support):
```C
static PyModuleDef_Slot module_slots[] = {
    ...
#ifdef Py_GIL_DISABLED
    {Py_mod_gil, Py_MOD_GIL_NOT_USED},
#endif
```
or for single-phase initialization:
```C
PyMODINIT_FUNC
PyInit__module(void)
{
...
#ifdef Py_GIL_DISABLED
    PyUnstable_Module_SetGIL(mod, Py_MOD_GIL_NOT_USED);
#endif
```

The way this is done in Cython and `numpy.f2py` is similar, with an opt-in flag to `cython` and `f2py`:
- Cython: `cython -Xfreethreading_compatible=True` (xref https://github.com/cython/cython/pull/6242)
- NumPy f2py: `f2py --freethreading-compatible` (xref https://github.com/numpy/numpy/pull/26981/files)

For Pythran I'd propose the same, an opt-in flag that when given inserts one of the `#ifdef Py_GIL_DISABLED` snippets in the generated C++ code:
```
pythran --freethreading-compatible
```

serge-sans-paille:
Thanks for the extra pointers. This could be a config option, probably `backend.freethreading_compatible=true` and indeed, pythran only uses the C API for conversion so a good start would be to update the module initialization and start from there.
I can handle that, unless you already have experience in that particular topic and want to give it a go?

rgommers:
> I can handle that, unless you already have experience in that particular topic and want to give it a go?

If you could handle it, that'd be great. The C module initialization part I did before, and it's the simplest part here probably - there is nothing more to it than the code snippets above. How to plumb it into Pythran and expose the CLI flag is most of the work I guess, and I'm not too sure about that.

The other part is perhaps if you'd want a CI job for free-threading? That is probably a matter of:
[...]

Issue #2257, opened 2024-11-25
Improving pythran-config so SciPy et al can use it


rgommers:
Hey @serge-sans-paille, I am looking at streamlining the SciPy build, and would like to start using `pythran-config`. To get that to work, I need a cleaner output from `pythran-config --cflags` though. What it looks like now for me:

```bash
$ pythran-config --cflags
-DENABLE_PYTHON_MODULE -D__PYTHRAN__=3 -DPYTHRAN_BLAS_BLAS -I/home/rgommers/code/tmp/pythran/pythran -I/home/rgommers/mambaforge/envs/scipy-dev-py312/lib/python3.12/site-packages/numpy/_core/include -I/home/rgommers/mambaforge/envs/scipy-dev-py312/lib/python3.12/site-packages/numpy/_core/include -I/home/rgommers/mambaforge/envs/scipy-dev-py312/include/python3.12

# or
$ pythran-config --cflags --no-python
-D__PYTHRAN__=3 -DPYTHRAN_BLAS_BLAS -I/home/rgommers/code/tmp/pythran/pythran
```

The first flavor is way too much; the second is better but requires:
1. `-DENABLE_PYTHON_MODULE` to still be there (this is pythran-specific, while the Python include paths are not) 
2. `-DPYTHRAN_BLAS_BLAS` to turn into `-DPYTHRAN_BLAS_NONE`

Also, it should work without `setuptools`.

I'm happy to make the changes needed, but there's a decent amount of design freedom here so I thought it's better to open an issue to discuss first. Options I can think of:

- A separate `--cflags-pythran-only` flag combined with a `--[no-]blas` flag
- All in one options:  `--cflags-pythran-noblas` and `--cflags-pythran-blas`
- New flags that influences the existing `--cflags`, e.g. `--cflags --pythran-only --no-blas` (this would be a bigger refactor it looks like)

The correct/desired output for SciPy as well as scikit-image looks like:
```
-DENABLE_PYTHON_MODULE -D__PYTHRAN__=3 -DPYTHRAN_BLAS_NONE -I/home/rgommers/code/tmp/pythran/pythran
```

Any preferences?

serge-sans-paille:
Hey Ralf,
I agree with you: `pythran-config` needs improvement. We could learn from `python-config`

```
% python-config -h
[...]

PR #2260, opened 2024-11-26
Feature/free threaded

serge-sans-paille:


serge-sans-paille:
@rgommers does that look good to you?

rgommers:
That was fast! CI and C++ code look good, will test with SciPy to make sure it works. 

Main question I have is about the lack of a `--freethreading-compatible` CLI flag to `pythran`, to make this opt-in (as I suggested in gh-2258). I'm fairly sure that for SciPy we don't have Pythran kernels that are thread-unsafe, but it is possible to write for example a function that modifies its input arguments, right? That kind of mutation tends to be implicitly protected by the GIL, and without it may have race conditions. That is the reason that Cython and F2py make adding `Py_MOD_GIL_NOT_USED` in generated C code opt-in, so the user can think about this.

rgommers:
The C++ code does work as advertised in SciPy. And testing this allowed me to get past the Pythran modules and find the next place to go fix the same thing, so this testing was already quite helpful.

rgommers:
Just stumbled across https://pythran.readthedocs.io/en/latest/MANUAL.html#thread-safety in the user manual. Is that still up to date, and if so should the generated code default to the behavior enabled by the `-DTHREAD_SAFE_REF_COUNT` flag under a free-threaded interpreter?

rgommers:
I think I actually like that the added flag in the `.cfg` file defaults to `true`. That makes it quite easy to start using this, and the user with odd code that turns out to be unsafe can still opt out.

serge-sans-paille:
On Wed, Nov 27, 2024 at 10:52:51AM -0800, Ralf Gommers wrote:
> Just stumbled across [1]https://pythran.readthedocs.io/en/latest/MANUAL.html#
> thread-safety in the user manual. Is that still up to date, and if so should
> the generated code default to the behavior enabled by the
> -DTHREAD_SAFE_REF_COUNT flag under a free-threaded interpreter?

I don't think so. THREAD_SAFE_REF_COUNT is for pythran-owned reference counter
on a shared pointer, it does not interact with any python memory.

rgommers:
Okay, testing done - all happy! Perhaps this still needs an entry in the docs, so the new option shows up at https://pythran.readthedocs.io/en/latest/MANUAL.html#customizing-your-pythranrc? Other than that it LGTM. 🚀

paugier:
> but it is possible to write for example a function that modifies its input arguments

Just a question about this remark. It seems to me that it is quite common for efficient numerical kernel to modify inplace numpy arrays (a very simple example [here](https://foss.heptapod.net/fluiddyn/fluidfft/-/blob/branch/default/src/fluidfft/fft3d/operators.py?ref_type=heads#L26)).

Will I have to call pythran with `--config backend.freethreading_compatible=false` in this case ?

rgommers:
@paugier only if `vector_product` in your example is a public function in a package, because then you have to guard against the user sharing the input array over multiple threads when they use the `threading` module in their own code. Such code is rare though, and typically already buggy (there's a reason you have a WARNING comment there). If it's within your own code base and you don't use `threading` yourself, that code should be safe.

rgommers:
@serge-sans-paille friendly ping on the question above. Also, will this go into 0.18.0 or can it be part of 0.17.1?

serge-sans-paille:
Can be part of a 0.17.1. Do you think we can sleak in https://github.com/serge-sans-paille/pythran/issues/2257 as well?

rgommers:
Great! I think I can get something done for gh-2257. Let me at least try to get a PR up this week that works with SciPy; if review shows it's nontrivial to get it right then we probably should aim for 0.18.0 instead.

PR #2270, opened 2024-12-18
ENH: improvements to pythran-config for build system integration

rgommers:
xref gh-2257. Putting this up for discussion early, with some comments. I considered adding `--config compiler.blas=none|openblas` support, but that then also has to interact with `--cflags` etc., so it's not clear that that works well for inclusion in 0.17.1 - and this can be added later.

The `--cflags-pythran-only` is the only thing we need for SciPy. It allows me to remove all this special handling:
https://github.com/scipy/scipy/blob/b98e79e4b326bf3beb4d861a744101ee7f438471/scipy/meson.build#L103-L138

and replace it with:
```meson
pythran_dep = []
if use_pythran
  pythran_dep = dependency('pythran', version: '>=0.17.1')
endif
```

### DIFF
--- a/pythran/config.py
+++ b/pythran/config.py
@@ -308,10 +308,7 @@ def run():
     Dump on stdout the config flags required to compile pythran-generated code.
     '''
     import argparse
-    import distutils.ccompiler
-    import distutils.sysconfig
     import pythran
-    import numpy
 
     parser = argparse.ArgumentParser(
         prog='pythran-config',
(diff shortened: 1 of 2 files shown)


pythran/config.py
```diff
@@ -344,6 +358,26 @@ def run():
...
 
+    # This should not rely on distutils/setuptools, or anything else not in the
+    # stdlib. Please don't add imports higher up.
+    if args.cflags_pythran_only:
+        if args.cflags:
+            print("Error: --cflags and --cflags-pythran-only are mutually "
+                  "exclusive, please use only one.")
+
+        compile_flags = [
+                "-D__PYTHRAN__=3",
```
serge-sans-paille, line 369: I've removed `__PYTHRAN__` in 075cefc33e461b491353be5b1222fad7be14fcb4

pythran/config.py
```diff
@@ -344,6 +358,26 @@ def run():
...
+                "-DENABLE_PYTHON_MODULE",
+                "-DPYTHRAN_BLAS_NONE",
```
serge-sans-paille, line 371: I understand that this blas flag is the one you need for scipy, but it should really depend on current config.

You mentioned

> I considered adding --config compiler.blas=none|openblas support, but that then also has to interact with --cflags etc., so it's not clear that that works well for inclusion in 0.17.1 - and this can be added later.

In the gommit message but I think that's fine, it should just be also forwaded to extension = `pythran.config.make_extension(python=args.python)` below

serge-sans-paille (changes requested):

pythran/config.py
rgommers: Okay thanks, I'll see if I can figure out how to introduce this now then.

serge-sans-paille: gentle ping :-) If that's too much of work for you, I can also accept this diff then patch it, that's ok.

serge-sans-paille:
I'll merge this one and work on a improvement to prevent it from stalling

PR #2283, opened 2025-03-04
Adjust --cflags-pythran-only implementation

serge-sans-paille:
Use argparse support and do not generate obsolete reference to __PYTHRAN__


serge-sans-paille:
cc @rgommers

PR #2321, opened 2025-06-06
Add missing operators to nditerator

cielavenir:
As nditerator implements `operator-`, it is supposed to be a random_access_iterator. Then, some comparators are missing and should be added.

pythran cannot build when `CXXFLAGS=-D_GLIBCXX_DEBUG` as https://github.com/gcc-mirror/gcc/blob/releases/gcc-14.3.0/libstdc%2B%2B-v3/include/debug/helper_functions.h#L190 uses `<=` operator:

```
scipy/interpolate/_rbfinterp_pythran.cpp:1442:130:   required from here
/usr/include/c++/10/debug/helper_functions.h:172:13: error: no match for ‘operator<=’ (operand types are ‘{anonymous}::pythonic::types::const_nditerator<{anonymous}::pythonic::types::ndarray<double, {anonymous}::pythonic::types::pshape<long int, long int> > >’ and ‘{anonymous}::pythonic::types::const_nditerator<{anonymous}::pythonic::types::ndarray<double, {anonymous}::pythonic::types::pshape<long int, long int> > >’)
  172 |  && __first <= __last;
      |     ~~~~~~~~^~~~~~~~~
```

I refered https://github.com/google/flatbuffers/pull/8309 for the implementation.

(By the way the initial motivation was similar issue in https://github.com/opencv/opencv/issues/27410)

### DIFF
--- a/pythran/pythonic/include/types/nditerator.hpp
+++ b/pythran/pythonic/include/types/nditerator.hpp
@@ -66,6 +66,9 @@ namespace types
     bool operator!=(nditerator<E> const &other) const;
     bool operator==(nditerator<E> const &other) const;
     bool operator<(nditerator<E> const &other) const;
+    bool operator>(nditerator<E> const &other) const;
+    bool operator<=(nditerator<E> const &other) const;
+    bool operator>=(nditerator<E> const &other) const;
     nditerator &operator=(nditerator const &other);
   };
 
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
Nice! Would be good to add a CI task for this. I can do it unless you want to give it a try?

jwakely:
Will this fix https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120622 ?

jwakely:
Hmm, this only adds some relops, not the missing `==` so maybe not.

cielavenir:
oh do we need similar patch for pythran/pythonic/include/numpy/ndenumerate.hpp ?

jwakely:
I think so - it's now #2324

PR #2352, opened 2025-09-07
Add support for generation of .abi3.* limited api extensions.

serge-sans-paille:
Through config/backend.limited_api entry. Possible values are false (the default), true (default to 3.7) or a python version (say 3.10).

This is a follow-up to #2350


serge-sans-paille:
@paugier I've updated this patch with some compatibility stuff, can you check it on your setup?

PR #2381, opened 2025-12-15
Fix np.concatenate when receiving negative axis and/or matrices

serge-sans-paille:


serge-sans-paille:
@ev-br looks good now!

PR #2403, opened 2026-02-03
Move benchmarks outside pythran package

stratakis:
The test_ prefix of the benchmark file was making
pytest collect it and use the benchmarks directory as
root, having the tests failing.

Rename the file so it's not collected via the usual
pytest runs.


stratakis:
It seems pytest might still be collecting this file, changing to draft for a bit till I properly test it.

stratakis:
Changing the name of the file so pytest doesn't collect it, hence it won't change directory seems to work. My initial assumptions of the lack of an init file was not correct, the file naming scheme was the culprit. Already tested in Fedora and tests run successfully.

serge-sans-paille:
thanks!

PR #2196, opened 2024-04-05
setuptools 62+ required to build

adamjstewart:
Originally discovered by @rgommers 

Starting with #2162, building pythran with setuptools 60 or older results in failure to detect project name or version:
```
  Created wheel for UNKNOWN: filename=UNKNOWN-0.0.0-py3-none-any.whl size=2302 sha256=b46d11132f5384c143756b1ed741d7a1848b92f77ddc837eaebf68ad91ae333b
```
and setuptools 61 fails with:
```
  ValueError: invalid pyproject.toml config: `tool.setuptools.dynamic`
```
Setuptools 62 or newer is required to build from source.

PR #2406, opened 2026-02-15
Freeze setuptools for older python version

serge-sans-paille:


adamjstewart:
What was the motivation behind this? Was this tested? I think https://github.com/serge-sans-paille/pythran/pull/2196 is still the case and the project won't build correctly without setuptools 62+.

PR #2421, opened 2026-04-10
Move to C++17

serge-sans-paille:


serge-sans-paille:
cc @JohanMabille and @AntoinePrv

PR #2431, opened 2026-05-26
Fix typing of array.typecodes for Python 3.15

hroncok:
Python 3.15 changed array.typecodes from str to tuple to support type codes longer than 1 character (Zf and Zd).
See https://github.com/python/cpython/issues/148675

Pythran's fill_constants_types() uses type() introspection on live Python constants to set signatures, which now produces tuple instead of str. Meanwhile, the explicit ConstantIntr(signature=str) was silently ignored because ConstantIntr.__init__ never stored it.

Fix ConstantIntr to preserve an explicitly provided signature, and skip introspection in fill_constants_types when one is already set.

Without the fix, pytest pythran/tests/test_array.py -k test_typecodes fails with:

```
        else:
>           raise NotImplementedError(t)
E           NotImplementedError: <class 'tuple'>

pythran/types/tog.py:480: NotImplementedError
```

Assisted-By: Claude Opus 4.6

### DIFF
--- a/pythran/intrinsic.py
+++ b/pythran/intrinsic.py
@@ -255,6 +255,8 @@ class ConstantIntr(Intrinsic):
         """ Forward arguments and remove arguments effects. """
         kwargs["argument_effects"] = ()
         super(ConstantIntr, self).__init__(**kwargs)
+        if 'signature' in kwargs:
+            self.signature = kwargs['signature']
 
     def isliteral(self):
         """ Mark this intrinsic as a literal. """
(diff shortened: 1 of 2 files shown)


hroncok:
I opened this as a draft. I want to see the CI results and will [test this in the Fedora package](https://copr.fedorainfracloud.org/coprs/g/python/python3.15-b1/package/pythran/).

serge-sans-paille:
Hey Miro, thanks for the PR. We don't have python 3.15 CI setup yet, I'll add that. Hopefully it will showcase the issue.

serge-sans-paille:
From what I see in https://github.com/serge-sans-paille/pythran/pull/2432, there's also a required gast update to do, I'll have a look.

serge-sans-paille:
#2432 confirms the issue

serge-sans-paille:
Merged as 1165bb09b8d502e16a403667ce084270af4dcd4a on top of #2432

Issue #2433, opened 2026-05-28
Bug: pythran does not match python. Latest version mac OS


jeanlaroche:
In the following reproducer, the pythran code outputs 2 different values while python outputs the same in both cases.

```
import numpy as np


def sum_via_lazy(x):
    """BAD"""
    iMax = x.shape[0]
    ramp = np.expand_dims(np.linspace(1, 2, iMax), -1)   # shape (iMax, 1)
    print(ramp* x[0:iMax, :])
    return float(np.sum(ramp * x[0:iMax, :]))
    #return float(np.sum((ramp * x[0:iMax, :]).copy())). # <--- this does not compile BTW.



def sum_via_materialized(x):
    """GOOD: pre-allocate `prod`, then fill via `prod[:] = ramp * x[...]`."""
    iMax = x.shape[0]
    ramp = np.expand_dims(np.linspace(1, 2, iMax), -1)   # shape (iMax, 1)
    prod = np.empty((iMax, x.shape[1]), dtype=np.float64)
    prod[:] = ramp * x[0:iMax, :]
    print(prod)
    return float(np.sum(prod))

#pythran export main_test()
def main_test():
    x = np.array([
        [0.123,  0.246,  0.123,  1.0, -1.4,   0.58],
        [1.0,    2.0,    1.0,    1.0, -1.0,   0.61],
    ], dtype=np.float64)

    bad  = sum_via_lazy(x)
    good = sum_via_materialized(x)

    print(f"sum_via_lazy         = {bad:.6f}")
    print(f"sum_via_materialized = {good:.6f}")
```

serge-sans-paille:
@jeanlaroche can we close it now?

jeanlaroche:
yes!!!

PR #2434, opened 2026-05-29
Fix broadcasting issue.

jeanlaroche:
I used claude code to diagnose and fix the broadcasting issue (hard to believe we've never encountered it before).
Jean

Fix _no_broadcast_vectorize on inner-dim broadcast with runtime shape

The check only compared arg.shape<0>() == size(), so a broadcast on
a non-outer dim (e.g. a (2,1) arg from expand_dims against (2,3))
got mis-classified as "no broadcast" when the arg's shape type was
fully runtime (no integral_constant<long, 1> marker). Mirror
_no_broadcast_ex by comparing the full shape with the
is_trivial_broadcast short-circuit. Also adds .copy() to numpy_expr.

### DIFF
--- a/pythran/pythonic/numpy/reduce.hpp
+++ b/pythran/pythonic/numpy/reduce.hpp
@@ -139,7 +139,7 @@ namespace numpy
     template <class T>
     reduce_result_type<Op, E> operator()(E const &expr, T p) const
     {
-      if (utils::no_broadcast_vectorize(expr))
+      if (utils::no_broadcast_ex(expr))
         return _reduce<Op, E::value, types::vectorizer_nobroadcast>{}(expr, p);
       else
         return _reduce<Op, E::value, types::vectorizer>{}(expr, p);
--- a/pythran/tests/test_numpy_broadcasting.py
+++ b/pythran/tests/test_numpy_broadcasting.py
@@ -232,3 +232,9 @@ class TestBroadcasting(TestEnv):
             data[1:] = np.ones([1, n - 1])
             return data'''
         self.run_test(code, 4, downcasting1=[int])
+
+    def test_broadcast_sum_runtime_shape(self):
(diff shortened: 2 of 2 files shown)


serge-sans-paille:
Thanks for the patch. I'm not quite sure it's a good fix, but at least with the reproducer I can investigate :-)

jeanlaroche:
True, if you prefer I can just pass the reproducer and forget about the solution suggested by claude, which I don't really trust myself.
I've encountered a few more compilation issues recently, if you're still up for it I have a few minimal reproducers.

pythran/pythonic/include/types/numpy_expr.hpp
```diff
@@ -779,6 +779,11 @@ namespace types
     long flat_size() const;
 
     long size() const;
+
+    ndarray<dtype, shape_t> copy() const
```
serge-sans-paille: this does not seem needed.

pythran/pythonic/numpy/reduce.hpp
```diff
@@ -139,7 +139,7 @@ namespace numpy
     template <class T>
     reduce_result_type<Op, E> operator()(E const &expr, T p) const
     {
-      if (utils::no_broadcast_vectorize(expr))
+      if (utils::no_broadcast_ex(expr))
```
serge-sans-paille, line 142: but this works

serge-sans-paille (commented):
Can you remove the change in  `numpy_expr` and see how it goes?

jeanlaroche:
The other fix fixes the copy() compilation error that was in my bug report so it's not needed for the unit test, but it's needed to fix 
```    #return float(np.sum((ramp * x[0:iMax, :]).copy())). # <--- this does not compile BTW. ```
in bug https://github.com/serge-sans-paille/pythran/issues/2433

serge-sans-paille:
ok, let's create a new PR for that other issue then

PR #2442, opened 2026-07-01
Fix slice construction when the slice bound is a compile-time const

jeanlaroche:
This fix was suggested by claude, fixes bug 2441

### DIFF
--- a/pythran/pythonic/include/types/NoneType.hpp
+++ b/pythran/pythonic/include/types/NoneType.hpp
@@ -116,6 +116,8 @@ namespace types
     none();
     none(none_type const &);
     none(T const &data);
+    template <class U, U V>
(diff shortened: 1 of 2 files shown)


pythran/pythonic/include/types/NoneType.hpp
```diff
@@ -116,6 +116,8 @@ namespace types
     none();
     none(none_type const &);
     none(T const &data);
+    template <class U, U V>
+    none(std::integral_constant<U, V> const &) : data(V), is_none(false) {}
```
serge-sans-paille, line 120: I think we should check if `T == U` or at least if U is convertible to T. I'll double check that.

serge-sans-paille:
Obsoleted by #2443 , the AI-generated solution was not generic, and the example was full of cruft.

PR #2464, opened 2026-08-03
Fix combiner between numpy_gexpr and numpy_expr

### LINKED ISSUE #2456: Loop-carried lazy expression aliases itself, giving a wrong result
```python
import numpy as np

#pythran export go(float[:], int)
def go(a, n):
    acc = a[0:2]
    for i in range(n):
        acc = acc + a[i:i + 2]
    return np.sum(acc)
```

```
>>> go(np.array([3., -1., 4., -1., 5., 9., 2., 6.]), 3)
9.0   # compiled
10.0   # CPython
```

pythran 0.18.1, Apple clang 15.0.0, Python 3.13.13, numpy 2.3.4, macOS 14.8.5 arm64

serge-sans-paille:
Fix #2456


serge-sans-paille:
Obsoleted by #2480

PR #2479, opened 2026-08-28
Support limited API for free-threaded code

### LINKED ISSUE #2445: Add support for free-threaded stable ABI
Python 3.15 has support for free-threaded stable ABI extensions with [PEP 803](https://peps.python.org/pep-0803/). This is feature request for Pythran to add support for building abi3t pythran extensions similar to how it supports building limited API extensions. 

Pythran depends on the numpy C API, which already supports the PEP 803 opaque PyObject layout (numpy/numpy#30704), so Pythran can support this ABI by requiring `numpy ≥ 2.5` for free-threaded stable ABI builds.

From a brief look, there are two important changes needed to implement this: 

1. **Build flags (`pythran/config.py`)** — when `backend.limited_api` is enabled on a
   free-threaded 3.15+ interpreter, define `Py_TARGET_ABI3T` alongside `Py_LIMITED_API`

2. **Module init codegen (`pythran/cxxgen.py`)** —  For free-threaded stable ABI `PyModExport_` needs to be implemented as `PyModuleDef ` is not usable on 3.15+ free-threaded ABI.

serge-sans-paille:
Fix #2445


espressolee (changes requested):
I tested exact head `d2c3fdacfd3343a3fe269e1227d57085c48c9148` on macOS/arm64 with CPython 3.15.0b4t, NumPy 2.5.2, and Apple Clang. The current 3.13t CI is a false-positive for the claimed abi3t support: those headers do not implement `Py_TARGET_ABI3T`, so defining it does not enforce the free-threaded Limited API.

There are two blocking failures on the first CPython version that implements PEP 803.

First, `backend.limited_api = true` still selects the historical abi3 minimum `(3, 7)`, so `make_extension()` emits:

```text
Py_TARGET_ABI3T=0x03070000
```

CPython 3.15 rejects this at `Include/pyabi.h`:

```text
error: "Py_TARGET_ABI3T must be 0x030f0000 (3.15) or above"
```

Second, after explicitly setting `backend.limited_api = 3.15` as a control, the target-version check is satisfied but the generated module still fails to compile because it uses the legacy module-definition path:

```text
use of undeclared identifier 'PyModuleDef_HEAD_INIT'
variable has incomplete type 'struct PyModuleDef'
use of undeclared identifier 'PyUnstable_Module_SetGIL'
```

This matches the remaining requirement already stated in #2445: abi3t module initialization needs the PEP 793/803 export API rather than `static PyModuleDef` + `PyModule_Create` + `PyUnstable_Module_SetGIL`.

The float/complex and NumPy 2.5 accessor changes appear directionally necessary, but they are not enough to produce an abi3t extension. Please:

1. only select abi3t on CPython 3.15+ and clamp/reject targets below 3.15;
2. implement the abi3t-compatible module export/initialization path (including the runtime ABI declaration required by PEP 803);
3. run the limited-API job on an actual 3.15t interpreter and compile/import a generated extension, rather than treating a 3.13t build that ignores `Py_TARGET_ABI3T` as qualification.

Minimal local source was a two-float exported function. Both failures occur before link/import, so they are independent of the function body. I did not run the full Pythran suite on 3.15t because the central generated extension does not compile yet.

